PR #17, opened 2018-04-10
Working devstack

regisb:
This allows the user to run their own devstack inside the containers.
Yay!

Also, we handle file permissions cleanly: in `docker-entrypoint.sh` we chmod the data and edx-platform files to the same UID of the user on the host machine. No more permission headaches!

@natea  @sampaccoud can you give it a try? I think it's pretty neat ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,13 @@
 # Open edX quick install (in Docker containers)
 
-This is a **one-click production install of [Open edX](https://openedx.org)**.
+This is a **one-click install of [Open edX](https://openedx.org), both for production and local development**.
 
-The deployment of a full-featured Open edX platform is a highly technical and complex project... but we can remove most of the complexity by relying on pre-configured Docker containers and by using only a subset of all possible features. We made this so that non-technical people could still install Open edX by themselves: knowing how to launch a server and ssh into it should be enough. But we also made sure that every step of the deploy process could be customized if you have the technical skills.
+The deployment of a full-featured Open edX platform is a highly technical and complex project. Here, we greatly simplify it by:
+
+1. relying on pre-configured Docker containers for external services, such as MySQL and MongoDb
+2. activating only a subset of all Open edX features
+
+We made this project so that non-technical people could still install Open edX by themselves: knowing how to launch a server and ssh into it should be enough. But we also made sure that every step of the deploy process could be customized if you have the technical skills.
 
 ## Quickstart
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -102,6 +113,74 @@ Open a python shell in the lms or the cms:
...
+    make build
+    make migrate
+
+Point to your local install of [edx-platform](https://github.com/edx/edx-platform/) on your host machine:
+
+    export EDX_PLATFORM_REPOSITORY=/path/to/your/edx-platform
+
+Point to your settings file:
+
+    export EDX_PLATFORM_REPOSITORY=development
```
natea: @regisb shouldn't this be `EDX_PLATFORM_SETTINGS` and not `EDX_PLATFORM_REPOSITORY`?

regisb: Yes, absolutely.

README.md
```diff
@@ -102,6 +113,74 @@ Open a python shell in the lms or the cms:
...
+    ...
+
+Then, build the images and prepare the database:
+
+    make build
+    make migrate
+
+Point to your local install of [edx-platform](https://github.com/edx/edx-platform/) on your host machine:
+
+    export EDX_PLATFORM_REPOSITORY=/path/to/your/edx-platform
```
natea: @regisb Is the right environment variable or should it be `EDX_PLATFORM_PATH`?

edxapp/Dockerfile
```diff
@@ -23,15 +23,19 @@ VOLUME /openedx/data
 WORKDIR /openedx/edx-platform
 
 ## Checkout edx-platform code
-RUN git clone https://github.com/edx/edx-platform.git --branch open-release/ginkgo.master --depth 1 .
+ARG EDX_PLATFORM_REPOSITORY=https://github.com/edx/edx-platform.git
```
natea: This is nice because it also makes it easier to use a forked version of edx-platform.

edxapp/Dockerfile
```diff
@@ -23,15 +23,19 @@ VOLUME /openedx/data
...
+ARG EDX_PLATFORM_VERSION=open-release/ginkgo.master
```
natea: And try out different branches of the repo instead of hardcoding it to open-release/ginkgo.master

README.md
regisb: You're right, it should be EDX_PLATFORM_PATH

edxapp/Dockerfile
```diff
@@ -23,15 +23,19 @@ VOLUME /openedx/data
...
+# will be overwritten when we mount edx-platform)
+RUN pip install --src ../venv/src -r requirements/edx/pre.txt
+RUN pip install --src ../venv/src -r requirements/edx/github.txt
+RUN pip install --src ../venv/src -r requirements/edx/local.txt
+RUN pip install --src ../venv/src -r requirements/edx/base.txt
+RUN pip install --src ../venv/src -r requirements/edx/post.txt
+RUN pip install --src ../venv/src -r requirements/edx/paver.txt
 
 # Install nodejs requirements
 RUN npm install
```
sampaccoud: node_modules is also overriden when you mount edx-platform.

regisb: Hmmm I had not realised that... Is that an issue? Do we need node_modules at runtime?

sampaccoud: not at runtime but if you use this as a devstack and the developper is changing things on the frontend then he might need to recompile the js I presume?

sampaccoud: One solution could be to re-run "npm install" after mounting your local "edx-platform" ? I chose to "docker cp" the edx-platform from the container to my local disk before doing the mount but I'm not fully satisfied with the beauty of it :-/  We need a better idea...

regisb: I'll assume that Open edX developers are aware that they need to recompile. So they will run `paver assets` in the devstack, wich will install node_modules in the mounted repository and will solve the problem.

I strongly believe that mounting edx-platform in a volume is the way to go: this is the only way to have seamless integration with quick reload in the devstack.

PR #56, opened 2018-09-02
Allow to accept user-specified Email address

flytreeleft:
For the variables`DEFAULT_FROM_EMAIL`, `DEFAULT_FEEDBACK_EMAIL`, `SERVER_EMAIL`, etc.


regisb:
@flytreeleft,
Thanks for opening this PR! I understand your need to override contact emails. However, with the proposed change, the user will have to manually edit the `config/openedx/lms.env.json` file after running `make configure`. Why is that a better solution than to directly edit `config/openedx/universal/lms/production.py` file? 

Note that in the future I would like to get rid of the env.json and auth.json files, which are an artifact of Open edX configuration which never made much sense to me. I'd rather store all configuration in a single file: `production.py` or `development.py`. So it would be better if your proposed change went into that direction. 

In order for me to better understand your need, can you tell me which email addresses you expect override? Will you set them all to the same email address, or do you have custom contact addresses for each purpose? (contact, feedback, press, etc.)

flytreeleft:
@regisb Sorry, I think I didn't realize the role of `config/openedx/universal/lms/production.py ` and `config/openedx/lms.env.json` :-(

Putting variables in `production.py` or `development.py` is OK.

In my case, I have to use the same email address for **all email variables**, because I haven't my own email service, and registering different public emails is quite difficult and will make some trouble, many email service providers ask me to provide my phone number and input the register code that was sent to my phone. So, using the same email will save my life :-)

By the way, it will be better if we can configure these email variables via the environments, so we can set them automatically.

PS: Thanks for your nice work! :+1:

regisb:
I agree that your scenario makes total sense. However, your PR does not solve the problem properly, as dicussed, because it requires the user to manually edit the `*.env.json` files.

Instead, what I propose is that all emails are sent from a custom email address that will be asked from the user during `make configure`. Is that OK with you?

regisb:
Fixed by https://github.com/regisb/openedx-docker/commit/f76a4feb539ff506a1fafc33b70eee84bd1a4053

flytreeleft:
@regisb Yes, that's OK. Thank you! :)

Issue #69, opened 2018-09-24
How can I change  nginx port numbers?


frohro:
I am also running an instance of apache2 for other web duties on my server, and would like to have open edX use ports 8443, and 8080.  I went into the nginx running container and installed vi and edited the ports in /etc/nginx/... for the notes, cms and lms, but I can't find where to change port 443, and it still conflicts with apache2.  Do you have any tips?
I grepped your source code, and found a lot of :443's in there.  I am contemplating changing them and rebuilding, but before I try that, I thought asking might be better.

Thanks,
Rob

tony-h:
The port numbers that are accessible from outside of the docker instances are defined in the [docker-compose.yml](https://github.com/regisb/openedx-docker/blob/master/docker-compose.yml) file. They come in pairs. The first number is the port to access the Docker instance from the outside (the published port). The second number is the internal port of the Docker instance. You would change your port configurations to 8080:80 and 8443:443, which forwards 8080 and 8443 to their respective ports in the docker instances. You can read about how Docker publishes ports at https://docs.docker.com/config/containers/container-networking/

For reference, here is my configuration. 
```
  nginx:
    image: nginx:1.13
    restart: unless-stopped
    ports:
      - "8091:80"
      - "4491:443"
```

frohro:
Thanks, that worked perfectly, after I changed the previous edits.
Again, thanks!
Rob

tony-h:
Glad to hear!

regisb:
Thanks @tony-h for the support, as always ;)
@frohro unfortunately, there's no easy way to make these port numbers configurable, so you'll have to follow Tony-h's recommendation and edit them manually, for now.

frohro:
One problem I'm having with this method is that edX creates URLs for emails, and maybe other places that do not contain the :8080.  I'm trying to figure out how to fix this.  Is it possible with VirtualHosts, or should I make configure with the :8080 on the domain name, or is there a better way to fix this?  Thanks for the help.  I'm just a teacher, not an IT professional.  :-)

regisb:
@frohro you're right, the proposed solution isn't bulletproof. IMHO, it's really not practical to have to rely on http://domain:port urls. These urls are ugly and it's hard to give them to people, as most people just don't understand the `:port` part. 
Would you consider modifying your apache configuration on the host? All you would have to do is to setup a proxy from forwarding trafic from yourdomain:80 and yourdomain:443 to localhost:8080 and localhost:8443, respectively.

[...]

PR #92, opened 2018-11-13
Update docker-compose.yml

frob:
Added optional alternate ports for nginx to aid in proxy pass deployments without needing to edit docker-compose file.

To deploy with alternate ports, add a `.env` file with the following contents:

```
OPENEDX_NGINX_PORT=8080
OPENEDX_NGINX_TLSPORT=8443
```

The .env file should not be added to the repository.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,8 +41,8 @@ services:
     image: nginx:1.13
     restart: unless-stopped
     ports:
-      - "80:80"
-      - "443:443"
+      - "${OPENEDX_NGINX_PORT:-80}:80"
+      - "${OPENEDX_NGINX_TLSPORT:-443}443:443"
     volumes:
       - ./config/nginx:/etc/nginx/conf.d/
       - ./data/lms:/openedx/data/lms:ro
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,8 +41,8 @@ services:
     image: nginx:1.13
     restart: unless-stopped
     ports:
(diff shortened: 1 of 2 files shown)


frob:
This is to aid the issues brought up in #69

docker-compose.yml
```diff
@@ -41,8 +41,8 @@ services:
     image: nginx:1.13
     restart: unless-stopped
     ports:
-      - "80:80"
-      - "443:443"
+      - "${OPENEDX_NGINX_PORT:-80}:80"
+      - "${OPENEDX_NGINX_TLSPORT:-443}443:443"
```
regisb: Can I suggest naming the variables `NGINX_HTTP_PORT` and  `NGINX_HTTPS_PORT`? We don't need the `OPENEDX` prefix, as we are inside the openedx-docker repo, and HTTPS is more understandable (IMHO) than TLS.

Also, there is a typo in the 443 declaration. It should be:

    - "${OPENEDX_NGINX_TLSPORT:-443}:443"

regisb:
This is a great idea, thanks @frob! We haven't used the docker-compose `.env` file for now, but it could be useful in many more ways.

docker-compose.yml
frob: Done

frob:
> it could be useful in many more ways.

My thoughts exactly.

regisb:
Merged in 2f9766a3a193707be846520a535a70b070531610
Thanks @frob for working on this!

PR #98, opened 2018-11-23
 Make running containers as the host user optional

silviot:
The chown operation inside docker-entrypoint.sh is extremely costly.

By virtue of how docker works, it requires a full copy of all files
that need an owner change.
See https://medium.com/@lmakarov/the-backlash-of-chmod-chown-mv-in-your-dockerfile-f12fe08c0b55

This amounts to more than 60k files and more than one GB.
It takes time and space to copy all these files.

This PR makes it possible to opt out and not have the `chown` command invoked on every container startup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ endif
 ifeq ($(ACTIVATE_HTTPS), 1)
 	post_configure_targets += https-certificate
 endif
-extra_migrate_targets = 
+extra_migrate_targets =
 ifeq ($(ACTIVATE_XQUEUE), 1)
 	extra_migrate_targets += migrate-xqueue
 	DOCKER_COMPOSE += -f docker-compose-xqueue.yml
@@ -131,7 +131,7 @@ assets-development-cms:
 		&& NODE_ENV=development ./node_modules/.bin/webpack --config=webpack.dev.config.js \
 		&& ./manage.py cms --settings=$(EDX_PLATFORM_SETTINGS) compile_sass studio \
 		&& python -c \"import pavelib.assets; pavelib.assets.collect_assets(['studio'], '$(EDX_PLATFORM_SETTINGS)')\""
-	
+
 
 ##################### Information
 
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,17 @@ endif
 ifeq ($(ACTIVATE_PORTAINER), 1)
 	DOCKER_COMPOSE += -f docker-compose-portainer.yml
 endif
+DOCKER_USER_OPTIONS =
+ifeq ($(DONT_RUN_AS_ROOT), 1)
+	DOCKER_USER_OPTIONS = -e USERID=$(USERID)
+endif
+
+ifeq ($(ACTIVATE_PORTAINER), 1)
+	DOCKER_COMPOSE += -f docker-compose-portainer.yml
+endif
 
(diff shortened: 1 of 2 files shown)


regisb:
Hi @silviot! Thanks for taking the time to open this PR!

I understand your concern about the `chmod` operation which makes the containers much slower to run. However, unless I'm mistaken, I don't think that it affects the size of the docker images. If I read the article you mentioned correctly, the size of the image is affected only if the `chmod` command is executed at build time. Here, it is executed at run time.

It's difficult to keep the right balance between efficiency and ease of use. I don't want mounted file permissions to become a difficult issue to understand for users. On the other, we probably only need to run `chmod` when edx-platform is mounted from the host, i.e: in "development" mode. I'm thinking that maybe we could pass the USERID environment variable only when EDX_PLATFORM_PATH is defined. I'll experiment with this.

silviot:
Hi! Thanks for the quick response! Also, thanks for the work on `openedx-docker`: I love it.

You're right that the size of the images is not affected. Still, when running the containers, two gigabytes and a long time are spent copying the files.

I'm looking into the development mode to understand how it works.

I understand your concerns about ease of use, but it's not clear to me the exact scenario where the `chown` would be useful.

In the interest of making sure I understand the philosophy of the project, I'd like to ask a question on a related matter.

One requirement for ease of use IIUC is that the user should be able to clone this repo, run `make all`, play around for a bit, and then delete the cloned repo without the need to fiddle with `chown` or `chmod`. This is currently not possible as these directories are not owned by the user running `make`:

```
drwx------ 2 999 999  20480 Nov 21 16:44 ./data/mysql/mysql
drwx------ 2 999 999 143360 Nov 21 16:51 ./data/mysql/openedx
drwx------ 2 999 999  12288 Nov 21 16:44 ./data/mysql/performance_schema
drwx------ 2   0   0   4096 Nov 22 10:22 ./data/portainer/compose
drwx------ 2   0   0   4096 Nov 22 10:22 ./data/portainer/tls
```

Do I understand correctly that this is an issue that needs fixing? Or is it intended behaviour?

regisb:
The files for portainer, mysql, and other services are owned by some random user, and I'm fine with that. However, for edx-platform it's much better if the files are owned by the user running the commands. There is one specific reason for that: developers who need to modify and test edx-platform locally will want to modify their own local repo. On my machine, it's located in `~/code/edx-platform`. Whenever I want to hack on edx-platform, I define `EDX_PLATFORM_PATH=/home/regis/code/edx-platform` and run `make lms`. Then I can run `./manage.py lms runserver 0.0.0.0:8000` and all kinds of commands and test my local patches to edx-platform. If the container user does not have the same UID as the user on the host, then my edx-platform repo gets filled with files that do no belong to the right user: *.pyc files, compiled static assets, etc.

So, to answer your question: the `chmod` command is only really necessary whenever we mount edx-platform from the host. 

The other typical usage scenario is for someone to run vanilla Open edX in production, without ever requiring to mount a local edx-platform repo. In that case, they just have to run `make all` and they're good to go. Whenever they want to get rid of openedx-docker, they just have to run `sudo rm -rf openedx-docker`. As you noticed, `sudo` is necessary, but it's a minor inconvenience.

To conclude, I will make sure that `chmod` is not run outside of development mode. It should considerably accelerate pretty much every command.

silviot:
Thanks for the clarification. I see now what problems the `chown` is solving.

Also thanks for making it run only in development!

PR #99, opened 2018-11-23
Add missing line continuation in `assets-development-[cl]ms`

silviot:


regisb:
Good catch! Thanks @silviot!

PR #102, opened 2018-12-07
Assets

regisb:
@silviot what do you think of that? This PR packages themes and static assets inside the container. There is still a `make assets` command, but all it does is to copy files to the shared nginx volume.


openedx/Dockerfile
```diff
@@ -37,52 +53,58 @@ RUN pip uninstall -y ora2 && \
...
+# argument: THEMES=dark-theme,edx.org
+# Custom themes added to /openedx/themes can also be compiled.
+# Here, we don't run "paver update_assets" which is slow, compiles all themes
+# and requires a complex settings file. Instead, we decompose the commands
+# and run each one individually to collect the production static assets to
+# /openedx/data/staticfiles.
+# 1. xmodule_assets: run xmodule.static_content.main, which lists all xblocks and generates webpack manifest in common/static/xmodule
+# 2. pavelib.assets.process_npm_assets: copy libraries installed via npm to common/static/common/*/vendor
+# 3. webpack: generate webpack-stats.json
+# 4. compile_sass: compile each sass file individually with libsass. The list of themes to process
```
silviot:
```suggestion
# 4. compile_sass: compile each sass file individually with libsass.
```

README.md
```diff
@@ -254,26 +254,59 @@ This will open a shell in the LMS (or CMS) container. You can then run just any
...
+                ...
+            lms/
+                ...
+        mycustomtheme2/
+            ...
+
+Then you must rebuild the openedx Docker image and indicate which themes you want to build:
 
-Comprehensive theming is enabled by default. Just drop your themes in `data/themes` and compile assets:
+    make build-openedx THEMES=mycustomtheme1,mycustomtheme2,dark-theme
```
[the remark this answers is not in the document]
regisb: I agree with you, and I did try to do without the THEMES variable, but did not manage to list the directories in /openedx/themes and join the names with a comma. However, I just discovered the `-m` option to `ls`, so something like that could work:

    paver compile_sass --theme-dirs=/openedx/edx-platform/themes,/openedx/themes --themes=$(ls -m -d /openedx/themes/ | sed 's/\////g' | sed 's/ //g')

It's really ugly, though, and I'd like to have somehting better.

Makefile
```diff
@@ -99,24 +99,19 @@ reindex-courses: ## Refresh course index so they can be found in the LMS search
...
-		&& ./manage.py lms --settings=$(EDX_PLATFORM_SETTINGS) compile_sass lms \
-		&& python -c \"import pavelib.assets; pavelib.assets.collect_assets(['lms'], '$(EDX_PLATFORM_SETTINGS)')\""
-assets-cms:
-	$(DOCKER_COMPOSE_RUN_OPENEDX) --no-deps cms bash -c \
-		"NODE_ENV=production ./node_modules/.bin/webpack --config=webpack.prod.config.js \
-		&& ./manage.py cms --settings=$(EDX_PLATFORM_SETTINGS) compile_sass studio \
-		&& python -c \"import pavelib.assets; pavelib.assets.collect_assets(['studio'], '$(EDX_PLATFORM_SETTINGS)')\""
+assets: ## Generate production-ready static assets
+	docker-compose -f docker-compose-scripts.yml run --rm \
+		--volume=$(PWD)/data/:/data/lms/ --volume=$(PWD)/data/cms/:/data/cms/ openedx bash -c \
```
silviot:
```suggestion
		--volume=$(PWD)/data/lms/:/data/lms/ --volume=$(PWD)/data/cms/:/data/cms/ openedx bash -c \
```

README.md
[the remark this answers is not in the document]
silviot: How about `$(ls /openedx/themes | paste -s -d, -)`?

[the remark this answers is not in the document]
regisb: TIL `paste` :)

This is an improvement, but we still need to make sure that only directories get listed, and not stray files.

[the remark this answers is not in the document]
silviot: The risk of listing stray files is that a message like this is emitted from the assets generation step:

> Skipping 'a-file-that-should-be-ignored': 
> Theme (a-file-that-should-be-ignored) not found in any of the theme dirs (/openedx/edx-platform/themes, /openedx/themes).

I see this as an acceptable nuisance.

openedx/Dockerfile
```diff
@@ -37,52 +53,58 @@ RUN pip uninstall -y ora2 && \
...
-# settings, nor mounting the staticfiles/ folder
-RUN xmodule_assets common/static/xmodule
-RUN python -c "import pavelib.assets; pavelib.assets.process_npm_assets()"
+# Install private requirements: this is useful for installing custom xblocks.
+# In particular, to install xblocks from a private repository, clone the
+# respositories to ./requirements on the host and add `-e ./myxblock/` to
+# ./requirements/private.txt.
+COPY ./requirements/ /openedx/requirements
+RUN touch /openedx/requirements/private.txt \
+    && pip install --src ../venv/src -r /openedx/requirements/private.txt
```
silviot:
```suggestion
    && pip install -e -r /openedx/requirements/private.txt
```

If the sources are left in `/openedx/requirements`, the files can be edited "in place" to see changes. Also, duplication can be avoided and some Kbs can be saved by leaving source files where they are.

regisb: Editing would have to take place inside the container, right? Is that a likely scenario?

As mentioned by email, we want to allow developers to mount edx-platform inside the container, so it makes sense to keep the source outside of the edx-platform folder; pretty much like we would if we had a virtualenv.

README.md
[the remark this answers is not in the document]
regisb: > I see this as an acceptable nuisance.

I agree.

openedx/Dockerfile
silviot: Sorry, I did not comunicate my vision very well.
Plus, it was based on a misconception I had about docker volumes.

I thought volumes and bind mounts would, on creation, always be pre-filled with the contents of the image.
But TIL this is not the case, and that only happens if you have an unnamed volume.

But still, I stand by my proposed change. I'll try to better state my point.

These are the steps that I think will be needed to work on an xblock:

* When building an image, the desired xblock is cloned into the directory
 `openedx/requiremets`, and referenced from the file `private.txt` in the same
 directory as editable, like this: `-e /openedx/requirements/problem-builder`.

* The image is then built and pushed, and maybe also used in production

* A bug pops up, and someone needs to fix it.

* dev pulls the docker images, _copies_ the contents of `/openedx/requirements` inside
  the image to the local directory `./openedx/requirements` and starts docker compose
  with an option that mounts `./openedx/requirements` to `/openedx/requirements`.

* dev makes changes to the xblock, commits and pushes those changes to some repo

* someone (maybe the dev, maybe a CI job, maybe an operator) builds the docker images
  again and pushes the bug fix to production.

Something similar can happen wrt a theme:

* Pull the relevant docker images that bundle the theme
  (this also gives you an easy way to test changes)

* Copy the themes directory (`/openedx/themes`) from the image to some local dir
  (maybe using the same `./openedx/themes` dir that is used for docker image creation)

* Start the docker compose using a compose file that specifies an additional bind mount
  that will mount `data/themes-src` at `/openedx/themes` inside the container

* Edit the copied theme, commit and push

* Rebuild the docker image with the updated theme, and push the new image version


The main risk that I see with this approach is that some automated script might erase
files the user worked on but did not commit/push.

For this reason I think the automated script should always require the user to delete
everything from a destination directory, before attempting to copy data from a docker
image to some local directory.

What do you think in general about this approach?

regisb: Let's keep the themes question for later :sweat_smile: Your scenario makes sense for private requirements. Can we agree that:

1. We get rid of `-e` and `--src` for private requirements? We would then have: `pip install -r /openedx/requirements/private.txt` The developers are then free to add `-e` flags inside the requirements.txt file.
2. We keep the `--src` flag for requirements listed in `edx-platform/requirements`

?

frob:
Why all the manual work? I read the proposed process and I see a lot of "copy from image to host" and "run a bind mount for development" why not just run with a bind mount?

I am starting to think that some of this is fighting against the single image that openedx-docker is built on. Just a thought, why not build this functionality into a new layer? That would leave the `regis/openedx:hawthorn` image alone and we could build the custom stuff into a docker image that sits on top. Then we could put the theme/requirements stuff into a new image. This should also allow for quicker development as we could all rely on using the base image.

Thoughts?

regisb:
@frob the discussion on this PR is the continuation of a conversation we had by email, so some pieces are missing. Sorry about that! 

The idea is that the openedx image should be ready to run very quickly once it was downloaded or built. Currently, the static assets generation step takes a very long time. Hence the idea of packaging assets inside the image. Assets depend on themes, so if we want to package assets we need to package themes as well.

We could package the assets inside a dedicated docker image, using for instance [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/). But then, my impression is that it would introduce a lot of complexity for little gain.

We must not forget that static assets need to be shared with the nginx container, because assets are served by nginx. This is why we used to have bind-mounts shared between nginx and the LMS/CMS containers.

The problem with bind-mounts is that as soon as we mount a volume inside a docker image, we lose the content of that folder inside the docker image. This happens unless the volume was just created, in which case the image folder content get copied to the volume. Thus, to re-generate assets at every build, we would need to destroy and re-create the bind-mount volume every time. 

All in all, it seems to me that the `docker-compose run lms cp...` command is the most straightforward way to get nginx to serve the static assets.

Does this clarify the goal of this PR?

frob:
It clarifies the goal, but wont we run into the same problem with all the configuration? I don't see should we wouldn't have this issue with anything in a bind mount.

The other issue I have with this is, how does one edit/develop the theme? It looks like after this change users will have to run the container in order to edit the theme.

regisb:
> It clarifies the goal, but wont we run into the same problem with all the configuration? I don't see should we wouldn't have this issue with anything in a bind mount.

I don't quite understand your question. Maybe you can give a practical example to illustrate your point? Just to be on the same page: we agree that volume mounts occur at runtime, but not at build time, right? I.e: it's not possible to mount a volume inside a container while it is being built.

> The other issue I have with this is, how does one edit/develop the theme? It looks like after this change users will have to run the container in order to edit the theme.

One of the goals of this PR is also to make it easier for developers to customize themes. Currently, in the master branch, theme modification requires the developer to recompile the assets, or to manually launch a `watch_assets` paver command. We want to create a dedicated `watch_assets` command for quick iteration, so that developers do not have to re-build the image at every modification. I haven't managed (or tried) to do so for now, which is one of the reasons this PR wasn't merged yet :) So there's still work to be done!

regisb:
@silviot @frob I think you'll both be happy with my latest changes. The previous assets command were a pain to maintain, so I ended up creating a convenient `openedx-assets` script which is great for watching theme changes. Compared to the previous approach, recompiling theme assets is pretty fast. I also added `lms/cms-runserver` commands for running a debugging server locally. README instructions are clarified as well.

I think I'm going to merge this pretty soon to avoid a dragging PR that accumulates too much stuff. As always, I'll be happy to hear your thoughts, even after the merge :) thanks!

Issue #87, opened 2018-11-06
Migration from openedX Native


kbitr:
Hey, I have a customized openedX Native version up and running on an AWS EC2 instance. Now I want to migrate to the docker version provided.

Any suggestions how to do this in the first place?

regisb:
Basically, all you need to do is to migrate the data in MySQL and Mongodb. Did you make important changes to your platform? Do you have some experience with docker and server administration?

kbitr:
Hey @regisb,

some changes were made to the platform which weren't documented, but I see some changes in the git repos of edx-platform eg. 

Server administration yes, docker so-so.

regisb:
# Mysql data migration

1. Dump the data from your existing MySQL `edxapp` database:

    mysqldump -u root --password --host yourmysqlhost  edxapp > /tmp/dump.sql

2. Move this dump to a folder shared with the LMS container and run the LMS container:

    mv /tmp/dump.sql /path/to/openedx-docker/data/lms/
    make lms

3. Insert the dumped data in the `openedx` database:

    mysql --user=root --password openedx < /openedx/data/lms/dump.sql

You will need to root password to the database, which is indicated in openedx-docker/config/config.json

# Mongodb data migration

Similarly, you have to transfer the `cs_comments_service` and `edxapp` mongodb databases to `cs_comments_service` and `openedx`, respectively.

Dump databases:

    mongodump --host=yourmongodbhost --out=/path/to/openedx-docker/data/mongodb/dump

Connect to mongodb container:

    docker-compose exec mongodb bash

Import data:

    mongorestore /data/db/dump

Launch a mongodb shell and rename the `edxapp` collection to `openedx` (as per [this stackoverflow answer](https://stackoverflow.com/a/32494587]):

```
$ mongo
db.copyDatabase("edxapp", "openedx");
use edxapp;
db.dropDatabase();
```

And you *should* be fine. Please try out these recommendations and come back with the result of your attempt. It would be great if others could benefit from your experience.

kbitr:
I am trying and will write a Hoot, when I'm finished completely.

- mysql dump seems to be successful
- where is the password for the mongoldb located?
- what is about all the resources like /edxapp/data or themes?

regisb:
> where is the password for the mongoldb located?

Do you mean the password for Mongodb in your current native install? It should be in `lms.auth.json`, in the `DOC_STORE_CONFIG` entry.

> what is about all the resources like /edxapp/data or themes?

[...]

PR #107, opened 2018-12-18
WIP: Add backup and restore commands to the makefile

frob:
After reading #87 I realized that there seemed to be no non-filesystem based backup provisions. This MR adds them.

## Adds the Makefile commands:

 - `backup-lms`
 - `backup-cms`
 - `restore-lms`
 - `restore-cms`

The restore commands allow users to specify the `DUMP` used for the restoration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,8 @@ post_configure: $(post_configure_targets)
 
 ##################### Database
 
+DOCKERIZEWAIT = dockerize -wait tcp://mysql:3306 -timeout 20s
+
 databases: provision-databases migrate provision-oauth2 ## Bootstrap databases
 
 provision-databases: ## Create necessary databases and users
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ reindex-courses: ## Refresh course index so they can be found in the LMS search 
 
 ##################### Backup Utilities
 
-ISO_NOW=$(shell date -u "+%Y-%m-%d--%H:%M:%S%z--%Z")
+ISO_NOW=$(shell date -u "+%Y-%m-%dT%H:%M:%S%z")
 backup-lms: ## Backup the mysql database used for the lms.
 	@echo "Exporting mysql database..."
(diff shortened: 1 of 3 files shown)


frob:
Documentation is coming.

frob:
Okay, documentations is done. I have rebased off regisb/master I am doing some final testing before removing WIP.

regisb:
This is a useful feature. I like how the `backup` and `restore/restore-dangerous` commands work together. My feeling is that the bash commands are a bit brittle, and too difficult to understand. Moreover, the backup commands only work for the openedx database, and not for the xqueue and the notes databases. But this PR is a good starting point. I'll have some time to work on this next week, during the Christmas holidays. In particular, I'd suggest renaming the make commands to something like `backup-openedx-sql` and `backup-openedx-mongodb`, since the lms and the cms essentially share the same database. Are you ok with that?

frob:
Good to know [about the other databases]. I was basing this off of the issue #87 so I thought those where the only two DBs that need the backup feature. I agree with you about the bash commands but I cannot think of a good way to do it in make and breaking it into shell scripts seemed a bit obfuscating.

The main difficulty is the mongp db backup. It would be easier if the mongodump utility was on the lms/cms images, but its not and I like to avoid adding dependencies for utilities. Another option to my approach is to just use the base ubuntu image and add everything we need in a single inline docker command; instead of relying on docker-compose.

frob:
I am looking through the services, I see these possibly needing backups:

- [x] lms
- [x] cms
- [ ] xqueue
- [ ] notes
- [ ] forum

did I miss anything?

regisb:
@frob I don't know if you've resumed work on this, but I think it's better if you wait a little until the release of tutor v3. A lot of things are going to change, and you may have to throw away most of the things you did.

frob:
Nope, I have been preoccupied. Thanks for the heads-up as I was going to
get back at the end of the week.

regisb:
Making backups is not a tutor command, but there are instructions for it: https://docs.tutor.overhang.io/local.html?#backups-migrating-to-a-different-server
I believe this is sufficient, for now. Please open an issue if you think otherwise.

PR #108, opened 2018-12-21
WIP: Added the docker namespacing functionality.

frob:
This feature branch adds the ability to add custom namespacing to the docker images that are built with this project. 

## API changes

This removes the existing variable `OPENEDX_DOCKER_IMAGE` and replaces it with individual pieces that allow for finer control. The new variables are also set to work with the makefile so `make build` (and other docker image commands) will continue to work.

The new variables are:

```
OPENEDX_DOCKER_NAMESPACE
OPENEDX_DOCKER_IMAGE_PREFIX
OPENEDX_DOCKER_TAG
```

These have defaults values that are backward compatible with the current default docker image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+- 2018-12-21 [Feature] Add fine-grain control of docker image namespace
 - 2018-12-02 [Feature] Download extra locales from [openedx-i18n](https://github.com/regisb/openedx-i18n/) to the Open edX Docker image
 - 2018-11-28 [Feature] Easily change openedx docker image
 - 2018-11-28 [Feature] Enable comprehensive theming!
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ USERID ?= $$(id -u)
 EDX_PLATFORM_SETTINGS ?= universal.production
 DOCKER_COMPOSE = docker-compose -f docker-compose.yml
 -include $(PWD)/config/Makefile.env
+-include $(PWD)/.env
+
+OPENEDX_DOCKER_NAMESPACE ?= regis
+OPENEDX_DOCKER_IMAGE_PREFIX ?= ''
+OPENEDX_DOCKER_TAG ?= hawthorn
 
 post_configure_targets =
 ifneq ($(DISABLE_STATS), 1)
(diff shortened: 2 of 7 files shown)


frob:
Agh, my computer died as I was finishing this PR. Back to it now.

frob:
Okay, tech issues aside I force pushed the branch and it should be ready for PR.

regisb:
Hi @frob! Thanks for creating this PR! The changes you propose are rather substantial and make the docker-compose* files much harder to read. I see the point of using a different openedx image, but do you think it's a useful feature to run different images for notes/xqueue? I must admit I'm not a big fan of the change, but maybe I fail to see your use case here. Maybe you can tell me more about the problem you are trying to solve?

frob:
@regisb sorry, I took Christmas off completely --so now I am getting back to this. 

It looks like notes and xqueue are their own codebases and thus the possibility could exist that they need their own containers. At the very least I feel a custom namespace is required as forcing anyone using this repo as their base will mean forcing them to use your regis namespaced images as their base and for some clients that simply isn't going to work. 

I defininatly see the issue with the names being unreadable in the compose files. Maybe the assebly of the namespaces could be done in the Makefile instead and that would leave a single env variable used for the image:

```
  image: $(XQUEUEDOCKERIMAGE)
```

or, to make the default self documented:

```
  image: $(XQUEUEDOCKERIMAGE:-regis/openedx-xqueue:hawthorn)
```

From the looks of it, much work has been done so I will have to refactor whatever I do anyway.

regisb:
> forcing anyone using this repo as their base will mean forcing them to use your regis namespaced images as their base and for some clients that simply isn't going to work.

Why is that? I was considering of renaming the docker image to a more "professional-sounding" repo that would correspond to my company. Would that work?

frob:
Some projects/clients require more control over the stack. Also, imagine needing another patch or customization for a one off. Or not a requirement to not mix images. For 85% of use-cases, yes, an official sounding namespace would be enough, but for the more difficult 15%, which I could see myself having as clients, this would be a nice addition.

Let see if I can make this less of an encumbrance by moving the assembly of the image names into the Makefile.

frob:
Wow, so much has changed I think we need some discussion before this moves forward again.

Closing for now as this branch is now woefully out of date. Discussion can continue in #122

PR #110, opened 2018-12-28
Fix SSL issue with the custom readthedocs domain

OmarIthawi:
```
$ curl https://docs.tutor.overhang.io/
curl: (51) SSL: certificate subject name (*.readthedocs.io)
does not match target host name 'docs.tutor.overhang.io'
```

Fixing it with readthedocs is one option, the other one is to fix the link in README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,7 @@ Quickstart
 Documentation
 -------------
 
-Extensive documentation is available online: https://docs.tutor.overhang.io/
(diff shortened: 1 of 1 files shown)


regisb:
Hi Omar! Thanks for warning me about this! I'm going to be away from a computer all day, so it's great that you opened a pull request I can merge in one click. Would you mind modifying your pr such that the README points to http://docs.tutor.overhang.io? Thanks!

regisb:
Solved! Thanks again :)

PR #138, opened 2019-01-26
fix VIDEO_IMAGE_SETTINGS and VIDEO_TRANSCRIPTS_SETTINGS path when export a course

dannielarriola:
The srt upload well, but when we export a course this fails. It's because the VIDEO_TRANSCRIPTS_SETTINGS var in common.py. This took the MEDIA_ROOT var from his own file (/edx/var/edxapp/media/) and upload the srt file on it. But this path is not a volume, then it`s only exists in one service(cms_worker) and when export ocurrs this search the file on cms service. The same with VIDEO_IMAGE_SETTINGS.
This fix redefine those vars on production.py with the correct path (/openedx/data/uploads/video-transcripts)

### DIFF
--- a/deploy/local/templates/docker-compose.yml
+++ b/deploy/local/templates/docker-compose.yml
@@ -52,6 +52,8 @@ services:
     volumes:
       - ../env/nginx:/etc/nginx/conf.d/
       - ../../data/openedx:/var/www/openedx:ro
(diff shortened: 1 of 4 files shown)


regisb:
Can you please confirm that my changes work for you?

PR #149, opened 2019-02-08
Improve import in openedx config settings

tonytan4ever:
This is to improve in situations where we need to override default production/development settings.


regisb:
@tonytan4ever as discussed on #150, I think this is a relevant improvement. Would you mind to make the same changes in the CMS settings?

PR #157, opened 2019-02-10
Improve import in openedx config settings

tonytan4ever:
This is to improve in situations where we need to override default production/development settings.

Signed-off-by: tonytan4ever <tonytan198211@gmail.com>


regisb:
:+1: !

PR #161, opened 2019-02-11
Align `docker-compose logs` invocation to the `docker-compose up` call

silviot:


regisb:
This makes total sense. Thanks!

PR #162, opened 2019-02-11
Change import: `?ms.env.devstack` → `?ms.envs.devstack`

silviot:
I might be missing something, but without this change I was unable to run tutor.


regisb:
@silviot Absolutely, this is a bug that was recently introduced in https://github.com/regisb/tutor/pull/157/
The `development.py` need to be changed too, and this is kind of urgent, so I'll do the change myself. Thanks a lot for the heads up!

PR #197, opened 2019-04-16
Tell openedx gunicorn to fork workers based on CPU count

silviot:
Tutor currently runs a single worker `gunicorn` process.

This PR makes the number of workers dependent on the number of cores: `(2 x $num_cores) + 1`, as recommended in the [gunicorn documentation](http://docs.gunicorn.org/en/stable/design.html#how-many-workers).

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -89,6 +89,9 @@ ENV SETTINGS tutor.production
 # Entrypoint will fix permissions of all files and run commands as openedx
 ENTRYPOINT ["docker-entrypoint.sh"]
 
+# Copy gunicorn settings file
+COPY gunicorn_conf.py /openedx/gunicorn_conf.py
+
 # Run server
 EXPOSE 8000
-CMD gunicorn --name ${SERVICE_VARIANT} --bind=0.0.0.0:8000 --max-requests=1000 ${SERVICE_VARIANT}.wsgi:application
+CMD gunicorn -c file:/openedx/gunicorn_conf.py --name ${SERVICE_VARIANT} --bind=0.0.0.0:8000 --max-requests=1000 ${SERVICE_VARIANT}.wsgi:application
--- a/tutor/templates/build/openedx/gunicorn_conf.py
+++ b/tutor/templates/build/openedx/gunicorn_conf.py
@@ -1,0 +1,5 @@
+import multiprocessing
+
+# The recommended number of workers is twice the
+# CPU count plus one
+workers = (multiprocessing.cpu_count()-1) * 2 + 2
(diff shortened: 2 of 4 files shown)


silviot:
`notes` and `xqueue` will need a similar update. I can work on it, but IMO this PR can already be merged leaving them as TODOs.

tutor/templates/build/openedx/gunicorn_conf.py
```diff
@@ -0,0 +1,5 @@
+import multiprocessing
+
+# The recommended number of workers is twice the
+# CPU count plus one
+workers = (multiprocessing.cpu_count()-1) * 2 + 2
```
regisb: Shouldn't we write instead:

    workers = multiprocessing.cpu_count() * 2 + 1

?

tutor/templates/build/openedx/Dockerfile
```diff
@@ -89,6 +89,9 @@ ENV SETTINGS tutor.production
 # Entrypoint will fix permissions of all files and run commands as openedx
 ENTRYPOINT ["docker-entrypoint.sh"]
 
+# Copy gunicorn settings file
+COPY gunicorn_conf.py /openedx/gunicorn_conf.py
+
 # Run server
 EXPOSE 8000
-CMD gunicorn --name ${SERVICE_VARIANT} --bind=0.0.0.0:8000 --max-requests=1000 ${SERVICE_VARIANT}.wsgi:application
+CMD gunicorn -c file:/openedx/gunicorn_conf.py --name ${SERVICE_VARIANT} --bind=0.0.0.0:8000 --max-requests=1000 ${SERVICE_VARIANT}.wsgi:application
```
regisb: It's interesting, I didn't know it was possible to use a python configuration file like this. http://docs.gunicorn.org/en/stable/configure.html#configuration-file

I like the fact that we keep the `bind` and `max-requests` parameters out of the configuration file: I think it makes it easier to find the information about the exposed port.

regisb:
thanks @silviot, it's great that you spotted that out! It's crazy that gunicorn works with just a single worker by default: I was absolutely sure that it adapted to the number of cores out of the box.

IMHO xqueue and notes don't need to handle too many requests. I think it's reasonable to set `--workers=3` in those apps. Maybe you can do that in this PR?

tutor/templates/build/openedx/gunicorn_conf.py
silviot: Oh, yes, we should!

/me hides in shame ;-)

tutor/templates/build/openedx/Dockerfile
silviot: In my opinion a good spot for the `bind` and `max-requests` parameters is the `GUNICORN_CMD_ARGS` environment variable. But we can't do that with the current gunicorn version, so for now I agree they look good in the `CMD` directive.

silviot:
AFAIC this should be good to merge now.

regisb:
thanks again for the contribution!

PR #198, opened 2019-04-17
Make sure tutor asks for permission before a possibly destructive action

silviot:
We were bit by this. Fortunately we had everything under revision control, so we didn't lose any work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .*.swp
 !.gitignore
 TODO
+__pycache__
 
 /tutor.spec
 /build/tutor
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -77,13 +77,13 @@ def load(root):
     load_defaults(config)
 
     if not env.is_up_to_date(root):
-        click.echo(fmt.alert(
+        click.confirm(fmt.alert(
             "The current environment stored at {} is not up-to-date: it is at "
             "v{} while the 'tutor' binary is at v{}. The environment will be "
             "upgraded now. Any change you might have made will be overwritten.".format(
                 env.base_dir(root), env.version(root), __version__
             )
-        ))
+        ), abort=True)
         should_update_env = True
 
     if should_update_env:
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -1,5 +1,6 @@
 import json
 import os
+import sys
 
 import click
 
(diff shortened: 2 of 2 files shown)


regisb:
> We were bit by this

Can you elaborate? (even though it's recent and probably painful)

The environment is also automatically re-generated on `tutor config save`. Isn't that an issue?

silviot:
> Can you elaborate? (even though it's recent and probably painful)

Sure! We had our `git` seatbelts on, so it was just a bump: not so painful for us.

For a project we created a repository with code that downloads tutor and sets it up.

In the repository there's also the tutor project root: we changed some things here and there. A customized settings file in `env/apps/openedx/settings/lms/`, changes to `docker-compose.yml` etc.

We noticed the file `env/version`, failed to assess its purpose, and removed it. Next thing we knew tutor had reverted our customisations.

We saw the warning 
```
⚠️  The current environment stored at /some/path/.local/share/tutor/env is not up-to-date: it is at v0 while the 'tutor' binary is at v3.3.3. The environment will be upgraded now. Any change you might have made will be overwritten.
```
but we were not asked our opinion about that action. We feel like we should really be asked before destructive actions.

regisb:
Makes sense. And I guess you avoid running `tutor config save` after making manual changes, right?

The interactive part makes me a bit anxious, though. It's quite possible that the `load()` function gets called in an environment where there is no human interaction (e.g: deployment scripts). In such as case, it's probably better to simply crash with a big fat red alert.

I need to think a little about what's the best behaviour here. On the top of my head, there are many possible solutions:

- detect if the current shell supports interaction (is that even possible? does click do that out of the box?)
- create a `tutor config upgradeenv` command
- create a `TUTOR_OVERWRITE_ENV` environment variable
...

silviot:
I see your concern. My favourite option is #1: detect if we're running in a terminal, and aborting if we're not. I'll have a look at if/how it's possible.

silviot:
How about [checking if `sys.stdin` is a tty](https://stackoverflow.com/a/4213118/41387)?

```python
import sys, os
if os.isatty(sys.stdin.fileno()):
    # ask the question
else:
    # abort execution
```

regisb:
Yes, that could be an option. Isn't that what click does already in the background? Hmmm, apparently not. Click has a `isatty` function but it's only used to detect whether colors are supported.

silviot:
I amended the PR including my proposal from above.

I also sneaked in the addition of `__pycache__` to `.gitignore`: I expect it to not be controversial.

regisb:
I will merge this now and add some minor modifications. In particular, I don't think anymore that a `TUTOR_OVERWRITE_ENV` environment variable is necessary (sorry about that).

Thanks again!

Issue #127, opened 2019-01-17
Streamline Containers


frob:
## Steps to reproduce

run `make build`

## Unexpected behavior

Building images takes a very long time and leaves us with very large containers --over a 2GB.

## Proposed solution

If we streamline the build to use a common base image layer for all the edx-platform based images then that will reduce the pull time and the only thing that will build will be the custom layers. I also think more work can be done by using alpine based images instead of ubuntu:16.04

This will likely require more work to get it working, but once working this should result in a much faster build and a much smaller image.

frob:
> First of all, "faster" and "lighter" are two completely different issues. I'd dare to say, they are even incompatible.

I agree they are different, but I don't agree they are incompatible. Also, I think the faster part needs more clarification. I think the builds can be made faster. I also think that we can make the download faster as well, but we will see.

About those benchmarks. The numbers are great starting points, but I am not sure they mean that the image is just going to be big. Looking at the docker python images:

```
python                                2.7                 bfa54426aeda        9 days ago          912MB
python                                3.6                 55fb8aca33df        9 days ago          922MB
python                                3.6-alpine          1837080c5e87        3 weeks ago         74.4MB
python                                2.7-alpine          66c225e226f9        4 weeks ago         58.3MB
```

Looking at these images as a baseline of improvements that can be made, aside from the static assets, I think we can shrink the image down considerably. 

I also think there is some low hanging fruit. I know that edx-platform is built (so far) to only work with Ubuntu 16.04, but there are other images that we can start with.

- **Forum** : ruby based: ubuntu:18.04
- **Notes** : python based: ubuntu:18.04
- **XQueue** : python based: ubuntu:18.04

I am less worried about android because that is used for development and the image should never be deployed.

As I think through the process for streamlining all this it begs a question. What about testing. That might be a first step to make sure something obscure isn't broken by these changes. Opened #129 to discuss testing.

frob:
@regisb I was going to see if I can get some of this working. I know you said some changes where coming. Would the changes impact work done here?

regisb:
This issue has not received any attention for a long time. I don't really see a clear way forward, so I'm going to close this. Feel free to re-open or comment if you have fresh ideas.

PR #208, opened 2019-04-30
Do not require bash to be installed in docker containers.

silviot:
Tutor is not using any bash-specific feature.
This PR changes (almost) all invocations to /bin/bash replacing them with
`/bin/sh`. It allows building a docker image that does not bundle bash in it.

The file `k8s.py` still has a bash invocation, to make sure this change does not
impact functionality for any user.

### DIFF
--- a/tutor/k8s.py
+++ b/tutor/k8s.py
@@ -66,7 +66,7 @@ def delete(yes):
 )
 @opts.root
 def databases(root):
-    scripts.migrate(root, run_bash)
+    scripts.migrate(root, run_sh)
 
 
 @click.command(help="Create an Open edX user and interactively set their password")
@@ -76,14 +76,14 @@ def databases(root):
 @click.argument("name")
 @click.argument("email")
 def createuser(root, superuser, staff, name, email):
-    scripts.create_user(root, run_bash, superuser, staff, name, email)
+    scripts.create_user(root, run_sh, superuser, staff, name, email)
 
 
 @click.command(help="Import the demo course")
 @opts.root
 def importdemocourse(root):
     click.echo(fmt.info("Importing demo course"))
-    scripts.import_demo_course(root, run_bash)
+    scripts.import_demo_course(root, run_sh)
     click.echo(fmt.info("Re-indexing courses"))
     indexcourses.callback(root)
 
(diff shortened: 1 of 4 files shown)


regisb:
Hey @silviot! I find bash very convenient to debug containers and to develop features on edx-platform. What do we gain by removing it? I guess it makes the images lighter and faster to build, but by how much?

silviot:
This PR is not about *removing* bash: It's about stating minimum requirements. Since `tutor` only uses standard POSIX shell features, all scripts can run with any POSIX compliant shell. The image still has bash with all its interactive goodies in place. The change is in the fact that the `tutor` python code currently *requires* bash, even if it doesn't need it.

Alpine linux for instance does not ship with `bash` by default, but with the lighter `ash`. I'm building an openedx image based on Alpine, and by shaving little things here and there I managed to get it to [half the size](https://hub.docker.com/r/silviot/openedx-alpine/tags) of the [current tutor image](https://hub.docker.com/r/regis/openedx/tags). I understand small docker image sizes are not a priority of the tutor project, but I thought a PR like this would more correctly state its dependencies.

regisb:
> This PR is not about removing bash: It's about stating minimum requirements.

Right! It makes total sense then. Let me test this before merging.

> I'm building an openedx image based on Alpine, and by shaving little things here and there I managed to get it to half the size of the current tutor image.

This is very interesting. Can you share the dockerfile for this image? It makes sense to have a tutor API that is compatible both with the Ubuntu and the Alpine images, if little efforts are necessary.

> I understand small docker image sizes are not a priority of the tutor project

Indeed, they are not, in the sense that there are more pressing features for me to work on. But then everybody -- including me -- would be very happy with lighter, faster-to-build images (see #127).

silviot:
@regisb 
> This is very interesting. Can you share the dockerfile for this image?
Unfortunately not, because it's not built using Docker ;).
I'm using buildah to have better control over when layers are created and to mount directories during build steps.
[Here's the current version](https://github.com/Abstract-Tech/tutor/tree/azure-testing-alpine/tutor/templates/build/openedx-alpine).
I'm [testing it on azure here](https://dev.azure.com/abstract-technology/tutor/_build?definitionId=2).
The image generation is still manual. I need to add a job to the Azure pipeline to build the images.

Issue #199, opened 2019-04-17
Host '172.18.0.9' is not allowed to connect to this MySQL server


natea:
## Steps to reproduce

Run `tutor local quickstart`

## Unexpected behavior

I get the following error (password redacted to `xxxxxxxxx`):

```
Nathans-MacBook-Pro-2:tutor nateaune$ tutor local quickstart
==================================================
        Interactive platform configuration
==================================================
Your website domain name for students (LMS) [www.myopenedx.com]
Your website domain name for teachers (CMS) [studio.openedx.com]
Your platform name/title [My Open edX]
Your public contact email address [nate@myemaildomain.com]
The default language code for the platform [en]
Activate SSL/TLS certificates for HTTPS access? Important note:this will NOT work in a development environment. [yN]
Activate Student Notes service (https://open.edx.org/features/student-notes)? [yN]
Activate Xqueue for external grader services (https://github.com/edx/xqueue)? [yN]
Configuration saved to /Users/nateaune/Library/Application Support/tutor/config.yml
Environment generated in /Users/nateaune/Library/Application Support/tutor/env
==============================================
        Stopping any existing platform
==============================================
docker-compose -f /Users/nateaune/Library/Application Support/tutor/env/local/docker-compose.yml --project-name tutor_local rm --stop --force

Going to remove tutor_local_elasticsearch_1, tutor_local_rabbitmq_1, tutor_local_mongodb_1, tutor_local_memcached_1, tutor_local_smtp_1, tutor_local_mysql_1
Removing tutor_local_elasticsearch_1 ... done
Removing tutor_local_rabbitmq_1      ... done
Removing tutor_local_mongodb_1       ... done
Removing tutor_local_memcached_1     ... done
Removing tutor_local_smtp_1          ... done
Removing tutor_local_mysql_1         ... done
================================================
        Database creation and migrations
================================================
Creating lms/cms databases...
docker-compose -f /Users/nateaune/Library/Application Support/tutor/env/local/docker-compose.yml --project-name tutor_local run --rm lms bash -e -c dockerize -wait tcp://mysql:3306 -timeout 20s
mysql -u root --password="xxxxxxx" --host "mysql" -e 'CREATE DATABASE IF NOT EXISTS openedx;'
mysql -u root --password="xxxxxxx" --host "mysql" -e 'GRANT ALL ON openedx.* TO "openedx"@"%" IDENTIFIED BY "xxxxxxx";'
Creating tutor_local_mysql_1         ... done
[...]

Issue #160, opened 2019-02-11
ERROR 1130 (HY000): Host '172.19.0.9' is not allowed to connect to this MySQL server


a20180502:
PLEASE PLEASE PLEASE READ THIS AND FOLLOW THE INSTRUCTIONS WHEN OPENING A NEW ISSUE

## Steps to reproduce

Include the exact command that you are running and that is causing an error. In case of a web error, include the URL of the page that is causing the error.

## Unexpected behavior

Include the full, exact output from the command that is causing your issue. Also include relevant error logs; for instance, to debug the LMS provide the output of `tutor local logs lms --tail=100`

## Additional info (IMPORTANT)

Provide the output of `tutor --version`.

regisb:
This is a duplicate of #156

Issue #136, opened 2019-01-25
Starting local_mysql takes forever


kbitr:
## Steps to reproduce

Have an older version of tutor. git pull. make all. wait.

## Unexpected behavior

```docker-compose up -d mysql
Starting local_mysql_1_3aca34de5483 ... done
until docker-compose logs mysql | grep "MySQL init process done. Ready for start up."; do \
        sleep 1;\
done
```
takes forever

## Additional info (IMPORTANT)

Related to commit: c42b747e3538eb8f0ead1534c381933afdc0b87e

## intermediate solution

just don't wait - I assume this happens, because one is not supposed to run make all again, once everything is up?

regisb:
Hi @kbitr! Waiting for mysql to be ready is key, and prior to introducing this change there were many  bugs that were difficult to solve. E.g: #3, #75, #120.
The problem is that on the first boot, mysql needs to initialize the root user with the right password. If it doesn't, databases cannot be created and migrations do not run. Basically, it's a race condition problem.

To better understand your problem, please answer the following questions:

1. What does it mean when you say "forever"? Is it just a long time, or do you mean that the command never ends?
2. If the command ends, how long does it take? (approximately)
3. If the command does not end, what is the full output of `docker-compole logs mysql`?

kbitr:
Hey, I have to do this on my production server, so I did not wait long this time to have no downtime. But since the output is the following, I assume that nothing will happen anyway.
```
2019-01-27 13:09:20 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
2019-01-27 13:09:20 0 [Note] mysqld (mysqld 5.6.36) starting as process 1 ...
```

regisb:
@kbitr you're right: the message is not printed once the mysql db has already been initialized. I just commited a fix. Please confirm that it works for you, as usual :)

Issue #120, opened 2019-01-08
Failed to connect to mysql 


efatuarte95:
## Steps to reproduce

I have just cloned the repo following the steps included in the guide:
```
git clone https://github.com/regisb/tutor
cd tutor/deploy/local
make all
```
## Unexpected behavior

It is supposed that when i exec ```make all```, the docker images are pulled and run immediately but this error appears:
```
ERROR 1045 (28000): Access denied for user 'root'@'172.19.0.9' (using password: YES)
Waiting for MySQL database to be ready...
MySQL: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'172.19.0.9' (using password: YES)
Waiting for MySQL database to be ready...
MySQL: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'172.19.0.9' (using password: YES)
Waiting for MySQL database to be ready...
MySQL: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'172.19.0.9' (using password: YES)
Waiting for MySQL database to be ready...
```

## Additional info (IMPORTANT)

I have just watched a similar issue (#75 & #51) here where the following commands are recommended:
```
make stop
make databases
```
But when i run them instead of solving the problem the same error appears again...

Any help would be appreciated

regisb:
Hi @efatuarte95! Can you please paste here the complete output of `make databases`?

efatuarte95:
I finally solved the issue using
```docker-compose down``` to drop all my docker containers and images and then with
```docker system prune --volumes``` I deleted my volumes.
Then I run make all again and it worked! I don't know really what happened but now everything it's fine.

regisb:
Awesome! Good job solving that on your own ;)

PR #209, opened 2019-04-30
Fix database initialization

silviot:
Currently, for a newly created environment, `tutor local databases` will error out on the first run and succeed on the second one.

This is because the `mysql` container is started, and after it initialises its directories it's shut down.

So on the first run (when `os.path.exists(mysql_data_path)` is `False`) by the time code in `scripts.migrate` runs, the mysql container is stopped.

The PR also includes changes to not require the `mysql` client binary to be installed alongside openedx: it uses the client from the `mysql` container instead.

This PR is based on #208

### DIFF
--- a/tutor/k8s.py
+++ b/tutor/k8s.py
@@ -66,7 +66,7 @@ def delete(yes):
 )
 @opts.root
 def databases(root):
-    scripts.migrate(root, run_bash)
+    scripts.migrate(root, run_sh)
 
 
 @click.command(help="Create an Open edX user and interactively set their password")
@@ -76,14 +76,14 @@ def databases(root):
 @click.argument("name")
 @click.argument("email")
 def createuser(root, superuser, staff, name, email):
-    scripts.create_user(root, run_bash, superuser, staff, name, email)
+    scripts.create_user(root, run_sh, superuser, staff, name, email)
 
 
 @click.command(help="Import the demo course")
 @opts.root
 def importdemocourse(root):
     click.echo(fmt.info("Importing demo course"))
-    scripts.import_demo_course(root, run_bash)
+    scripts.import_demo_course(root, run_sh)
     click.echo(fmt.info("Re-indexing courses"))
     indexcourses.callback(root)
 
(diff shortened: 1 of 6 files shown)


regisb:
> Currently, for a newly created environment, tutor local databases will error out on the first run and succeed on the second one.

This is not what I observe. If you are facing this problem, can you please open a new issue?

There have been many issues related to mysql initialization (#199, #160, #136, #120, #75, #51, #34, #3), so this is a very sensitive piece of code. In particular, replacing `docker-compose run` by `docker-compose exec` has many side effects. In the particular case of mysql migrations, it implies that the mysql container is already running, which is not always the case. For instance, on the first `tutor local quickstart`, the mysql container is not running.

If you describe your issue more precisely, perhaps we can try to find a more appropriate solution?

silviot:
I realized I was missing [this commit](https://github.com/regisb/tutor/commit/92d7e7c4186ac3790d8420a81182dd104c771d51): that's the issue I was running into.

I understand the current solution (invoking `docker-compose run` on the `lms` container) works because of the dependency of `lms` over the `mysql` container [specified here](https://github.com/regisb/tutor/blob/master/tutor/templates/local/docker-compose.yml#L118). There is no mechanism currently in place in the Kubernetes code to make sure mysql is running when running the [`databases`](https://github.com/Abstract-Tech/tutor/blob/855f25f67e0fa9005e21bfc6c5a1a8221b08a9a3/tutor/k8s.py#L69) function, as far as I can tell.

My proposal is to solve that problem by removing [the line that stops mysql](https://github.com/Abstract-Tech/tutor/commit/855f25f67e0fa9005e21bfc6c5a1a8221b08a9a3) and rely on the fact that [mysql had just been started](https://github.com/Abstract-Tech/tutor/blob/855f25f67e0fa9005e21bfc6c5a1a8221b08a9a3/tutor/local.py#L143) if we went through `tutor.local:databases` (which is currently the only path in non-kubernetes code).

I fail to understand the purpose of that line anyway; removing it will also speed up the setup process.

I'll close the PR, since it got stale. My 2 ¢ are:

* The line stopping mysql should go, speeding up the docker-compose database setup process
* The database creation code should not introduce the need to ship the `mysql` client binary inside the openedx image

regisb:
> The line stopping mysql should go, speeding up the docker-compose database setup process

It would make sense, purely from a performance point of view. It's just that I don't like the fact that we exit the `init_mysql` function with an extra running container. I need to think about it.

> The database creation code should not introduce the need to ship the mysql client binary inside the openedx image

This is true. This would be the main benefit of using `mysql exec` vs `run lms`.

PR #205, opened 2019-04-24
Send tracking logs to docker via stdout and give the option to users to save log files locally

silviot:
This PR introduces no changes in behaviour, by default.

It adds an option to persist log files outside docker containers, and documents how to do it.

PR #212, opened 2019-05-05
Persist LMS/CMS logs to disk by default

regisb:
This is the continuation of #205 and #201. The goal is to persist LMS/CMS logs to disk by default.


regisb:
@silviot what do you think? can we merge this?

silviot:
Looks like an improvement to me: everyone is familiar with logs in a file, so having it as default (besides the docker logging) is a step forward.

PR #237, opened 2019-07-23
Fix tutor binary permission error on MacOSX

OmarIthawi:
Otherwise it will not be readable by normal users.


regisb:
Thanks @OmarIthawi!

PR #254, opened 2019-09-30
update django setting module

tuan-nng:
Use correct variable as seen in the code here
https://github.com/edx/edx-platform/blob/master/manage.py#L104

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -102,7 +102,7 @@ RUN mkdir /openedx/data
 
 # service variant is "lms" or "cms"
 ENV SERVICE_VARIANT lms
-ENV SETTINGS tutor.production
+ENV EDX_PLATFORM_SETTINGS tutor.production
 
 {{ patch("openedx-dockerfile") }}
 


regisb:
@tuan-nng Thanks for your interest! However, I'm not sure this change is required, as the `DJANGO_SETTINGS_MODULE` environment variable is already set by the docker entrypoint: https://github.com/overhangio/tutor/blob/master/tutor/templates/build/openedx/bin/docker-entrypoint.sh#L2
I'm going to close this PR. Feel free to re-open if you think this change is nonetheless required.

tuan-nng:
Thanks @regisb. Now I understand the logic there. However, that variable is only available inside the shell session of entry-point. It makes it not obvious when debugging inside the container.
I think we should set it more explicitly inside the dockerfile.

regisb:
Are you proposing that the `DJANGO_SETTINGS_MODULE` environment variable be defined not in the entrypoint, but in the Dockerfile? I believe the reason why it was defined in the entrypoint is that it depends both on the `SETTINGS` and the `SERVICE_VARIANT` environment variables. If I remember correctly, this made it easier to define modular settings.

You can try to make the change, but I suspect there will be far-ranging consequences...

tuan-nng:
Thanks for your explanation. I'll let it as is.

PR #255, opened 2019-10-06
Adds to_json jinja filter

arbrandes:
Rationale: plugins may require the user to set yaml configuration variables that are something other than strings.  Such as, for instance, a custom [XBLOCK_SETTINGS](https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/services.py#L36) dict entry in `lms-env`.  This makes it so they can define it in yaml, and render it with `PLUGIN_XBLOCK_SETTINGS|to_json` where needed.

### DIFF
--- a/tutor/env.py
+++ b/tutor/env.py
@@ -35,6 +35,7 @@ class Renderer:
             environment.filters["common_domain"] = utils.common_domain
             environment.filters["reverse_host"] = utils.reverse_host
             environment.filters["walk_templates"] = walk_templates
+            environment.filters["to_json"] = utils.to_json
             environment.globals["TUTOR_VERSION"] = __version__
 
             cls.ENVIRONMENT_CONFIG = deepcopy(config)
--- a/tutor/utils.py
+++ b/tutor/utils.py
@@ -1,4 +1,5 @@
 import os
+import json
 import random
 import shutil
 import string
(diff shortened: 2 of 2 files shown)


arbrandes:
If there's a better way to have plugins specify custom filters, I'm all ears. :)

regisb:
Hey @arbrandes! Usually I'm not a fan of dict-defined settings. Dict-like settings cannot be easily defined or modified from the command line. That being said, it is sometimes necessary to have a dict as a setting. I had this requirement in the ecommerce plugin, and I used a dirty solution: https://github.com/overhangio/tutor-ecommerce/blob/2194154fb96cabc6b1e16958e0f56fbd9eaeaa7c/tutorecommerce/templates/ecommerce/apps/tutor.py#L80

```
PAYMENT_PROCESSOR_CONFIG = {
    "openedx": json.loads("""{{ ECOMMERCE_PAYMENT_PROCESSORS|tojson(indent=4) }}""")
}
```

You can see that I used the `tojson` [native jinja2 filter](https://jinja.palletsprojects.com/en/2.10.x/templates/#tojson). Actually, I should probably have written:

```
PAYMENT_PROCESSOR_CONFIG = {
    "openedx": json.loads({{ ECOMMERCE_PAYMENT_PROCESSORS|tojson(indent=4) }}
}
```

as json is compatible with Python syntax.

Anyway, I think the `tojson` filter does the same thing as your proposed `to_json` filter, doesn't it?

arbrandes:
Yeah, not a fan of dict settings either, but in the case of the plugin I'm
writing, there's no way around it.

But... huh.  There's a built-in Jinja tojson?  How the hell did I miss
that?  That should work perfectly, thanks! :D

Issue #135, opened 2019-01-23
Follow up - deploy locally 


JeffKatzy:
Thanks for all of the help @regisb.  I was able to make it past that issue, but now am running into another one that does not allow me to deploy locally while linking my code.

## Steps to reproduce
```
env | grep EDX 
EDX_PLATFORM_SETTINGS=tutor.development
EDX_PLATFORM_PATH=/Users/jeff/Documents/jigsaw/software/edx-software/downloaded-edx/edx-platform
```

```
make lms-runserver

docker-compose run -e SETTINGS=tutor.development --volume="$(pwd)/../../build/openedx/themes:/openedx/themes" -e USERID=$(id -u) --volume="/Users/jeff/Documents/jigsaw/software/edx-software/downloaded-edx/edx-platform:/openedx/edx-platform" -p 8000:8000 lms ./manage.py lms runserver 0.0.0.0:8000
Creating local_mongodb_1       ... done
Creating local_smtp_1          ... done
Creating local_rabbitmq_1      ... done
Creating local_elasticsearch_1 ... done
Creating local_mysql_1         ... done
Creating local_memcached_1     ... done
Creating local_forum_1         ... done
creating new user 'openedx' with UID 501
chown: changing ownership of '/openedx/edx-platform/common/static/edx-ui-toolkit/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/fonts': No such file or directory
make: *** [lms-runserver] Error 1
````

I receive the same error message when running `make cms-runserver`

## Things I tried

1. Neither `docker-compose logs` nor `docker container logs -f local_lms_run_9bc186300ca6` provided me with more info.  
2. On one of the launches, saw an error of `Invalid http_host header` so led me to copy over the settings, also as per the documentation 

```
cp -r tutor/deploy/templates/openedx/settings/cms/ downloaded-edx/edx-platform/cms/envs/tutor/
cp -r tutor/deploy/templates/openedx/settings/lms/ downloaded-edx/edx-platform/lms/envs/tutor/
``` 

* Notice that my path to the settings is different than that in the documentation of `config/openedx/tutor/lms/`, I couldn't find files in that path.

## Additional info (IMPORTANT)

```

[tutor (master)*]$ make info
uname -a
[...]

PR #257, opened 2019-10-10
docs: updated local installation guide

adrian-marcelo-gallardo:
While trying to run `tutor` with a local `edx-platform` repository I was getting these errors:
```
chown: changing ownership of '/openedx/edx-platform/common/static/edx-ui-toolkit/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/fonts': No such file or directory
```

After searching for solutions, I've found this issue which suggested installing `npm` dependencies in the `edx-platform` project: https://github.com/overhangio/tutor/issues/135

After installing the dependencies I was able to run the other commands necessary to install python dependencies, call the setup file and compile the assets.

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -64,6 +64,8 @@ Prepare the edx-platform repo
 In order to run a fork of edx-platform, dependencies need to be properly installed and assets compiled in that repo. To do so, run::
 
     export TUTOR_EDX_PLATFORM_PATH=/path/to/edx-platform
+    cd /path/to/edx-platform
+    npm install
     tutor dev run lms pip install --requirement requirements/edx/development.txt
     tutor dev run lms python setup.py install
     tutor dev run lms paver update_assets --settings=tutor.development


regisb:
Hi @adrian-marcelo-gallardo! Indeed, your issue is related to the fact that node dependencies are not installed in your edx-platform repo. However, that step is supposed to be taken care of when you run `tutor dev run lms paver update_assets --settings=tutor.development`, as indicated in the docs. Did you try to run this command?

I'd rather not introduce a dependency on npm on the host and perform all dependency installation inside the container.

adrian-marcelo-gallardo:
Hello @regisb, I remember seeing those error-messages even when running `tutor dev run lms pip install --requirement requirements/edx/development.txt` and `tutor dev run lms python setup.py install` 🤔 
But maybe this is not a primary issue and the installation can continue despite those messages, right?

At the beginning I was getting some fatal-errors that prevented the application from running, and I thought it was because `node_modules` not being present. But then I realized I was using `master` branch instead `ironwood` and switching branch solved the issue.

I might need to reinstall `edx-platform` in the following days, so I can come back with updates about how it goes without manually installing `node_modules`, but feel free to close the issue if you are confident about this :)

regisb:
Indeed, these "chown: changing ownership" messages are not important and can be safely ignored until you install node dependencies ("with `tutor dev run lms paver update_assets --settings=tutor.development`).

I'll close this for now, please re-open when you encounter this issue again.

adrian-marcelo-gallardo:
Hello again! I've tried installing again an `edx-platform` fork, and it was failing because of the issue mentioned above.

The problem is fixed by manually running `npm install` before executing any of the following commands:
```
tutor dev run lms pip install --requirement requirements/edx/development.txt
tutor dev run lms python setup.py install
tutor dev run lms paver update_assets --settings=tutor.development
``` 

So, adding those lines to the doc can work; but, if you think it's better to fix the issue from its root I completely agree too (although I won't have enough time to dig deeper).

Please, let me know what do you think.

regisb:
@adrian-marcelo-gallardo Which command is failing and what is the error log?

adrian-marcelo-gallardo:
In the console output I only see this:
```
chown: changing ownership of '/openedx/edx-platform/common/static/edx-ui-toolkit/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/js': No such file or directory
chown: changing ownership of '/openedx/edx-platform/common/static/edx-pattern-library/fonts': No such file or directory
```
and then this:
image (10)

Either of these 3 commands fail the same way:
```
tutor dev run lms pip install --requirement requirements/edx/development.txt
tutor dev run lms python setup.py install
tutor dev run lms paver update_assets --settings=tutor.development
```

regisb:
@adrian-marcelo-gallardo can you please post the complete console output, either here or in a pastebin?

adrian-marcelo-gallardo:
@regisb here: https://pastebin.com/0kAHijxf

You'll see basically the following:
image

There is not much information about the issue other than the warnings about the npm packages.

Only after installing them with `npm install` everything started to work properly.

regisb:
I can't reproduce the issue on my Linux machine, so this might be caused by a different behaviour on Mac OS. Let's try to reproduce the issue with a more simple piece of code:

```
cd /tmp 
mkdir test                    
ln -s ./nonexistingfile ./test/linkto
docker run --rm -it -v "$(pwd)/test":/test ubuntu:16.04 chown --no-dereference -R root /test      
echo $?
```

On my machine, this command is successful and "0" is printed in the console. If I understand your problem correctly, this should print "1" on your machine.

adrian-marcelo-gallardo:
I think you are right @regisb:
```bash
➜  ~ cd /tmp
➜  /tmp mkdir test
➜  /tmp ln -s ./nonexistingfile ./test/linkto
➜  /tmp docker run --rm -it -v "$(pwd)/test":/test ubuntu:16.04 chown --no-dereference -R root /test
Unable to find image 'ubuntu:16.04' locally
16.04: Pulling from library/ubuntu
16c48d79e9cc: Pull complete 
3c654ad3ed7d: Pull complete 
6276f4f9c29d: Pull complete 
a4bd43ad48ce: Pull complete 
Digest: sha256:6239ca16d54d8edd38788aacde7c6426e5999ffbbd7e97ac3cf04d474341d079
Status: Downloaded newer image for ubuntu:16.04
chown: changing ownership of '/test/linkto': No such file or directory
➜  /tmp echo $?
1
```

regisb:
\o/ that means we have identified a bug with chown on docker for Mac OS. Maybe we'll try to fix that one day; in the meantime I'll try to propose a solution for tutor users. The `npm install` instructions is not satisfying, as it requires that npm is installed on the host. I'll work out something.

adrian-marcelo-gallardo:
@regisb totally agree on looking for a better option and a more final solution :)

Let me know if you need something else.

regisb:
@adrian-marcelo-gallardo can you please try out the master branch and see if it solves your problem? You will need to build a custom openedx-dev docker image. Please read the instructions here: https://github.com/overhangio/tutor/blob/master/docs/dev.rst

adrian-marcelo-gallardo:
Hello @regisb , I think the `npm` problem is now solved, and the delay of the `chown` was reduced to `1m 30s` (on MacOs Mojave) 👏 👏 .

I'm now struggling to get the `watchthemes` command working again, but I think that should be something to discuss in another issue :P

regisb:
Awesome! 1.5min is still a long time to wait, but I'm afraid I can't do much about it (unless you are willing to face permission issues on the host).

Concerning the watchthemes command, did you notice that it was modified? (see the docs)

PR #258, opened 2019-10-11
Openedx/changelogo wry

jekeans:


regisb:
I believe this PR was mistakenly opened here, and not on the base repo. Now closing.

PR #271, opened 2019-12-03
doc: Add instructions for running Tutor on Podman

fghaas:
This is a work-in-progress. I'm still testing this, but it looks like it's working as expected, so far.

> https://podman.io/ is meant to be a drop-in replacement for Docker.
> Thus, with some tweaking to the installation environment, it appears
> to be perfectly feasible to run Tutor in a Docker-less environment
> that only has Podman and podman-compose installed.
>
> Add installation instructions for doing just that.

### DIFF
--- a/docs/podman.rst
+++ b/docs/podman.rst
@@ -1,0 +1,51 @@
+.. _podman:
+
+Running Tutor with Podman
+-------------------------
+
+You have the option of running Tutor with `Podman <https://podman.io/>`__, instead of the native Docker tools. This has some practical advantages: it does not require a running Docker daemon, and it enables you to run and build Docker images without depending on any system component running ``root``. As such, it is particularly useful for building Tutor images from CI pipelines.
+
+The ``podman`` CLI aims to be fully compatible with the ``docker`` CLI, and ``podman-compose`` is a fully-compatible alias of ``docker-compose``. This means that you can use both together with Tutor, without making any changes to Tutor itself.
+
+.. warning::
+   You should not attempt to run Tutor with Podman on a system that already has native ``docker`` and ``docker-compose`` installed. If you want to switch to ``podman`` and ``podman-compose`` using the aliases described here, uninstall the native Docker packages first. 
+
+
+Enabling Podman
+^^^^^^^^^^^^^^^
+
+Podman is supported on a variety of development platforms, see the `installation instructions <https://podman.io/getting-started/installation>`_ for details.
+
+Once you have installed Podman and its dependencies on the platform of your choice, you'll need to make sure that its ``podman`` binary, usually installed as ``/usr/bin/podman``, is aliased to ``docker``, and is included as such in your system ``$PATH``. On some CentOS and Fedora releases you can install a package named ``podman-docker`` to do this for you, but on other platforms you'll need to take of this yourself.
+
+- If ``$HOME/bin`` is in your ``$PATH``, you can create a symbolic link there::
+
+    ln -s $(which podman) $HOME/bin/docker
+
+- If you want to instead make ``docker`` a system-wide alias for ``podman``, you can create your symlink in ``/usr/local/bin``, an action that normally requires ``root`` privileges::
+
+    sudo ln -s $(which podman) /usr/local/bin/docker
+
+
+Enabling podman-compose
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``podman-compose`` is available as a package from PyPI, and can thus be installed with ``pip``. See `its README <https://github.com/containers/podman-compose/blob/devel/README.md>`_ for installation instructions. Note that if you have installed Tutor in its own virtualenv, you'll need to run ``pip install podman-compose`` in that same virtualenv.
+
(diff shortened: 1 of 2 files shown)


regisb:
Hi @fghaas! This is very interesting. I know that @silviot also wanted to run Tutor with Podman, but at the time I thought it was not so relevant, because Podman did not have enough mind share. I realize now that I was wrong, as docker is losing more and more ground in the container field.
I'll have to test tutor more extensively with Podman myself. Meanwhile, I suggest you provide less documentation on how to install Podman, and instead point to the relevant upstream documentation [page](https://podman.io/getting-started/installation).

docs/podman.rst
```diff
@@ -0,0 +1,51 @@
...
+The ``podman`` CLI aims to be fully compatible with the ``docker`` CLI, and ``podman-compose`` is a fully-compatible alias of ``docker-compose``. This means that you can use both together with Tutor, without making any changes to Tutor itself.
+
+.. warning::
+   You should not attempt to run Tutor with Podman on a system that already has native ``docker`` and ``docker-compose`` installed. If you want to switch to ``podman`` and ``podman-compose`` using the aliases described here, uninstall the native Docker packages first. 
+
+
+Enabling Podman
+^^^^^^^^^^^^^^^
+
+Podman is supported on a variety of development platforms, see the `installation instructions <https://podman.io/getting-started/installation>`_ for details.
```
fghaas: @regisb:

> Meanwhile, I suggest you provide less documentation on how to install Podman, and instead point  > to the relevant upstream documentation [page](https://podman.io/getting-started/installation).

Yeah it does exactly that, here. It then goes on to explain what *else* is necessary to make *Tutor* play nicely with Podman, which I do consider relevant, and which isn't discussed in the upstream docs.

docs/podman.rst
```diff
@@ -0,0 +1,51 @@
...
+
+- If you want to instead make ``docker`` a system-wide alias for ``podman``, you can create your symlink in ``/usr/local/bin``, an action that normally requires ``root`` privileges::
+
+    sudo ln -s $(which podman) /usr/local/bin/docker
+
+
+Enabling podman-compose
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``podman-compose`` is available as a package from PyPI, and can thus be installed with ``pip``. See `its README <https://github.com/containers/podman-compose/blob/devel/README.md>`_ for installation instructions. Note that if you have installed Tutor in its own virtualenv, you'll need to run ``pip install podman-compose`` in that same virtualenv.
```
fghaas: Likewise, it references the upstream installation instructions for `podman-compose` here, and then explains what is necessary so that _Tutor_ can use it as a drop-in alias for `docker-compose`.

docs/podman.rst
```diff
@@ -0,0 +1,51 @@
...
+
+- If you run Tutor and ``podman-compose`` in a virtualenv, create the symlink in that virtualenv's ``bin`` directory: activate the virtualenv, then run::
+
+    ln -s $(which podman-compose) $(dirname $(which podman-compose))/docker-compose
+
+- If you do not, create the symlink in ``/usr/local/bin``, using ``root`` privileges::
+
+    sudo ln -s $(which podman-compose) /usr/local/bin/docker-compose
+
+
```
regisb: According to my tests, the `docker.io` search registry should be added to registries.conf:

```
$ cat /etc/containers/registries.conf
[registries.search]
registries = ['docker.io']
```

fghaas: Yes, and https://podman.io/getting-started/installation#registriesconf does mention that. You previously asked to not duplicate the information from that page, but rather include just a reference — have you changed your mind?

regisb: The thing is that users will not be aware that tutor requires access to the docker.io registry until they actually run tutor with Podman. Maybe we should give them a hint that points to the relevant paragraph from the podman installation instructions.

On second thought, it would make sense for tutor to specify that the base images come from docker.io, instead. That way, we would not have to configure podman.

regisb: Fixed here: https://github.com/overhangio/tutor/commit/c03e86f43b2d0dad20457ceaaf91dc6a1d4842d9

regisb:
As far as I am concerned, this is ready to be merged. Would you like to test something else @fghaas?

docs/podman.rst
fghaas: Ah yes, that looks like a good approach to me. It's also [PEP 20](https://www.python.org/dev/peps/pep-0020/) compliant ("Explicit is better than implicit.")

fghaas:
If you're happy then I think we're good! I just dropped the `WIP:` prefix.

regisb:
Thanks a lot for pushing this! And @silviot you might be interested in this ;)

fghaas:
Just following up here because https://docs.tutor.overhang.io/install.html doesn't yet reflect the changes. (Which might be intentional — just making sure the doc rebuild wasn't _unintentionally_ omitted.)

regisb:
This is normal; we have to do a new release in order for the docs to be updated, and I was waiting for more substantial changes before I created a new minor tag. But it doesn't make much sense to wait, so I'll create a new release now.

Issue #276, opened 2019-12-11
Add the ability to customize CMS and LMS Features setting.


cacciaresi:
**Is your feature request related to a problem? Please describe.**

We needed to customize only LMS Features setting in its config file, we didn't want to have the new setting entry in CMS settings.

**Describe the solution you'd like**

We have added this change in lms.env.json:
`{{ patch("lms-env-features", separator=",\n", suffix=",")|indent(4) }}`

and this in cms.env.json:
`{{ patch("cms-env-features", separator=",\n", suffix=",")|indent(4) }}`

By doing this then we can add a file named: `lms-env-features` in a plugin and that will add the setting only in the LMS settings.

regisb:
This sounds reasonable. Would you like to make a PR?

PR #278, opened 2019-12-12
Add the ability to customize CMS and LMS Features setting

cacciaresi:
Solution for #276.

I noticed that there was a missing plugin in docs, let me know that it was just missing or if it was remove for a particular cause.


regisb:
:ok_hand:

PR #283, opened 2020-01-24
Docs should better explain how to renew SSL/TLS certificates

cacciaresi:


regisb:
See the relevant conversation here: https://discuss.overhang.io/t/renew-lets-encrypt-cert-failing/268/4

cacciaresi:
@regisb do you think that we should add more info to this PR?

regisb:
This is fine, sorry about the delay.

cacciaresi:
@regisb :heart: no issues, just wanted to know if we need to add something else. thanks!

PR #322, opened 2020-05-08
Smaller Dockerfile base image

jjasonkal:
- Using python:3.7-slim-stretch as base decrease the size of the tutor image from 1.05GB to 282MB.


regisb:
Hi @jjasonkal! Thanks for your pr. This is a good change, but I wonder: do you actually use this docker image? Even I never run tutor from inside docker.

jjasonkal:
Thank you too @regisb , i optimized it for someone who want to run it from inside docker but yes it is easier to run it otherwise.

PR #341, opened 2020-06-14
docs: system prune to complete removal

antoinehidden:
The images, containers and volumes linger even after the removing the
.local directory. They can be removed with 

    docker system prune --all

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -105,6 +105,7 @@ Then, delete all data associated to your Open edX platform::
     
     # WARNING: this step is irreversible
     sudo rm -rf "$(tutor config printroot)"
+    sudo docker system prune --all
 
 Finally, uninstall Tutor itself::
     


antoinehidden:
It would be better to have something like ```tutor local down``` (i.e. the equivalent of ```docker-compose down```). Although this could be confused with ```stop``` by the user.

regisb:
Hi! Instructions for managing locally-running images should not be added to the general uninstallation instructions, as they do not apply to users who run Open edX on Kubernetes. Moreover, I'm not too excited by the idea of educating users about how to run Docker: this is just not Tutor's job. If we start going down that road, there are too many questions that we will need to address. Maybe it would make sense to add a couple pruning instructions to the Troubleshooting section? This would help users clear some disk space, for instance.

antoinehidden:
It's maybe a confusion on my part: after stopping tutor (with tutor local stop) and removing the .local directory, I noticed there was three volumes left. Which I probably mistook for volumes used to store tutor data (databases maybe ?). But maybe those volumes are not persistent tutor data, I did not check.

regisb:
>  But maybe those volumes are not persistent tutor data, I did not check.

I'd be curious to see what those three volumes are, if you manage to reproduce this issue.

PR #350, opened 2020-07-10
Feature to disable certificate views

GustavoMS97:
While using tutor for a project, we felt the need to hide the platform certificate data so that we can use our own. And to avoid removing code left and right, it's easier if it is configurable. Oh, and i made sure it was backwards compatible.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -186,6 +186,16 @@ To renew the certificate, run this command once per month::
 
 .. _customise:
 
+Disable default course certificates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- ``DISABLE_CERTIFICATES_HTML_VIEW`` (default: ``false``)
+
+By activating this feature, the system will no longer show certain views related to the default certificates platform.
+
+This feature was developed so that it is retrocompatible, so that it wont break previous versions.
+
+
 Custom Open edX docker image
 ----------------------------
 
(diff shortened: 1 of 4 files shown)


regisb:
Hi @GustavoMS97! Thanks for your PR. I feel like you are right that users should have a choice to enable certificate views, but I don't think that this setting should be added to the Tutor configuration. We cannot add a Tutor setting for every edx-platform/ecommerce/discovery etc. setting, as there are just too many of them.

As an alternative, may I suggest that you create a Tutor plugin to enable/disable this feature? It should be relatively simple with the basic yaml syntax: https://docs.tutor.overhang.io/plugins.html#yaml-file
Basically, you would need to add `FEATURES["CERTIFICATES_HTML_VIEW"] = False` to the `openedx-common-settings` patch.

I'm thinking of adding a set of base plugins to the default Tutor installation: simple yaml-based plugins that would affect the behaviour of the platform. This is one example of a plugin that could be added to this list.

Do you agree?

GustavoMS97:
Hi @regisb 
I agree with that! Makes sense that adding a setting for every little feature wouldn't be the ideal, and adding a base plugin to change those behaviors would make everyone's life easier.

LucGranato:
@regisb, this pull request looks very relevant because the `CERTIFICATE_HTML_VIEW` is enabled by default on both CMS and LMS env template files.

https://github.com/overhangio/tutor/blob/66ab851848d726629d562e45c5bbe3ac1bb90e5f/tutor/templates/apps/openedx/config/lms.env.json#L11

https://github.com/overhangio/tutor/blob/66ab851848d726629d562e45c5bbe3ac1bb90e5f/tutor/templates/apps/openedx/config/cms.env.json#L11

@GustavoMS97 implementation is careful because it will not break the current default behavior.

regisb:
It looks like this plugin could be implemented as simply as:

```
name: disablehtmlcertificates
version: 0.1.0
patches:
    # Disable certificate HTML view
    FEATURES["CERTIFICATES_HTML_VIEW"] = True
```

Thus I'll close this for now. Feel free to re-open if you think it's necessary.

PR #354, opened 2020-07-16
reload-gunicorn: look for python3 binary

moraleja39:
Quite a simple change. I found the `reload-gunicorn` bin quite handy, which is broken since the migration to juniper.


regisb:
Good catch!

PR #355, opened 2020-07-17
pass add-host to docker build command

grinderz:
sometimes need static domain->ip maping for docker build process, use this feature from docker build.


regisb:
Hi @grinderz! First of all, sorry for the late response. This is a legitimate requirement, but I was not so happy with the fact that the list of arguments got longer and longer. What I did was that I implemented this by picking some of your changes, but limiting the number of arguments passed to `image.build()`: https://github.com/overhangio/tutor/commit/b1c7dab03e84ad49f110cf022134b119152a97ca

This change will be included in the next release.

Thanks for the pull request!

PR #356, opened 2020-07-18
k8s workers queues fix

susguzman:
Hello,

This fix avoid that a CMS Worker can read a message for a LMS Worker and vice versa for k8s templates.

Best Regards,


regisb:
Good catch, thanks!

PR #364, opened 2020-08-18
Enhance kubernetes support

menardorama:
Hi
In order to support secure installation on K8s, this PR add the following changes :

* support for security context in order to containers as a limited user account
> * Dockerfiles for openedx and forum has been updated
> * change nginx to nginx-unprivileged
> * config maps have been updated
> * switch smtp docker image 
* service type : NodePort or ClusterIP (which is generally the default service type)
* fix to tutor k8s quickstart (just add a 5 seconds pause after object creation and before migrations)
* add support for ingress class (usefull when you have several ingress controllers, internal/external for example)


regisb:
Hey @menardorama, I'm really sorry to have left this rot for so long. I do want to enable users to run unprivileged containers on Kubernetes, but there are incompatibilities with the way the local installation works. I'm afraid that if I merge your PR the local installation will not work anymore -- mostly because media assets will not be writable anymore. We might be able to resolve this issue in the next Open edX release (Lilac) to come out on June 9th, on the condition that we migrate all media assets to object storage (aka: [minio](https://github.com/overhangio/tutor-minio)).

Meanwhile, I have a question for you: how to you build Docker images for your Kubernetes cluster? Do you build them in Kubernetes, with [Docker-in-Docker](https://hub.docker.com/_/docker/), or do you build them on a dedicated host? If you do the former, do you manage to run the Docker container in unprivileged mode?

menardorama:
Hi,

I totally understand your concern regarding incompatibilities between local and kubernetes deployment.

Regarding kubernetes build we are actually doing it on dedicated ci nodes on our kubernetes cluster.

As Docker-in-Docker require privileged mode, that was the only way to guarantee a basic security. But as we are running on premise Gitlab and kubernetes cluster it's not really an issue for us.

On the other hand having a public service that run in privileged mode is an issue for us.

For now our openedx project is on hold until we found a suitable solution (we already have an old openedx platform hosted in a virtualmachine).

No regarding my PR, I think modifying the Dockerfile is just a nice to have but not mandatory (you can use whatever userid you want), but nginx port and security context should be configurable with default value.

So that you can cover all the usecases

menardorama:
Closing as the app doesn't fit to be running on our k8s clusters, even with custom configs

I can be more explicit, but that's not the right place to talk

regisb:
Thanks for your feedback @menardorama. Would you feel more confortable describing your reasons by email? My address is available on my public profile: https://github.com/regisb

PR #367, opened 2020-10-01
Configurable emails config settings

odemakov-epfl:
Allow users to define each *_EMAIL settings separately.


regisb:
Thanks for the PR!

PR #372, opened 2020-10-15
Add replicaSet support to MongoDB connection

odemakov-epfl:
Hi Regis,

Here is another pull request for tutor. I've tested it with our staging infra and don't know do I need to add test cases here? Like set up `tutor local` with replicated mongodb(one Docker image with mongodbs on different ports)?

Allow to set replicaSet to MongoDB connection with MONGODB_REPLICASET env variable. In this case MONGODB_HOST must be an array.


regisb:
Hi @odemakov-epfl! I don't think this is a feature that we want to support for all Tutor users -- Especially because it can be implemented separately by a Tutor plugin. Is it OK for you to add replicaset support to your platform via a Tutor plugin? You will need to override the `replicaSet` attribute  of the `DOC_STORE_CONFIG`, `CONTENTSTORE`, and `MODULESTORE` objects from the LMS and CMS production settings. The simplest way to do so is to implement the "openedx-common-settings" patch.

Do you agree?

odemakov-epfl:
@regisb , are you sure you don't want replicaSet for production? I thought it is one of the must-have feature. 

If so, no problem, will do it via plugin.

regisb:
> are you sure you don't want replicaSet for production? I thought it is one of the must-have feature.

The thing is that there are [many](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient) possible parameters to pass to the Mongodb client constructor and I don't want to introduce a new Tutor setting for each one of them. On the other hand, I understand that having replicasets should be a common enough feature, and I don't want to make it difficult for users to use this setting.

So, what I think I'll do (if you agree) is that I'll add a `mongodb_parameter` to the LMS and CMS settings. Something along the lines of (in "settings/partials/common_all.py"):

```
mongodb_parameters = {
    "host": "{{ MONGODB_HOST }}",
    "port": {{ MONGODB_PORT }},
    {% if MONGODB_USERNAME and MONGODB_PASSWORD %}
    "user": "{{ MONGODB_USERNAME }}",
    "password": "{{ MONGODB_PASSWORD }}",
    {% endif %}
    "db": "{{ MONGODB_DATABASE }}"
}

DOC_STORE_CONFIG = mongodb_parameters
CONTENTSTORE["DOC_STORE_CONFIG"] = DOC_STORE_CONFIG
...
```

That way, you could easily write a plugin that would only modify `mongodb_parameters` and it would automatically change the other related settings.

What do you think?

odemakov-epfl:
> That way, you could easily write a plugin that would only modify `mongodb_parameters` and it would automatically change the other related settings.
> 
> What do you think?

I agree, this way it will be much flexible.

regisb:
Great! Can you please try out tutor's master branch? If it works for you I'll make a new release.

regisb:
@odemakov-epfl did you read this? https://discuss.overhang.io/t/mongodb-failover-replica-set-support/1001/2

odemakov-epfl:
> @odemakov-epfl did you read this? https://discuss.overhang.io/t/mongodb-failover-replica-set-support/1001/2

I've replied to TS.

With latest tutor I able to set up mongo connection I want with simple plugin, see below. Thank you, @regisb , for such quick support. When you will release it? :) 

```
name: mongodb_replicaset
version: 0.1.0
patches:
  openedx-common-settings: |
    mongodb_parameters = {
        "read_preference": "PRIMARY",
        "replicaSet": "{{ MONGODB_REPLICASET }}",
        "host": [{% for host in MONGODB_HOST %}"{{ host }}"{{ "," if not loop.last }}{% endfor %}],
        "port": {{ MONGODB_PORT }},
        "user": "{{ MONGODB_USERNAME }}",
        "password": "{{ MONGODB_PASSWORD }}",
        "db": "{{ MONGODB_DATABASE }}"
    }
    DOC_STORE_CONFIG = mongodb_parameters
    CONTENTSTORE = {
        "ENGINE": "xmodule.contentstore.mongo.MongoContentStore",
        "ADDITIONAL_OPTIONS": {},
        "DOC_STORE_CONFIG": DOC_STORE_CONFIG
    }

    # Load module store settings from config files
    update_module_store_settings(MODULESTORE, doc_store_settings=DOC_STORE_CONFIG)
```

regisb:
> When you will release it? :)

v10.3.1 is out: https://github.com/overhangio/tutor/releases

PR #388, opened 2020-12-07
Increase the timeout of the gunicorn worker command in openedx Dockerfile

mrtndwrd:
As discussed in https://discuss.overhang.io/t/lms-worker-timeout-issue/1057/8, the timeout of the gunicorn worker is a bit low, causing the worker to crashloop on slower systems. By increasing the timeout to 120 seconds, the worker should have time to come up before it gets restarted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Note: Breaking changes between versions are indicated by "💥".
 
+## Unreleased
+
+- [Improvement] Increase the timeout of the gunicorn worker command in openedx Dockerfile
+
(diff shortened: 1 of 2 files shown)


regisb:
Thanks for the PR @mrtndwrd!

regisb:
Koa is supposed to come out on Wednesday, and uwsgi will be replacing gunicorn/whitenoise in tutor v11. In order to get your patch in a release, I'm publishing 10.5.2 right now.

mrtndwrd:
Awesome, thanks a lot again for the quick response!

PR #396, opened 2021-01-17
Fix execute command

misterfish:
`args` should be a list, but we receive a tuple from Click.


regisb:
Hey @misterfish thanks for the PR! Somehow I missed the notification, sorry about that. I published a fix in v11.1.1. It's slightly different but it does essentially the same thing: https://github.com/overhangio/tutor/commit/df3d4325186c22947424734cfd5bbdd9c8c8de9e

PR #397, opened 2021-01-24
Fix default settings on development environment

arbrandes:
First, allow using custom Django settings on a development environment (as documented but not implemented), setting it to the
correct value of `tutor.development`.  Prior to this, `tutor dev runserver lms` would default to `tutor.production` when on a custom edX branch.

Second, fix the documentation so the correct environment variable is described, at the same time removing an option that doesn't seem to work.

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -217,7 +217,7 @@ Custom edx-platform settings
 
 By default, tutor settings files are mounted inside the docker images at ``/openedx/edx-platform/lms/envs/tutor/`` and ``/openedx/edx-platform/cms/envs/tutor/``. In the various ``dev`` commands, the default ``edx-platform`` settings module is set to ``tutor.development`` and you don't have to do anything to set up these settings.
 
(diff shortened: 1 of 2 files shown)


regisb:
Good catch @arbrandes! I'd like you to make the following changes, though:

1. I'd rather have all tutor-related env vars were prefixed by `TUTOR_`: could you please take this opportunity to replace all instances of `EDX_PLATFORM_SETTINGS` by `TUTOR_EDX_PLATFORM_SETTINGS`?
2. Could you also add a note to the Changelog? Make sure to prefix it by the breaking change symbol.
Thanks!

arbrandes:
@regisb, sure thing, will do!  Thanks for reviewing! :)

regisb:
@arbrandes I didn't realise the dev environment was broken. This was reported here: https://discuss.overhang.io/t/koa-dev-lms-doesnt-find-static-content/1250/7
So I took your PR and made the required changes. This can now be closed. Thanks for the fix!

arbrandes:
@regisb, no complaints from me!  Glad to have been of assistance.  Thanks!

PR #406, opened 2021-02-26
Fix SAML 'Invalid Request' Error

Ant1x:
After upgrading from Juniper to Koa, we have noticed our SAML authentications requests started to fail.

Upon investigation this was caused by the implementation of Caddy. 
Previously, nginx was setting the `X-Forwarded-Port` header upstream to openedx. It was explicitly defined in the configuration file in earlier versions of Tutor.
Caddy will set some headers like `X-Forwarded-For` by default, however it will not set `X-Forwarded-Port` by default.
This PR modifies the Caddyfile to set this header. The current nginx configuration will pass this header through transparently.

Further information about the SAML issue is below.

The OneLogin SAML library will use the incoming host and port information to validate that a client is calling back to the exact ACS URL that was specified in the original SAML authentication request.
The client's incoming request appears as: `https://tutor.com:8000/auth/....` which does not match the original server's ACS URL (of `https://tutor.com/auth/...`) in the authentication request.
However, in Juniper this was working and port 8000 was not appended - as nginx was telling openedx the client is on port 443.

The port mismatch causes the OneLogin library to deny the authentication request, and this library receives the port information ultimately from the `X-Forwarded-Port` header (if present) otherwise it will use the listening port of the lms (`8000`)

Unfortunately it was necessary to hard-code ports 443 and 80 into this PR, since the variables Caddy has available rely only on the incoming `Host` header - and most browsers/clients will not set a port in this header if it's using the well-known port for the protocol

### DIFF
--- a/tutor/templates/apps/caddy/Caddyfile
+++ b/tutor/templates/apps/caddy/Caddyfile
@@ -1,5 +1,7 @@
 {{ LMS_HOST }}{% if not ENABLE_HTTPS %}:80{% endif %} {
-    reverse_proxy nginx:80
+    reverse_proxy nginx:80 {
+        header_up X-Forwarded-Port {{ 443 if ENABLE_HTTPS else 80 }}
+    }
 }
 preview.{{ LMS_HOST }}{% if not ENABLE_HTTPS %}:80{% endif %} {
     reverse_proxy nginx:80


regisb:
Thanks for the thorough bug description @Ant1x. I think I understand your problem. I would like to avoid adding too much stuff to the Caddy configuration, if it can be avoided. Can you give me a link to the OneLogin library you are using? Is it this one? https://github.com/onelogin/python3-saml I see no reference to the x-forwarded headers outside of the demo apps.

Ant1x:
Hi @regisb 

I'm not actually using a library myself, I am only using the built in SAML SP functions provided by `edx-platform`
Under the hood, it looks like it uses https://github.com/python-social-auth/social-core which then uses the OneLogin library.
This file in `edx-platform` references it here https://github.com/edx/edx-platform/blob/open-release/koa.2a/common/djangoapps/third_party_auth/saml.py

From what I can tell, the OneLogin library takes in a port number `server_port` from the incoming `request` and uses that to build the initial login request. This would have been using the correct port of 443 with the nginx reverse proxy in Juniper, but by default now in Koa openedx sees port 8000 in the request. 

I believe this Django middleware code in `edx-platform` is responsible for setting the `request` port to the value of the `X-Forwarded-Port` header if it exists, otherwise it defaults it to `server-port` which is the port that the lms is listening on.
https://github.com/edx/edx-platform/blob/f1324710af5559375e0e13cb0d58f3b820c9ae27/openedx/core/lib/x_forwarded_for/middleware.py

In the previous Tutor releases (v10.x), nginx had the X-Forwarded-Port header set here: https://github.com/overhangio/tutor/blob/d3c842c8dad6f63762240d7780400373b1649d0d/tutor/templates/apps/nginx/lms.conf#L33

This post in the Caddy community says they will set `X-Forwarded-For` and `X-Forwarded-Proto` by default but they don't set `X-Forwarded-Port` by default: https://caddy.community/t/disable-auto-https-on-lan/11036/7

I know we can set the header in nginx as an alternative but I'm not sure what the long term plan is for that. I had a look through the OneLogin library also to see if there was a way to turn off this port validation feature, but wasn't able to find anything in my initial searches. My searching lead me to this code in the library which will always read in and process the port number if it exists, and in every case I have tested with Tutor, the port number always exists in the edx-platform requests:  

https://github.com/onelogin/python3-saml/blob/c25df8164563df28ec94dd6f5d7a9a6c65c06e83/src/onelogin/saml2/utils.py#L263-L270

regisb:
> The client's incoming request appears as: https://tutor.com:8000/auth/.... which does not match the original server's ACS URL (of https://tutor.com/auth/...) in the authentication request.

Wait: why does the client's incoming request appear as "https://tutor.com:8000/auth/"? where does this 8000 value come from? What is this client? It is unclear to me what this client actually is: is it edx-platform? or a different app?

regisb:
@Ant1x I slept on this and realised I am being too picky. This is a good fix and should be included in Tutor. Can you please add a line to the CHANGELOG? Then I'll merge this PR.
For reference, this seems to be related: https://discuss.overhang.io/t/sso-missing-backends-how-do-i-enable-them/251/13

regisb:
I eventually decided to add the Changelog entry myself. Thanks for your contribution @Ant1x!

PR #407, opened 2021-02-26
fix: Only ignore warnings if they exist.

feanil:
If we are running a newer version of DRF or are in the process of
upgrading DRF, we want to be able to drop the version where these
warnings exist without breaking tutor.  This pattern will suppress the
warnings if it can but will not fail to startup if that kind of warning
no longer exists.

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/pre_common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/pre_common_all.py
@@ -1,10 +1,19 @@
 # Silence overly verbose warnings
 import logging
 import warnings
-from django.utils.deprecation import RemovedInDjango30Warning, RemovedInDjango31Warning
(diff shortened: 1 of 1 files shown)


regisb:
Hi @feanil! This change would make it more difficult to upgrade Tutor to the next release. We would have to manually check whether the imports work, and if they don't remove them. Without the `except` statements, it will be much easier to detect that these warnings are obsolete.

I assume that this is needed in order to run Tutor on the Open edX master branch? The more I think about it, the more convinced I am that we would have to do that in a separate "edge" branch from the shared release branches. We could then remove these warnings in this edge branch, and the change would be merged during the next upgrade.

Does that make sense?

regisb:
Closing for now -- feel free to reopen if you think this is a wrong decision.

PR #408, opened 2021-03-03
Add patches that enable adding settings during build process (collectstatic and i18ncompile).

ericfab179:
As discussed here https://discuss.overhang.io/t/proposal-for-addition-of-some-patches/1309, this PR adds two new patches to
set some extra settings used during the collectstatic and i18ncompile processes.


regisb:
Thanks for the PR @ericfab179!

PR #412, opened 2021-03-30
fix: TypeError when PLUGINS is None

regisb:
When the PLUGINS config entry is None (`PLUGINS:`), the following error
was being triggered:

  File "/.../tutor/tutor/plugins.py",
  line 304, in is_enabled
      return name in config.get(CONFIG_KEY, [])
      TypeError: argument of type 'NoneType' is not iterable

This is ready for review @BbrSofiane.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


BbrSofiane (approved):
- [x] Works as intended
- [ ] Listed in changelog
- [ ] Git messages

Looks good, this is ready to merge.

PR #415, opened 2021-03-31
Added configuration changes for mongo atlas style clusters

zakum1:
Getting forum to work with a mongo cluster (e.g. Mongo Atlas) required two new environment variables to env/build/forum/Dockerfile:

```dockerfile
ENV MONGODB_PROTOCOL "mongodb"
ENV MONGODB_PARAMS ""
```

`MONGODB_PROTOCOL` can be set to mongodb+srv for a cluster
`MONGODB_PARAMS` is optional for adding query parameters to the URI (it must start with ?)

This change is propsoed to `env/build/forum/bin/docker-entrypoint.sh`:

```sh
#!/bin/sh -e


if test "$MONGODB_PROTOCOL" = 'mongodb+srv'; then
  export MONGOHQ_URL="$MONGODB_PROTOCOL://$MONGODB_AUTH$MONGODB_HOST/cs_comments_service$MONGODB_PARAMS"
  export MONGO_WAIT=""
else
  export MONGOHQ_URL="$MONGODB_PROTOCOL://$MONGODB_AUTH$MONGODB_HOST:$MONGODB_PORT/cs_comments_service$MONGODB_PARAMS"
  export MONGO_WAIT="-wait tcp://$MONGODB_HOST:$MONGODB_PORT"
  echo "Waiting for mongodb..."
fi

echo "Waiting for elasticsearch..."
dockerize $MONGO_WAIT -wait $SEARCH_SERVER -wait-retry-interval 5s -timeout 600s

exec "$@"
```

If the `MONGODB_PROTOCOL` is set to `mongodb+srv` then there are two key changes:

1.    The port must not be included in the URL
2.    The host cannot be used as a true server name for “waiting” because it is a DNS entry with only SRV records. I have elected not to wait at all for a cluster because it is unlikely that this cluster is starting up at the same time as the forum, because it is typically provisioned beforehand.

In addition, we add these three parameters to `config.yml`

`MONGODB_REPLICA_SET` replica set name for the cluster
`MONGODB_USE_SSL` indicates if the cluster uses SSL (`true` or `false`)
`MONGODB_AUTH_SOURCE` the database name of the authentication database on the cluster

These are for the LMS / CMS connecting to mongo - openedx supports these parameters in the pymongo initialisation, we just need to update them in the parameters that get passed in as `mongodb_parameters`


docs/configuration.rst
```diff
@@ -140,6 +140,10 @@ Mongodb
 - ``MONGODB_PORT`` (default: ``27017``)
 - ``MONGODB_USERNAME`` (default: ``""``)
 - ``MONGODB_PASSWORD`` (default: ``""``)
+- ``MONGODB_REPLICA_SET`` (default: ``""``)
```
regisb, line 143: Why isn't this variable taken into account in the forum config?

tutor/templates/build/forum/bin/docker-entrypoint.sh
```diff
@@ -1,8 +1,15 @@
 #!/bin/sh -e
 
-export MONGOHQ_URL="mongodb://$MONGODB_AUTH$MONGODB_HOST:$MONGODB_PORT/cs_comments_service"
+if test "$MONGODB_PROTOCOL" = 'mongodb+srv'; then
+  export MONGOHQ_URL="$MONGODB_PROTOCOL://$MONGODB_AUTH$MONGODB_HOST/cs_comments_service$MONGODB_PARAMS"
+  export MONGO_WAIT=""
```
regisb, line 5: The `MONGOHQ_URL` variable needs to be exported, but not `MONGO_WAIT`, which is only used in this entrypoint script.

tutor/templates/build/forum/bin/docker-entrypoint.sh
```diff
@@ -1,8 +1,15 @@
...
+else
+  export MONGOHQ_URL="$MONGODB_PROTOCOL://$MONGODB_AUTH$MONGODB_HOST:$MONGODB_PORT/cs_comments_service$MONGODB_PARAMS"
+  export MONGO_WAIT="-wait tcp://$MONGODB_HOST:$MONGODB_PORT"
+  echo "Waiting for mongodb..."
```
regisb, line 9: I suggest you write: `services='mongodb/elasticsearch'` and then: `echo "Waiting for $services"`.

regisb (changes requested):

docs/configuration.rst
zakum1: In the forum config, a URI with `mongodb+srv://...` is used and in this case there is no need to specify the replica set. The replica set is needed only when the shard hosts are explicitly listed in the `MONGODB_HOST`. I have tested and this works for the forum

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -16,6 +16,15 @@
     "password": None,
     {% endif %}
     "db": "{{ MONGODB_DATABASE }}",
+    {% if MONGODB_REPLICA_SET %}
+    "replicaSet": "{{ MONGODB_REPLICA_SET }}",
+    {% endif %}
+    {% if MONGODB_USE_SSL and MONGODB_USE_SSL == "true" %}
+    "ssl": True,
+    {% endif %}
```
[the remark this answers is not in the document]
zakum1: I have updated it along the lines you request - is it what you had in mind?

zakum1:
@regisb  I have implemented the changes that you requested - hope it covers all your concerns

docs/configuration.rst
pcliupc: Then if a replica set mongodb is used, do both MONGODB_HOST and MONGODB_REPLICA_SET need to be configured?

pcliupc: Then is it possible for forum to use replica set config if MONGODB_REPLICA_SET is configured, while MONGODB_PROTOCOL is configured as "mongodb"?

docs/configuration.rst
```diff
@@ -140,6 +140,10 @@ Mongodb
...
+- ``MONGODB_PROTOCOL`` (default: ``"mongodb"``)
+- ``MONGODB_USE_SSL`` (default: ``""``)
```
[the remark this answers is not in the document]
regisb: This variable should really be a boolean. Take a look for instance at the `ENABLE_HTTPS` config variable for inspiration.

tutor/templates/apps/openedx/settings/partials/common_all.py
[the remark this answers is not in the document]
regisb: This would be simpler if the variable were a boolean (see above).

docs/configuration.rst
regisb: @zakum1 Could you give an example of how you define all these variables to connect to Atlas? It would really help us better understand the use case. We are trying hard to minimize the number of Tutor config variables, which is why we ask so many questions :innocent:

regisb (changes requested):
Please see my additional comments. Also, please add a Changelog entry, squash commits and write an extensive commit message: https://docs.tutor.overhang.io/tutor.html#contributing
Also, please add a link to the forum topic in your commit message: https://discuss.overhang.io/t/connecting-to-a-mongodb-cluster-protocol-mongodb-srv/1272
Basically, if you find yourself writing a long PR description, then it means that you should instead be adding more information to the commit message :)

zakum1:
I have added the Changelog message, squashed all commits with a detailed commit message, including a link to the forum. I hope this is all OK now

zakum1 (commented):
I believe that I have implemented all requested changes

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -16,6 +16,13 @@
     "password": None,
     {% endif %}
     "db": "{{ MONGODB_DATABASE }}",
+    {% if MONGODB_REPLICA_SET %}
+    "replicaSet": "{{ MONGODB_REPLICA_SET }}",
+    {% endif %}
+    "ssl": {% if MONGODB_USE_SSL %} True {% else %} False {% endif %}
```
regisb, line 22: We need to add a coma "," at the end of the line, otherwise it will cause a parsing error when "authsource" is defined.

tutor/templates/build/forum/Dockerfile
```diff
@@ -48,5 +48,7 @@ ENV MONGODB_AUTH ""
 ENV MONGOID_AUTH_MECH ""
 ENV MONGODB_HOST "mongodb"
 ENV MONGODB_PORT "27017"
+ENV MONGODB_PROTOCOL "mongodb"
+ENV MONGODB_PARAMS ""
```
regisb, line 52: I don't understand what this `MONGODB_PARAMS` is supposed to do anymore, as it's not being defined anywhere.

docs/configuration.rst
regisb: OK, I think I understand: the `replicaSet` parameter is only required because the LMS/CMS connection string is outdated and does not make use of the `MONGODB_PROTOCOL`. Am I right? If yes, I think we can assume that the mongodb library will be updated soon-ish (by us or someone else) to support `MONGODB_PROTOCOL`. This will make the `replicaSet` parameter obsolete. If, for the moment, the `MONGODB_REPLICA_SET` configuration parameter can be replaced by a simple Tutor plugin, I'd rather document it than add a soon-deprecated configuration variable.

Am I making sense? If my interpretation is incorrect I think we should have a call to clarify things.


Could you please add a paragraph to the configuration docs (in the [mongodb section](https://docs.tutor.overhang.io/configuration.html#mongodb)) explaining how to connect to a MongoDB cluster?

regisb (changes requested):

.python-version
```diff
@@ -0,0 +1 @@
+3.6.8
```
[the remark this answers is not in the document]
BbrSofiane: It's a pyenv file for managing python versions, and yes it should be removed.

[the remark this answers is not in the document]
zakum1: I have added this to `.gitignore`

tutor/templates/apps/openedx/settings/partials/common_all.py
zakum1: I have fixed this as suggested

tutor/templates/build/forum/Dockerfile
zakum1: This was for additional parameters that may be needed for the connection to the cluster. I have removed it for now

docs/configuration.rst
regisb: What I was proposing in the comment above was:

- not adding the `MONGODB_REPLICA_SET` config variable.
- documenting how to write a plugin that would allow users to connect to an Atlas cluster.

I think we need to talk about this :) Can you please send me an email?

docs/configuration.rst
```diff
@@ -140,6 +140,9 @@ Mongodb
 - ``MONGODB_PORT`` (default: ``27017``)
 - ``MONGODB_USERNAME`` (default: ``""``)
 - ``MONGODB_PASSWORD`` (default: ``""``)
+- ``MONGODB_REPLICA_SET`` (default: ``""``)
+- ``MONGODB_USE_SSL`` (default: ``false``)
+- ``MONGODB_AUTH_SOURCE`` (default: ``""``)
```
regisb, line 145: `MONGODB_PROTOCOL` is a new configuration variable which should be added here.

zakum1: Updated as requested

regisb (commented):
@zakum1 can we get rid of the `MONGODB_REPLICA_SET` variable? I asked multiple times about this but you did not give me a clear answer.

zakum1:
> @zakum1 can we get rid of the `MONGODB_REPLICA_SET` variable? I asked multiple times about this but you did not give me a clear answer.

The replica set name is a needed parameter for the mongo client to connect to a clusrer. For example, on a typical cluster here is the connection URI for a cluster on Mongo Atlas, where the replica set must be specified

```
"mongodb://cluster0-shard-00-00.jnuls.mongodb.net:27017,cluster0-shard-00-01.jnuls.mongodb.net:27017,cluster0-shard-00-02.jnuls.mongodb.net:27017/myFirstDatabase?replicaSet=Cluster0-shard-0"
```

This parameter is also catered for in the OpenEdX client. In the mongo client code https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/mongo_utils.py the existence of the replicaSet parameter is used to select a specific client class to connect to the cluster (although the actual replica set name is not used after that), so I don't see a way to eliminate this parameter, since it is the only way to get the open edx mongo client to connect successfully to a cluster.

tutor/templates/config.yml
```diff
@@ -57,6 +57,11 @@ MONGODB_DATABASE: "openedx"
 MONGODB_PORT: 27017
 MONGODB_USERNAME: ""
 MONGODB_PASSWORD: ""
+MONGODB_REPLICA_SET: ""
+MONGODB_PROTOCOL: "mongodb"
+MONGODB_USE_SSL: ""
```
regisb, line 62: This variable must default to a boolean, as you explained in the docs. And then its usage must reflect the fact that its a boolean (see k8s/deployments.yml and k8s/jobs.yml below).

regisb (changes requested):
> In the mongo client code https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/mongo_utils.py the existence of the replicaSet parameter is used to select a specific client class to connect to the cluster (although the actual replica set name is not used after that)

I disagree with your interpretation. 

1. In Pymongo 3.10.1 (the version currently in use in Koa), the MongoReplicaSetClient class is a deprecated alias for MongoClient. Actually, we should open a PR on the edx-platform repo to get rid of this class.
2. The `replicaSet` argument is passed to the MongoClient/MongoReplicaSetClient via the `kwargs` object, so it is being used.
3. As far as I understand, to connect to an Atlas cluster, we should only have to modify the `db` parameter: https://pymongo.readthedocs.io/en/3.10.1/atlas.html

Your PR is adding extra configuration parameters, which is something I'd rather avoid. I have asked you multiple times to clarify the usage of these parameters by adding a paragraph to the docs. If you don't take the time to explain to us and to other Tutor users how these parameters have to be defined I'm afraid we are going to have to close this PR, which would we a shame because I'd rather make it possible for end users to connect to Atlas.

tutor/templates/config.yml
zakum1: As mentioned above, the Mongo connection to a cluster does not work without the replica set parameter (I saw this empirically and from the [documentation](https://pymongo.readthedocs.io/en/stable/examples/high_availability.html)  in which `replicasSet` needs to be specified in the client)

But I am actually not able to test this anymore to demonstrate that this is correct, because Mongo Atlas has stopped supporting v3.6 clusters (from April 30), and so I am no longer running Mongo on Atlas for Open edX. 

This PR is then only useful when Open edX upgrades to v4.0 - the timing for which seems to be [unclear](https://discuss.openedx.org/t/mongo-4-0-upgrade/3779) - or if anyone runs their own Mongo cluster.

I suggest that we close the PR.

zakum1: @regisb I am actually keen to do the test on lilac - do you have some pointers to running lilac with tutor? I see there is a `lilac` branch in the repo - can I just launch from that?

regisb: That would be great! Yes, you should install Tutor from the lilac branch. You should apply the first two steps listed here: https://discuss.overhang.io/t/how-to-upgrade-a-tutor-plugin/1488

> * Install Tutor from source, from the release candidate branch: in our particular case, it’s the lilac 4 branch.
> * Build all the tutor images locally (as they are not available from docker.io 1): tutor config save && tutor images build all.

Note that this version already depends on Mongodb 4.0.

zakum1:
I have created a new [PR 437](https://github.com/overhangio/tutor/pull/437) that targets the `lilac` branch (because lilac has upgraded to mongo v4 and Mongo Atlas no longer supports v3.6). I address the issue raised by @regisb in that PR.

PR #417, opened 2021-04-06
refactor: better config type

regisb:
I stumbled upon a bug that should have been detected by the type checking. Turns out, considering that config is of type Dict[str, Any] means that we can use just any method on all config values -- which is terrible. I discovered this after I set `config["PLUGINS"] = None`: this triggered a crash when I enabled a plugin.

We resolve this by making the Config type more explicit. We also take the opportunity to remove a few cast statements.

Pylint now also checks for unused imports.

I understand this is a complex PR -- well, we are still in the process of figuring out how to best use mypy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ test-format: ## Run code formatting tests
 	black --check --diff $(BLACK_OPTS)
 
 test-lint: ## Run code linting tests
-	pylint --errors-only --ignore=templates ${SRC_DIRS}
+	pylint --errors-only --enable=unused-import --ignore=templates ${SRC_DIRS}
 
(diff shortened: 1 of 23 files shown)


BbrSofiane (commented):
- [X] Works as intended
- [ ] Listed in changelog
- [X]  Git messages

@regisb Am I right to assume that refactor changes that don't affect the end-user don't need to be in the changelog?

regisb:
Thanks for the review @BbrSofiane! Yes, only major, backward compatibility-breaking refactor changes should appear in the changelog. The changelog is mostly designed for tutor users and plugin developers, such that they know what comes with each release.

PR #418, opened 2021-04-07
docs: mention MySQL external version

c3ho:
See: https://discuss.overhang.io/t/managed-rds-mysql-8-0-not-compatible-with-current-docker-compose/1429/3

When running the command `tutor local init` with MySQL >5.7, the script will error as the `IDENTIFIED BY` syntax is not supported for the `GRANT` command anymore in later versions of MySQL.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -123,6 +123,9 @@ By default, a running Open edX platform deployed with Tutor includes all necessa
     MYSQL_ROOT_USERNAME: <root user name>
     MYSQL_ROOT_PASSWORD: <root user password>
 
+.. note::
+    When configuring an external MySQL database, please make sure it is using version 5.7.
+
 Elasticsearch
 *************
 


BbrSofiane (approved):
This looks good to me!

docs/configuration.rst
```diff
@@ -109,6 +109,8 @@ Nginx is used to route web traffic to the various applications and to serve stat
 
 MySQL
 *****
+.. note::
+    When configuring an external MySQL database, please make sure it is using version 5.7.
```
regisb, line 113: Could you move this note further down? There is a dedicated paragraph on running a separate MySQL database, so I think you should move it after the list of config variables (i.e: line 126).

regisb (changes requested):

docs/configuration.rst
c3ho: No problem, moved the note to after the envs portion

regisb (approved):

regisb:
LGTM! @c3ho can you please squash your commits?

c3ho:
@regisb Thanks, sorry I missed your message! Squashed and ready to go.

regisb:
thanks for the pr @c3ho :)

PR #419, opened 2021-04-07
feat: Openedx Dockerfile enhancements

ericfab179:
As discussed here https://discuss.overhang.io/t/proposal-for-addition-of-some-patches/1309/4?u=eric.herrera, this PR adds patches to openedx Dockerfiles which make more flexible the installation of extra/private python requirements.

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -35,6 +35,10 @@ RUN mkdir -p /openedx/edx-platform && \
     git clone $EDX_PLATFORM_REPOSITORY --branch $EDX_PLATFORM_VERSION --depth 1 /openedx/edx-platform
 WORKDIR /openedx/edx-platform
 
+{% if patch("openedx-dockerfile-git-patches-default") %}
+# Custom edx-platform default patches
+{{ patch("openedx-dockerfile-git-patches-default") }}
+{% else %}
 # Patch edx-platform
 # Make it possible to disable learner records globally
 # https://github.com/edx/edx-platform/pull/25182
@@ -48,6 +52,7 @@ RUN curl https://github.com/overhangio/edx-platform/commit/bc0ab09f9945bd14aa6be
 # https://github.com/edx/edx-platform/pull/25957
 # https://github.com/overhangio/edx-platform/tree/overhangio/upgrade-django-pipeline
 RUN curl https://github.com/overhangio/edx-platform/commit/1c733e3ba11249cf16358684169e6137a308e796.patch | git apply -
+{% endif %}
 
 ###### Download extra locales to /openedx/locale/contrib/locale
 FROM minimal as locales
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -87,6 +87,8 @@ RUN pip install django-redis==4.12.1
 # Install uwsgi
 RUN pip install uwsgi==2.0.19.1
 
+{{ patch("openedx-dockerfile-post-python-requirements") }}
+
 # Install private requirements: this is useful for installing custom xblocks.
 COPY ./requirements/ /openedx/requirements
 RUN cd /openedx/requirements/ \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
+- [Feature] Add patches to extend python requirements installation process in openedx and openedx-dev Dockerfiles.
 
 ## v11.2.6 (2021-04-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
+
 - [Feature] Add patches to extend python requirements installation process in openedx and openedx-dev Dockerfiles.
+- [Improvement] Apply edx-platform patches during Docker image build using tutor patch 'openedx-dockerfile-git-patches-default'.
 
 ## v11.2.6 (2021-04-09)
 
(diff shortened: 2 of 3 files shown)


regisb (changes requested):
Thanks for this PR @ericfab179! Could you please:

1. rephrase the commit message to fall below the 80 characters limit. Also, add a thorough commit description. Basically, if you find yourself having to add extra words to the PR description, then it means that some information is probably missing from the commit description.
2. add a changelog entry. 

Also, did you want to make it optional to apply edx-platform patches, as suggested [here](https://discuss.overhang.io/t/proposal-for-addition-of-some-patches/1309/4?u=regis)? If yes, feel free to add a commit to this PR.

ericfab179:
@regisb Thanks for your suggestion. I already addressed your comments, let me know if it is better now.

tutor/templates/build/openedx/Dockerfile
```diff
@@ -36,20 +36,24 @@ RUN mkdir -p /openedx/edx-platform && \
...
+        # https://github.com/edx/edx-platform/pull/25771
+        # https://github.com/overhangio/edx-platform/tree/overhangio/fix-paver-warning
+        curl https://github.com/overhangio/edx-platform/commit/bc0ab09f9945bd14aa6be1dbbf928cce58f079d2.patch | git apply -; \
+        # Fix js upload in scorm packages by upgrading django-pipeline
+        # https://github.com/edx/edx-platform/pull/25957
+        # https://github.com/overhangio/edx-platform/tree/overhangio/upgrade-django-pipeline
+        curl https://github.com/overhangio/edx-platform/commit/1c733e3ba11249cf16358684169e6137a308e796.patch | git apply -; \
+    else \
+        echo "Skip application of edx-platform repo patches since EDX_PLATFORM_VERSION does not match OPENEDX_COMMON_VERSION"; \
+    fi
```
regisb, line 56: Thanks for adding this change @ericfab179! We're almost there. You'll need to rebase on top of the newly-released koa.3 version. But I also wanted to have your opinion on something: I think we should only exclude the patches that are already merged in koa.master, but not the custom bugfixes that are not yet present in koa.master. Thus, I would suggest the following:

```
RUN if [ "$EDX_PLATFORM_VERSION" = "{{ OPENEDX_COMMON_VERSION }}" ]; then \
        # Security patches
        curl https://github.com/overhangio/edx-platform/commit/a4369d300b85d94442844443a3d258a2b516955c.patch | git apply -; \
    else \
        echo "Skip application of edx-platform repo patches since EDX_PLATFORM_VERSION does not match OPENEDX_COMMON_VERSION"; \
    fi
# Make it possible to disable learner records globally
# https://github.com/edx/edx-platform/pull/25182
# https://github.com/overhangio/edx-platform/tree/overhangio/disable-learner-records-from-settings
RUN curl https://github.com/overhangio/edx-platform/commit/bd038bab3cf02df147e754f7743e46b68b43bac8.patch | git apply -
...
```

Basically, the idea is to allow forking from anywhere on koa.master, while still allowing custom bugfixes. But I want to make sure that this approach would work for your team with your fork. What do you think?

jfavellar90: @regisb thanks for the review. I was looking at this change with @ericfab179 and I was wondering if we can still have an option to bypass all these patching operations. We still have a level of customization that increases the chance of clashing with any of the patches. The issue would be those forks from koa.master not getting the custom bugfixes, but I think that if a fork needs to be maintained, these fixes could be included directly in the repo by the maintainer. Just to finish, Could it be considered a plugin for patches?

regisb: I understand your problem @jfavellar90: you need to run a custom fork without having to worry whether patches will apply correctly -- something you currently have no control over. What I need is to ensure that most people run edx-platform with the security patches and the bugfixes: this should include people who run a fork of koa.master. Also, this should happen by default and not require the installation of a custom plugin.

To resolve both our situations, I propose an innovative solution. I just tested it and it works. Here's what it looks like:

```
FROM minimal as code
ARG EDX_PLATFORM_REPOSITORY=https://github.com/edx/edx-platform.git
ARG EDX_PLATFORM_VERSION={{ OPENEDX_COMMON_VERSION }}
RUN mkdir -p /openedx/edx-platform && \
    git clone $EDX_PLATFORM_REPOSITORY --branch $EDX_PLATFORM_VERSION --depth 1 /openedx/edx-platform
WORKDIR /openedx/edx-platform

{% if patch("openedx-dockerfile-git-patches-default") %}
# Custom edx-platform default patches
{{ patch("openedx-dockerfile-git-patches-default") }}
{% else %}
# Patch edx-platform
# Make it possible to disable learner records globally
# https://github.com/edx/edx-platform/pull/25182
# https://github.com/overhangio/edx-platform/tree/overhangio/disable-learner-records-from-settings
RUN curl https://github.com/overhangio/edx-platform/commit/bd038bab3cf02df147e754f7743e46b68b43bac8.patch | git apply -
# Fix inconvenient pavelib warning
# https://github.com/edx/edx-platform/pull/25771
# https://github.com/overhangio/edx-platform/tree/overhangio/fix-paver-warning
RUN curl https://github.com/overhangio/edx-platform/commit/bc0ab09f9945bd14aa6be1dbbf928cce58f079d2.patch | git apply -
# Fix js upload in scorm packages by upgrading django-pipeline
# https://github.com/edx/edx-platform/pull/25957
# https://github.com/overhangio/edx-platform/tree/overhangio/upgrade-django-pipeline
RUN curl https://github.com/overhangio/edx-platform/commit/1c733e3ba11249cf16358684169e6137a308e796.patch | git apply -
{% endif %}
```

What we are doing here is to create a new "openedx-dockerfile-git-patches-default" patch. If some plugin implements this patch, then we scrap the default security and bugfix patches. That way, you only have to implement a `# noop` patch in your plugin to get rid of custom patches.

Would this work for you?

ericfab179: @regisb I updated the PR with your new idea which I think suites us perfectly.

regisb (approved):

regisb:
Thanks a lot for your patience @ericfab179 and @jfavellar90!

PR #420, opened 2021-04-07
feat: upgrade all services to open-release/koa.3

regisb:
We remove security patches and custom fixes which are now part of koa.3.

We take the opportunity to make it possible to build the openedx Docker image
without relying on a corresponding openedx-i18n repo tag: often, we want to
test whether the image simply builds successfully, and we don't need up-to-date
translations. For those cases, it's now possible to pass the `-a
OPENEDX_I18N_VERSION=oldertag` build argument.


regisb:
@overhangio/tutor-developers Since this change introduces a new [security patch](https://discuss.openedx.org/t/security-im-patch-for-xss-fixes-10/4702) I would like to merge this as soon as possible. Could you please review?

BbrSofiane:
> For those cases, it's now possible to pass the `-a OPENEDX_I18N_VERSION=oldertag` build argument.

Should this also be mentioned in the release notes?

pcliupc (approved):
Looks good to me.

regisb:
> Should this also be mentioned in the release notes? Otherwise looks good to me.

@BbrSofiane Yes, you're right. I'll amend the changelog and merge.

PR #425, opened 2021-04-23
fix: YamlParamType supports line terminators

ericfab179:
This fix allows using a multiple line formatted Yaml string as input for setting a Tutor config value.


regisb (approved):
Looks great! Can you please add a line to the changelog?

regisb (approved):

Issue #430, opened 2021-05-12
Error when run tutor command if click 8.0.0 is installed


pcliupc:
**Bug description**
Tutor installation requires click>=7.0, but if click 8.0.0 is installed, error happens when run tutor command.

**How to reproduce**
In python 3.7 env, pip install tutor from source (master branch), and check click version and make sure click==8.0.0 is installed. Run `tutor config printroot`, and below error happens:

```
Traceback (most recent call last):
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/bin/tutor", line 11, in <module>
    load_entry_point('tutor-openedx', 'console_scripts', 'tutor')()
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 490, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2859, in load_entry_point
    return ep.load()
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2450, in load
    return self.resolve()
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2456, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Users/liupengcheng/develop/fira/lilac/tutor/tutor/commands/cli.py", line 6, in <module>
    import click_repl
  File "/Users/liupengcheng/develop/fira/tutor_lilac_env/lib/python3.7/site-packages/click_repl/__init__.py", line 6, in <module>
    import click._bashcomplete
ModuleNotFoundError: No module named 'click._bashcomplete'

```
**Environment**

**Additional context**

PR #431, opened 2021-05-12
fix: click 8.0 is not compatible with click_repl.

pcliupc:
There is an [open issue](https://github.com/overhangio/tutor/issues/430) related to this pr.


regisb:
Thanks for this PR @pcliupc! I missed it and opened https://github.com/overhangio/tutor/pull/432 instead, so I'll close this one now.

PR #433, opened 2021-05-14
feat: Make redis db configurable.

pcliupc:
Before this ci, redis db 0 is used for celery broker (by default)
and db 1 is used for cache. Now redis db 1 is used for both cache
and celery if 'REDIS_DB' is not configured explicitly.


tutor/templates/config.yml
```diff
@@ -71,6 +71,7 @@ PLATFORM_NAME: "My Open edX"
 PLUGINS: []
 REDIS_HOST: "redis"
 REDIS_PORT: 6379
+REDIS_DB: 1
```
regisb, line 74: This DB would only be used for edx-platform, so if we decide to add this extra parameter, I suggest we rename it to `OPENEDX_REDIS_DB`.

regisb (commented):
This PR is merging both cache and celery databases, which is a dangerous change. These databases were different for a purpose. I am not 100% sure that there are no name conflicts in these databases. Also, I'd like to be able to clear the static asset database without impacting the celery task database.

Previously, the cache/celery databases were completely separated because they lived in different data stores (memcached and redis). I'm not sure it's a good idea to merge them. Can you explain what's your usage scenario?

If we do want to proceed, we'll need to add a breaking change warning to the changelog.

tutor/templates/config.yml
pcliupc: It might also be used by plugins, such as discovery and ecommerce.

pcliupc:
> This PR is merging both cache and celery databases, which is a dangerous change. These databases were different for a purpose. I am not 100% sure that there are no name conflicts in these databases. Also, I'd like to be able to clear the static asset database without impacting the celery task database.
> 
> Previously, the cache/celery databases were completely separated because they lived in different data stores (memcached and redis). I'm not sure it's a good idea to merge them. Can you explain what's your usage scenario?
> 
> If we do want to proceed, we'll need to add a breaking change warning to the changelog.

I see your considerations and concerns. I agree with you that it is more safe to keep cache/celery data in separate databases, however, we need separate configurations for them each.

tutor/templates/config.yml
regisb: Right. I see that the discovery cache is configured (in Tutor) to use DB=1, while ecommerce uses DB=0 for celery tasks; the latter is almost certainly incorrect. I agree with you that we should introduce configuration settings. I propose that we use separate databases for each, with the following names: `OPENEDX_CACHE_REDIS_DB`, `OPENEDX_CELERY_REDIS_DB`, `DISCOVERY_CACHE_REDIS_DB`, `ECOMMERCE_CELERY_REDIS_DB`.

Would you like to make the discovery pull request? Will you update this PR with this proposed change?

pcliupc: sure, I will make pr for discovery.

regisb (changes requested):
Almost there! Can you please add a line to the changelog?

pcliupc:
> Almost there! Can you please add a line to the changelog?

PR updated.

regisb (approved):

regisb:
This should be ready to merge. Can you please rebase?

PR #437, opened 2021-05-19
lilac: Added configuration changes for mongo atlas style clusters

zakum1:
Configuration for a Mongo Atlas cluster was discussed [in the forum](https://discuss.overhang.io/t/connecting-to-a-mongodb-cluster-protocol-mongodb-srv/1272).

Mongo Atlas clusters use SRV DNS records to record the cluster structure, which allows for a URI with a `mongodb+srv` protocol to point to the cluseter. This approach can be used with the forum, but for the LMS/CMS the python mongodb client uses an older style of configuration, so two different approaches are implemented for connecting to the cluster.

Getting forum to work with a mongo cluster required two new environment variables to env/build/forum/Dockerfile:

```dockerfile
ENV MONGODB_PROTOCOL "mongodb"
```

`MONGODB_PROTOCOL` can be set to mongodb+srv for a cluster

This change is applied to `env/build/forum/bin/docker-entrypoint.sh`

If the `MONGODB_PROTOCOL` is set to `mongodb+srv` then there are two key changes:

1.    The port must not be included in the URL
2.    The host cannot be used as a true server name for “waiting” because it is a DNS entry with only SRV records. I have elected not to wait at all for a cluster because it is unlikely that this cluster is starting up at the same time as the forum, because it is typically provisioned beforehand.

In addition, we add these three parameters to `config.yml`

`MONGODB_REPLICA_SET` replica set name for the cluster
`MONGODB_USE_SSL` indicates if the cluster uses SSL (`true` or `false`)
`MONGODB_AUTH_SOURCE` the database name of the authentication database on the cluster

These are for the LMS / CMS connecting to mongo - openedx supports these parameters in the pymongo initialisation, we just need to update them in the parameters that get passed in as `mongodb_parameters`

To achieve this, `tutor/templates/apps/openedx/settings/partials/common_all.py` is updated to handle these additional variables.

An example of setting these variables for a cluster in `config.yml`:

```yml
RUN_MONGODB: true
MONGODB_HOST: "example-dev-shard-00-00.psvut.mongodb.net:27017,example-dev-shard-00-01.psvut.mongodb.net:27017,example-dev-shard-00-02.psvut.mongodb.net:27017"
MONGODB_DATABASE: "openedx"
MONGODB_PORT: 27017
MONGODB_USERNAME: "openedx"
MONGODB_PASSWORD: "pwd123"
MONGODB_USE_SSL: true
```

Which will produce these parameters in the LMS / CMS settings script:

```py
mongodb_parameters = {
    "host": "example-dev-shard-00-00.psvut.mongodb.net:27017,example-dev-shard-00-01.psvut.mongodb.net:27017,example-dev-shard-00-02.psvut.mongodb.net:27017",
    "port": 27017,
    "user": "openedx",
    "password": "pwd123",
    "db": "openedx",
    "ssl": True,
}
```

Until these parameters are replaced by the URI connection string, a replica set is also required here, but this can be added by a plugin for the short 
interim:

```
mongo_parameters['replicaSet'] = 'replicat-set-name'
mongo_parameters['authsource'] = 'admin'
```


zakum1:
@regis, you had two requested changes in the original [PR 415](https://github.com/overhangio/tutor/pull/415):

1. Change the default for `MONGODB_USE_SSL` to `false` in the template
2. The question on whether the `MONGODB_REPLICA_SET` is really needed.

I have done the first in this PR.

On the second, you suggested that I remove it and use a plugin, on the basis that open edx will soon switch to a URI style of connection string for mongo, which does not require the replica set when `mongo+srv` is used in the URI. I will do that, and it then also makes sense to do the same for the auth source, since this is also associated with a cluster, and not required when URI style connection strings are used.

zakum1:
Actually, if we follow through the recommendations by  @regisb to their logical conclusion the best thing here is to update open edx to take a mongo connection uri (eg,  MONGODB_URI) and then all of the proposed new parameters become redundant. In the interim we can set the parameters with a plug-in.

regisb:
Hey @zakum1, thanks a lot for the thorough writeup and your patience in this PR. I like these changes, and I think they will go to Lilac.

> open edx will soon switch to a URI style of connection string for mongo, which does not require the replica set when mongo+srv is used in the URI.

This is the last remaining unclear point for me. From my understanding of the pymongo package and its usage in edx-platform, it should already be possible to use a URI style of connection string for mongodb: https://pymongo.readthedocs.io/en/3.10.1/atlas.html

The first constructor parameter is actually `host`: this goes to show (I think?) that we can pass `mongodb+srv` URIs to the MongoClient class. If you tried that, what didn't work?

zakum1:
Hi @regisb, sorry for the delay in getting to this. What you say looks like it should work. I will give it a try and give feedback here.

zakum1:
Hey @regisb, your proposal basically work, but with one small change that needs to be made. Using a `mongo+srv://` style URI in the `host` requires an additional python module that is not installed by openedx: `dnspython`.

For reference the error is:

```
pymongo.errors.ConfigurationError: The "dnspython" module must be installed to use mongodb+srv:// URIs
```

I guess the best is to add this dependency to openedx, but in the interim it could be done in tutor?

regisb:
That's great news @zakum1! I'm all for installing dnspython inside the openedx Docker image if it allows us to access mongodb+srv urls in the LMS/CMS. For that, you should *not* add dnspython to tutor/requirements/base.in: these requirements are installed on the host, and not in the Docker image. Instead, you should add a `RUN pip install dnspython=...` line to the openedx Dockerfile.

regisb:
FYI I'm going to merge the lilac branch into master. It's ok if this PR does not get merged before then: just change the target branch to "master".

regisb:
Merde! This PR was automatically closed when I deleted the lilac branch. Please re-open it with the master branch as a target @zakum1.

Issue #439, opened 2021-05-21
Discussion notification emails are not being sent


regisb:
**Bug description**

See description here: https://discuss.overhang.io/t/does-tutor-support-ace/1451/3?u=regis

When answering a forum conversation in the LMS, users should receive email notifications (provided the feature is enabled). This is not the case: the email sending task is processed by celery, but the SMTP server does not receive any request. 

**How to reproduce**

1. Enable forum notifications ([instructions](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/enable_discussion_notifications.html)) by adding `{"enable_forum_notifications":true}` to the site configuration.
2. Make a comment in a forum
3. With a different user, write a reply to the initial comment.

A notification email should be sent, but it's not the case.

To debug this issue, it is recommended to run the LMS without workers (`CELERY_ALWAYS_EAGER = True`).

pcliupc:
In my environment, I found settings.ACE_ROUTING_KEY was set to 'edx.core.low', that's the reason why the send mail task is never consumed.  After setting ACE_ROUTING_KEY to "edx.lms.core.default" in lms.env.json, the email notification works. @regisb

PR #441, opened 2021-05-25
fix discussion notification not work.

pcliupc:
Discussion email notification relies on edx_ace and celery routing
key for ace is not set properly by default, which causes the send
mail task is never consumed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Fix discussion notification not work.
 - [Improvement] Better error logging when loading a plugin from an incompatible version.
 
 ## v11.2.11 (2021-05-18)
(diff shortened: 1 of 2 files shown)


regisb:
Great catch @pcliupc! It turns out that the `ACE_ROUTING_KEY` is not being set to its expected value of 'edx.lms.core.default' in edx-platform/lms/common.py. Instead, it is set to "edx.core.low" in the edx_ace common settings.We should open a PR on edx-platform to:

1. Remove the `ACE_ROUTING_KEY` from lms/envs/common.py 
2. Change the default in the edx_ace default settings.

Would you like to open such a PR or do you want me to do it?

Meanwhile, this PR is almost ready to merge. Please add a line to the Changelog. Also, add "Close #439" somewhere in the commit message such that we can find the relevant issue later.

Thanks for investigating this!

regisb (approved):

regisb:
:+1: see: https://github.com/edx/edx-platform/pull/27719

Issue #284, opened 2020-01-31
CodeJail Support


fdns:
**Is your feature request related to a problem? Please describe.**
I was looking into implementing CodeJail (https://github.com/edx/codejail), as it can be an attack vector for tutor instances (and openedx in general), as it´s in charge of running untrusted code from studio and LMS (Ex: Custom Python-Evaluated Input).

Currently CodeJail is not supported in tutor, and is executed on the openedx context, which have a lot more permissions.

I will be out out this month and can´t continue working on this feature, so I will leave what I have done to support it, if anyone can continue (I will be able to continue on march).

**Current Work**
To implement codejail in the docker environment, first I installed manually the codejail code in docker as a build step, adding the sandbox user to the environment.
```
RUN useradd -ms /bin/bash sandbox && \
    git clone https://github.com/edx/codejail /openedx/sandbox && \
    virtualenv /openedx/sandbox/env && \
    /openedx/sandbox/env/bin/pip install -r /openedx/sandbox/requirements/sandbox.txt && \
    chown -R sandbox:sandbox /openedx/sandbox && \
    apt-get update && apt-get install sudo
```

After this, you must add in the cms.env and lms.env config files the codejail items with the correct paths.
```
  "CODE_JAIL": {
    "python_bin": "/openedx/sandbox/env/bin/python2",
    "user": "sandbox",
    "limits": {
      "CPU": 1,
      "VMEM": 268435456,
      "REALTIME": 3,
      "FSIZE": 1048576
    }
  },
```

After this, you must add the apparmor profile to jail the code executed by the sandbox (In the host system). I used the original apparmor profile included in the repository, and combined it with the docker-default environment, so we can execute the docker instance with the default security applied. This is done by creating `/etc/apparmor.d/containers/docker-edx-sandbox` with the following content, and activating it by running `sudo apparmor_parser -r -W /etc/apparmor.d/containers/docker-edx-sandbox`
```
#include <tunables/global>

profile docker-edx-sandbox flags=(attach_disconnected,mediate_deleted) {
    #include <abstractions/base>

    network,
    capability,
    file,
    umount,
    signal (receive) peer=unconfined,
    signal (receive) peer=snap.docker.dockerd,
    signal (send,receive) peer=docker-edx-sandbox,

    deny @{PROC}/* w,   # deny write for all files directly in /proc (not in a subdir)
    # deny write to files not in /proc/<number>/** or /proc/sys/**
[...]

PR #443, opened 2021-05-31
security: disable python-evaluated input by default

regisb:
As described in issue #284, tutor does not come with codejail enabled out of
the box. Actually, we don't even have a working plugin, yet. To prevent users
from running unsafe code, we explicitely disable python-evaluated input by
disabling the "python" interpreter. This might break some courses; thus, this
is a non-backward compatible change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


BbrSofiane (approved):
- [X] Code works as intended, is readable and properly documented (in this order).
- [X] Entry added to the changelog
- [X] Git message isinformative and well written

PR #446, opened 2021-06-03
K8s namespace permission improvement

regisb:
In most cases, it makes very little sense to edit the namespace that an
application is running in. Quite often, users are granted access to just one
namespace and don't have the necessary rights to edit the namespace -- and for
good security reasons. In such cases, the k8s namespace object already exists
and there is no need for the user to edit or create it. Here, what we do is
that we create the namespace only if it does not exist. This should solve quite
a few permission issues, notably for Openshift users.

This is ready for review @overhangio/tutor-developers. cc @menardorama

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


regisb:
I'm going to merge this such that it lands in Tutor before Lilac comes out. Please comment if you have other thoughts.

Issue #449, opened 2021-06-14
The Mongodb database name of the forums is hardcoded and cannot be modified


regisb:
**Is your feature request related to a problem? Please describe.**

People running Tutor for Open edX SaaS want to run run a single Mongodb cluster for all their customers. For that, they need to create different databases on the same server. The problem is that the MongoDb database is hardcoded to "cs_comments_service".

**Describe the solution you'd like**

Introduce a `FORUM_MONGODB_DATABASE` that will default to "cs_comments_service" and will be used in the forum entrypoint.

**Additional context**

This issue was first reported here: 
https://discuss.overhang.io/t/tutor-k8s-deployment-mongodb-password/1586
https://discuss.overhang.io/t/setup-open-edx-saas-based-with-tutor/1597

ChrisChV:
Hi @regisb!
Is anyone already working on this issue? I would like to contribute solving it

regisb:
No one is, and you are most welcome to start working on it :)

pcliupc:
> No one is, and you are most welcome to start working on it :)

@regisb @ChrisChV I have met this issue before and I would like to share my experience here:

1. add forum index name config in tutor and set env in k8s deployments
2. patch forum code to use the index config (can be set in cs_comments_service/models/content.rb)

regisb:
Oh damn, @pcliupc you mean that the index name is hardcoded in the forum service? @ChrisChV do you feel like opening an upstream pull request to address that?

pcliupc:
> Oh damn, @pcliupc you mean that the index name is hardcoded in the forum service? @ChrisChV do you feel like opening an upstream pull request to address that?

yes, in latest version, the indices for different types (such as comment, comment_thread) are separated but still hard coded.

pcliupc:
> Oh damn, @pcliupc you mean that the index name is hardcoded in the forum service? @ChrisChV do you feel like opening an upstream pull request to address that?

Yes, the index name is hard coded. But sorry for mistaking mongodb name as index name :(
The mongodb name can be configured through set MONGOHQ_URL in tutor/templates/build/forum/bin/docker-entrypoint.sh.

regisb:
Right! It's not too bad if the index/collection names are hard-coded, as long as we can modify the database name.

ChrisChV:
Yes, I agree. First we have to resolve the hard coded database name.

regisb:
This was resolved by https://github.com/overhangio/tutor/pull/453

PR #453, opened 2021-06-16
feat: FORUM_MONGODB_DATABASE config val added to fix the hardcoded forum database name

ChrisChV:
Issue: https://github.com/overhangio/tutor/issues/449

- In `config.yml` the new value `FORUM_MONGO_DB_DATABASE` was added with `cs_comments_service` as default value.
- In `docker-entrypoint.sh` of forum I changed the hardcoded `cs_comments_service` with the new config value.
- Multiple `.yml` files changed to handle the new config value.
- `CHANGELOG.md` updated.


tutor/templates/build/forum/Dockerfile
```diff
@@ -48,5 +48,6 @@ ENV MONGODB_AUTH ""
 ENV MONGOID_AUTH_MECH ""
 ENV MONGODB_HOST "mongodb"
 ENV MONGODB_PORT "27017"
+ENV FORUM_MONGODB_DATABASE "cs_comments_service"
```
pcliupc, line 51: Do you want to use {{ FORUM_MONGODB_DATABASE }} here?

regisb: > Do you want to use {{ FORUM_MONGODB_DATABASE }} here?

Actually, no: if we do, the Docker image will depend on the actual Tutor configuration, which is not a good thing. I think it's better if we keep it like this. I would simply rename the environment variable to `MONGODB_DATABASE`: since we are already in the context of the forum, it's not very useful to add the `FORUM_` prefix.

(and don't forget the rename the variable in the other templates, too)

regisb (changes requested):
Could you just rename the `FORUM_MONGODB_DATABASE` environment variable to `MONGODB_DATABASE`? The config variable should still be named `FORUM_MONGODB_DATABASE`, though.

ChrisChV:
@regisb done

regisb:
Looking good! ideally we should add some docs about the forum configuration variables: https://docs.tutor.overhang.io/configuration.html But it's outside the scope of this PR.

Can you please squash your commits? Then we'll be ready to merge.

tutor/templates/build/forum/Dockerfile
pcliupc: Make sense.

regisb (approved):

Issue #447, opened 2021-06-04
The default ACE emails look terrible


BbrSofiane:
**Is your feature request related to a problem? Please describe.**
The default rendered ACE emails currently require additional configuration to be usable out of the box as described in this [post](https://discuss.overhang.io/t/does-tutor-support-ace/1451/7?u=bbrsofiane).

1. There is no default logo
2. There is no default email address

**Describe the solution you'd like**
Add default sane values that will improve the usability of the platform out of the box.
- [ ] Set a default platform logo
- [x] Set a default mailing address

BbrSofiane:
There is a default logo set in [common.py](https://github.com/edx/edx-platform/blob/c0bed879541b628f819d71412ea6d29da1fa9e46/lms/envs/common.py#L4718). Should the steps required to set the logo added somewhere?

regisb:
Yes, I think it's important that we provide a better default for this setting. For instance: the URL of the LMS logo.

regisb:
@BbrSofiane do you think you will have time to work on this in the coming weeks?

BbrSofiane:
@regisb yes, I will have time to finish it this weekend.

pcliupc:
I think it will be better if the from address of the email can also be configured (the email address is used by default, for example, if the sending email address is contact@overhang.io, 'contact' is shown as from address). The from address can be configured through site_configuration['email_from_address'] in LMS.

regisb:
Wait, is this issue actually closed @BbrSofiane? I believe we still need to address the question of the logo, right?

> I think it will be better if the from address of the email can also be configured (the email address is used by default, for example, if the sending email address is contact@overhang.io, 'contact' is shown as from address). The from address can be configured through site_configuration['email_from_address'] in LMS.

@pcliupc As you point out, users always have the possibility of customizing the "from" address via the site configuration, right? In tutor there is now a `site-configuration` utility for that:

    tutor local run lms site-configuration set -d mylms.com email_from_address me@nospam.com

BbrSofiane:
>  I believe we still need to address the question of the logo, right?

Yes, but I am not sure I am clear on what to do with 

> For instance: the URL of the LMS logo.

Right now the default value for the email logo is the same as the LMS logo (`DEFAULT_EMAIL_LOGO_URL = 'https://edx-cdn.org/v3/default/logo.png'`). 

Are you suggesting we add `LOGO_URL_PNG ='https://edx-cdn.org/v3/default/logo.png'`?

[...]

PR #456, opened 2021-06-20
fix: set default CONTACT_MAILING_ADDRESS

BbrSofiane:
Setting the default CONTACT_MAILING_ADDRESS to be set as contact@{{ LMS_HOST }}.

Added the variable to `config.yml` to allow the value to be changed.

Fix for #447, @overhangio/tutor-developers this is ready for review.


tutor/templates/config.yml
```diff
@@ -21,6 +21,7 @@ RUN_REDIS: true
 RUN_SMTP: true
 CMS_HOST: "studio.{{ LMS_HOST }}"
 CONTACT_EMAIL: "contact@{{ LMS_HOST }}"
+CONTACT_MAILING_ADDRESS: "contact@{{ LMS_HOST }}"
```
regisb, line 24: I'd rather avoid adding new Tutor config variables for every LMS/CMS setting. Instead, we should create good defaults that people can override with plugins. Could we default to `{{ PLATFORM_NAME }} - {{ CONTACT_EMAIL }}` for the mailing address?

CHANGELOG.md
```diff
@@ -48,7 +49,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## v11.2.8 (2021-04-27)
 
-- [Bugfix] Fix parsing of YAML-formatted values in ``config save --set KEY=VALUE`` commands, in use for instance with Ecommerce.
+- [Bugfix] Fix parsing of YAML-formatted values in `config save --set KEY=VALUE` commands, in use for instance with Ecommerce.
```
regisb, line 52: I distinctly remember me adding double ticks to all verbatim statements. I wonder why I did that...

tutor/templates/apps/openedx/config/lms.env.json
```diff
@@ -27,6 +27,7 @@
   "CMS_BASE": "{{ CMS_HOST }}",
   "LMS_BASE": "{{ LMS_HOST }}",
   "CONTACT_EMAIL": "{{ CONTACT_EMAIL }}",
+  "CONTACT_MAILING_ADDRESS": "{{ PLATFORM_NAME }}@{{ LMS_HOST }}",
```
regisb, line 30: Can I suggest moving this settings to the py settings? I'd rather avoid adding stuff the the json config files, when possible. This setting will be necessary both in the LMS and the CMS, so you can add it to the common partial settings file.

regisb, line 30: Is this a good value? Can I suggest instead `{{ PLATFORM_NAME }} - {% if ENABLE_HTTPS %}https{% else %}http{% endif %}://{{ LMS_HOST }}`?

regisb (changes requested):

CHANGELOG.md
BbrSofiane: Oh that's probably my IDE, I'll revert that.

tutor/templates/apps/openedx/config/lms.env.json
BbrSofiane: My understanding is that the mailing address is meant to be an email address, is that incorrect?
image

BbrSofiane: Yes, no problem. What's a good rule of thumb to make decisions when it comes to where configurations settings should be placed?

pcliupc (approved):
LGTM.

tutor/templates/apps/openedx/config/lms.env.json
regisb: No, I believe that the `CONTACT_MAILING_ADDRESS` is the physical mailing address, as indicated by the variable usage in the edx-platform code base:

```
$ git grep CONTACT_MAILING_ADDRESS
cms/envs/common.py:CONTACT_MAILING_ADDRESS = _('Your Contact Mailing Address Here')
common/djangoapps/student/email_helpers.py:            settings.CONTACT_MAILING_ADDRESS
lms/djangoapps/instructor/enrollment.py:        settings.CONTACT_MAILING_ADDRESS
lms/envs/common.py:CONTACT_MAILING_ADDRESS = 'SET-ME-PLEASE'
openedx/core/djangoapps/ace_common/template_context.py:            'CONTACT_MAILING_ADDRESS', site=site, site_config_name='contact_mailing_address'),
openedx/core/djangoapps/schedules/docs/README.rst:    CONTACT_MAILING_ADDRESS = '<insert_physical_address_here>'
openedx/core/djangoapps/schedules/tests/test_resolvers.py:    @override_settings(CONTACT_MAILING_ADDRESS='123 Sesame Street')
openedx/core/djangoapps/schedules/tests/test_resolvers.py:    @override_settings(CONTACT_MAILING_ADDRESS='123 Sesame Street')
```

regisb: Some values from the json files are parsed in the edx-platform upstream settings files to define multiple setting values. For instance, `PLATFORM_NAME` is used to define the `CC_MERCHANT_NAME` setting. In such cases, it's better to add the values to the env.json files to avoid code duplication.

Some values must be present in the env.json files, otherwise causing a crash. This is the case of the `CACHES` variable for instance.

Feature flags are also defined in the env.json files because... well I'm not sure. Some feature flags are also used in multiple places in the settings. So if we defined all feature flags in the settings modules, it would cause some duplication. I'd rather avoid splitting feature flags across both the json and the py settings files.

In all other cases, it's better if settings are defined directly in the python modules, to avoid code indirections.

CHANGELOG.md
regisb: No but I think your IDE is actually right! In markdown, inline code is marked by single quotes, i.e: \`some inline code\`.

Sorry for not making this clearer in my initial comment.

tutor/templates/apps/openedx/config/lms.env.json
BbrSofiane: Ahh yeah makes sense!

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -39,6 +39,8 @@
   "port": {{ ELASTICSEARCH_PORT }},
 }]
 
+CONTACT_MAILING_ADDRESS: {{ PLATFORM_NAME }} - {% if ENABLE_HTTPS %}https{% else %}http{% endif %}://{{ LMS_HOST }}
```
regisb, line 42: Wrong copy-paste! :-D This should be written as:

```
CONTACT_MAILING_ADDRESS = "..."
```

regisb: @BbrSofiane will you have time to fix this? After that we should be ready to merge.

BbrSofiane: Fixed! I forgot to push the change.

PR #461, opened 2021-06-25
Swaggification

regisb:
1. Multiple minor improvements to the docs
2. We were transferred ownership of the "tutor" project on pypi \o/ https://pypi.org/project/tutor

For now, it's OK if plugins depend on tutor-openedx, as tutor is listed as an explicit dependency, so this is not a breaking change.


BbrSofiane (approved):
LGTM!

PR #462, opened 2021-06-28
fix: don't refer to obsolete proxy config in docs

regisb:
nginx/openedx.conf was a configuration file provided to configure
proxies on the host. This file no longer ships with Tutor since v11.0.0.

See: https://discuss.overhang.io/t/local-deployment-webproxy/1688


BbrSofiane (approved):
Looks good to me

PR #470, opened 2021-07-11
Revert "fix: empty entrypoints in docker-compose=2.0.0.beta4"

regisb:
This reverts commit 6f04223d0135e1debb005f2a9802f081ed21d203.

It turns out that we cannot ignore "command: []" statements in k8s manifests.
That's because there is no way to clear entrypoint in k8s manifests, but
sometimes we do need to bypass the entrypoints. For instance, the minio client
container sets "mc" as the default entrypoint, which does not work with our job
logic.

As a consequence, Tutor becomes incompatible with docker-compose 2.0.0.beta4.
The "entrypoint must be a string" error is actually an upstream bug:
https://github.com/docker/compose-cli/issues/1848

See:
- The corresponding minio issue: https://github.com/overhangio/tutor-minio/issues/9
- The previous conversation about empty entrypoints: https://discuss.overhang.io/t/undefined-entrypoint-throws-error-in-docker-compose-2-0-0-beta-4/1716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Fix "`sh` is not a recognized command" in some plugins, including minio.
+
 ## v12.0.2 (2021-07-06)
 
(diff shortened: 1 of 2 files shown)


regisb:
this is ready for review @overhangio/tutor-developers

PR #478, opened 2021-08-19
Fix third-party auth example

michaelwheeler:
My understanding is that `THIRD_PARTY_AUTH_BACKENDS` is a list in lms.env.json that is used to generate the value of Django's `AUTHENTICATION_BACKENDS` setting.

### DIFF
--- a/docs/plugins/examples.rst
+++ b/docs/plugins/examples.rst
@@ -48,12 +48,12 @@ Enable SAML authentication
     name: saml
     version: 0.1.0
     patches:
-     common-env-features: |
-        "ENABLE_THIRD_PARTY_AUTH" : true
+      common-env-features: |
+        "ENABLE_THIRD_PARTY_AUTH": true
 
-     openedx-lms-common-settings: |
+      openedx-lms-common-settings: |
         # saml special settings
-        THIRD_PARTY_AUTH_BACKENDS = "third_party_auth.saml.SAMLAuthBackend"
+        AUTHENTICATION_BACKENDS += ["third_party_auth.saml.SAMLAuthBackend", "django.contrib.auth.backends.ModelBackend"]
 
       openedx-auth: |
         "SOCIAL_AUTH_SAML_SP_PRIVATE_KEY": "yoursecretkey",


BbrSofiane (commented):
@michaelwheeler Look at this [example](https://discuss.overhang.io/t/migrating-from-native-install-to-tutor-juniper-tutor-10-x/1533/11#saml-juniper-4) in the forum, is the `SOCIAL_AUTH_SAML_SP_PUBLIC_CERT` also required?

michaelwheeler:
Hi @BbrSofiane thanks for taking a look. I'm not very familiar with the `SOCIAL_AUTH_SAML_SP_PRIVATE_KEY` or `SOCIAL_AUTH_SAML_SP_PUBLIC_CERT` settings (my auth backend isn't using SAML) so I left those lines unchanged. As I understand it those settings are required for a social-auth SAML backend.

docs/plugins/examples.rst
```diff
@@ -48,12 +48,11 @@ Enable SAML authentication
...
-     common-env-features: |
-        "ENABLE_THIRD_PARTY_AUTH" : true
+      common-env-features: |
+        "ENABLE_THIRD_PARTY_AUTH": true
 
-     openedx-lms-common-settings: |
-        # saml special settings
-        THIRD_PARTY_AUTH_BACKENDS = "third_party_auth.saml.SAMLAuthBackend"
+      lms-env: |
+        "THIRD_PARTY_AUTH_BACKENDS": ["third_party_auth.saml.SAMLAuthBackend"]
```
regisb, line 55: Could you please keep the patch in `openedx-lms-common-settings`, and not `lms-env`? When possible, I'd rather avoid adding stuff in the json files. (also, please keep the commend line 55)

docs/plugins/examples.rst
```diff
@@ -48,12 +48,11 @@ Enable SAML authentication
     name: saml
     version: 0.1.0
     patches:
-     common-env-features: |
-        "ENABLE_THIRD_PARTY_AUTH" : true
+      common-env-features: |
+        "ENABLE_THIRD_PARTY_AUTH": true
```
regisb, line 52: :+1:

regisb (changes requested):
Thanks for your PR @michaelwheeler! Can you please address the requested changes?

docs/plugins/examples.rst
michaelwheeler: Thanks for taking a look @regisb! I could change it to something like this...

```yaml
  # saml special settings
  AUTHENTICATION_BACKENDS = ["third_party_auth.saml.SAMLAuthBackend", "django.contrib.auth.backends.ModelBackend"]
```

But it feels a bit odd to me to override `AUTHENTICATION_BACKENDS` in this way, since it seems like `THIRD_PARTY_AUTH_BACKENDS` is specifically intended for the purpose of specifying custom auth backends [in edx-platform](third_party_auth.saml.SAMLAuthBackend). I'm new to the world of Open edX though, so I certainly could be misunderstanding something. Let me know if you'd still like me to proceed and I'll make the appropriate changes.

regisb: Yes, I think that you are right that the patch should be adding entries to the setting, and not override it. Still, I really dislike using the json config file. I think the following solution would be best:

```
  # saml special settings
  AUTHENTICATION_BACKENDS += ["third_party_auth.saml.SAMLAuthBackend", "django.contrib.auth.backends.ModelBackend"]
```

regisb (approved):

regisb:
Can you please squash your commits @michaelwheeler? Then we can merge this.

michaelwheeler:
I think things are all set now @regisb.

regisb:
Thanks for your contribution @michaelwheeler!

PR #486, opened 2021-09-06
docs: using docker-compose.override.yml

regisb:
This feature was previously undocumented.

See: https://discuss.overhang.io/t/override-and-docker-compose/1857

EDIT: in addition, we make it possible to override job-specific configs. To simplify the code, we get rid of the deprecated way of running jobs, which dates back from v3.12.0. See: https://discuss.overhang.io/t/problem-fetching-saml-idp-metadata/1330/23


pcliupc (approved):
LGTM

PR #493, opened 2021-09-15
Fix forum mongo connection

angonz:
Add MONGOID_AUTH_MECH (default ":scram") parameter and include it as environment variable for jobs and deployments when MONGODB_USERNAME and MONGODB_PASSWORD are not empty.


angonz:
I have tested locally and in k8s, using the mongodb container and also using an external MongoDB installation. Works fine, for modulestore and forum. I was able to import demo course, import and export a course, edit content, and add forum posts.

tutor/templates/local/docker-compose.yml
```diff
@@ -70,6 +70,7 @@ services:
       MONGODB_HOST: "{{ MONGODB_HOST }}"
       MONGODB_PORT: "{{ MONGODB_PORT }}"
       MONGODB_DATABASE: "{{ FORUM_MONGODB_DATABASE }}"
+      MONGOID_AUTH_MECH: "{{ MONGOID_AUTH_MECH }}"
```
regisb, line 73: Shouldn't this parameter also be added to docker-compose.jobs.yml?

tutor/templates/config.yml
```diff
@@ -51,6 +51,7 @@ LANGUAGE_CODE: "en"
 MONGODB_HOST: "mongodb"
 MONGODB_DATABASE: "openedx"
 MONGODB_PORT: 27017
+MONGOID_AUTH_MECH: ":scram"
```
regisb, line 54: In cs_comments_service we have the following line:

    config/mongoid.yml:    auth_mech: <%= ENV['MONGOID_AUTH_MECH'].nil? ? ':scram' : ENV['MONGOID_AUTH_MECH'] %>

Thus, `MONGOID_AUTH_MECH: ""` would be a better default value, wouldn't it?

tutor/templates/k8s/deployments.yml
```diff
@@ -155,6 +155,8 @@ spec:
               value: "{{ MONGODB_PORT }}"
             - name: MONGODB_DATABASE
               value: "{{ FORUM_MONGODB_DATABASE }}"
+            - name: MONGOID_AUTH_MECH
+              value: "{% if MONGODB_USERNAME and MONGODB_PASSWORD %}{{ MONGOID_AUTH_MECH }}{% endif %}"
```
regisb, line 159: From my understanding, it is unnecessary to add the `{% if MONGODB_USERNAME and MONGODB_PASSWORD %}` condition. Can you confirm?

regisb, line 159: Why is this value added only to the forum container, and not to the lms/cms?

regisb (changes requested):
@angonz I'm not sure what problem this PR resolves. Can you please describe more precisely your issue and how this PR resolves it? What mechanism do you use to authenticate with Mongodb?

angonz:
Hi @regisb! Thanks for your comments, and sorry for not being clear from the beginning.

The problem is discussed in [this forum post](https://discuss.overhang.io/t/edx-forum-mongodb-authentication/927/10):
By default, Tutor uses a MongoDB container in the same cluster and connects directly using empty username and password. However, if you need to use an external MongoDB, user authentication is required. 
But if you just add MONGODB_USERNAME and MONGODB_PASSWORD settings, authentication fails.

The solution: Setting [MONGOID_AUTH_MECH env to “:scram”](https://discuss.overhang.io/t/edx-forum-mongodb-authentication/927/5) allows authentication via user and password to the external MongoDB.

regisb:
Thanks for the explanations @angonz. I did not understand why we had to define `MONGOID_AUTH_MECH` because in my understanding, the auth mechanism was automatically set to ":scram"  thanks to the following line in cs_comments_service:

    config/mongoid.yml:    auth_mech: <%= ENV['MONGOID_AUTH_MECH'].nil? ? ':scram' : ENV['MONGOID_AUTH_MECH'] %>

However, it turns out that my understanding of ruby's `nil?` method was incorrect:

```
$ MONGOID_AUTH_MECH="" irb
irb(main):001:0> ENV["MONGOID_AUTH_MECH"].nil?
=> false
```

Thus, setting an empty string value for the MONGOID_AUTH_MECH environment variable is not a good idea.

I believe that the right fix should be to remove the `MONGOID_AUTH_MECH` environment variable from tutor entirely. Can you please attempt to do that? Then run:

```
tutor local images build forum
tutor local start -d
```

And check that the forum are still working, both when authentication is enabled and disabled.

angonz:
Looks like it was much easier than what I thought!!

I just commented out `ENV MONGOID_AUTH_MECH ""` from the forum dockerfile template, and seems to be working. I tested also in k8s successfully. Will try to patch the PR now.

regisb:
Sounds good. Can you please squash your commits and add a line to the changelog?

angonz:
Done!

regisb (approved):

mafonso:
I think this might have introduced an issue. 
I started having this error showing up on init, with built-in Mongo, a few hours ago since the new release.

This is an automated tutor release process, that repeats the same steps every-time. 
But I was not pinning the tutor version and it just got 12.1.0 in. 

```
NoMethodError: undefined method `encode' for nil:NilClass
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/user.rb:87:in `encoded_name'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:320:in `first_bare'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:263:in `client_first_message'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:181:in `start'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram.rb:58:in `login'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:249:in `block in authenticate!'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server.rb:267:in `handle_auth_failure!'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:248:in `authenticate!'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:100:in `connect!'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connectable.rb:84:in `ensure_connected'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:270:in `write'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:229:in `deliver'
/openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:144:in `block in dispatch'
```

regisb:
@angonz this is almost certainly why I defined this environment variable in the first place... Can you please quickly open a PR where you re-introduce the same environment variable, but this time set to ":scram"?

regisb:
Ok I made an emergency release v12.1.1 with the patch.

regisb:
The issue might come from the mongoid library which probably also uses the MONGOID_AUTH_MECH environment variable.

angonz:
Did local quickstart work for you? It keeps on failing for me even in 12.1.1. It's strange, because I could catch the fourm container when it was created and the variable was there:
```
# echo $MONGOID_AUTH_MECH
:scram
```
And the Docker file created in env looks good:
```
% grep MONGOID $(tutor config printroot)/env/build/forum/Dockerfile
ENV MONGOID_AUTH_MECH ":scram"
```
But the job fails again.

angonz:
12.1.1 is not working neither
Neither unsetting MONGOID_AUTH_MECH or setting it to “:scram” works. Both give the same result.
The only setting that works is
```
ENV MONGOID_AUTH_MECH ""
```
in the forum dockerfile template
I suggest you revert this change also until we find the root cause and a final solution
I guess my local test passed because I was using an old working image, while I was using the new one in my k8s repo

regisb:
@angonz what is the stack trace of the error that you observe?

regisb:
I'll answer my own question: the stack trace is pretty much the same:

```
forum_1          | /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/user.rb:87:in `encoded_name': undefined method `encod
e' for nil:NilClass (NoMethodError)                                                                                                                           
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:320:in `first_bare'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:263:in `client_first_
message'                                                                                      
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram/conversation.rb:181:in `start'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/auth/scram.rb:58:in `login'       
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:249:in `block in authentica
te!'                                                                                                                                        
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server.rb:267:in `handle_auth_failure!'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:248:in `authenticate!'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:100:in `connect!'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connectable.rb:84:in `ensure_connected'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:270:in `write'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:229:in `deliver'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:144:in `block in dispatch'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/monitoring/publishable.rb:47:in `publish_command
'                                                                                                                                   
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection.rb:143:in `dispatch'
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/operation/shared/executable.rb:33:in `block in d
ispatch_message'                                                                                                 
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server/connection_pool.rb:110:in `with_connectio
n'                                                                                                                                                            
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/server.rb:251:in `with_connection'              
forum_1          |      from /openedx/cs_comments_service/vendor/bundle/ruby/2.5.0/gems/mongo-2.5.3/lib/mongo/operation/shared/executable.rb:32:in `dispatch_m
essage'
```

The problem is that we should only define MONGOID_AUTH_MECH when there is an actual authentication taking place. I'll provide a fix and release v12.1.2 right away.

insad:
Did you see the observation under "Forum" here https://openedx.atlassian.net/wiki/spaces/COMM/pages/940048716/Juniper ?

regisb:
Good find @insad! I just had a look at this page's history and... I was the one who wrote this :sweat_smile:

angonz:
Fisrt of all, thanks for your support with this!
I feel quite embarrassed for having Tutor down and giving you so much work... Anyway it's a big learning for me and I wish to thank you all for your help, patience and tolerance!

**This is what I've found:**
Seems that if `MONGOID_AUTH_MECH` is set to any non-empty string, the ruby gem that handles mongodb will try to authenticate the user. But as the `MONGODB_USER` is unset, the user object will be _nil_ and any attempt to call a method on the nil object will raise an exception.
Unsetting `MONGOID_AUTH_MECH` will have rake set it to ":scram" with the same effect.
Looks like only setting `MONGOID_AUTH_MECH` to an empty string will skip authentication and will work with Tutor's mongodb image.

I can see that the Tutor mongodb image does not include a `mongod.conf` in /etc, thus authentication is not enabled there. In fact, you can access any collection without even authenticating just running `mongo`. That is why it works in local and k8s deployments with the built in mongo image.

So we are back at the beginning. 

I will restate the problem in another way.

**Use cases:**
We have at least for use cases: default local deployment, default k8s deployment, local deployment with remote mongodb, k8s deployment with remote mongodb.

**The problem:**
Enabling MongoDB authentication crashes deployments and jobs.

**The root cause:**
MongoDB authentication is defined in `/etc/mongod.conf`. The default Tutor mongodb image does not include this file, so this mongod accepts connections without authentication. Enabling MongoDB authentication will make mongod refuse connections without proper authentication.
The jobs and deployments don't include a setting for `MONGOID_AUTH_MECH`, so it is set by default with an empty string, from the **forum** image. Therefore no authentication is attempted. MongoDB deployments with authentication enabled refuse any connection attempt without authentication.

**Approaches:**

1) My original approach was to check (in the jobs and deployments templates) if `MONGODB_USER` was set and in this case, setting `MONGOID_AUTH_MECH` to `:scram`. We could go again this path. I understand this can be a bit confusing and complex.

2) Remove authentication from mongod.conf. I tested it even in k8s with external mongo and it works. But can be a security flaw, specially in remote mongodb installations. 

3) Enable authentication in `mongod.conf` in all deployments, including Tutor's out of the box image. There may be a reason why this file is not there. But I believe this could be the easiest and most secure way.

What do you think?

regisb:
>  I feel quite embarrassed for having Tutor down and giving you so much work... Anyway it's a big learning for me and I wish to thank you all for your help, patience and tolerance!

It's ok Andres. People who don't do anything don't break anything.

> My original approach was to check (in the jobs and deployments templates) if MONGODB_USER was set and in this case, setting MONGOID_AUTH_MECH to :scram. We could go again this path. I understand this can be a bit confusing and complex.

It looks like this corresponds to the solution I proposed: https://github.com/overhangio/tutor/commit/bf5d2b80d8dc92ae8f5a88637c8e4e1524ee14b9
Can you confirm that it works for you? (with Mongodb authentication)

PR #494, opened 2021-09-17
Add tutorials to docs, and pave the way for the edge branch

regisb:
This PR introduces a few minor changes: the general idea is to pave the way for the introduction of the "edge" branch, which will be used to run Open edX from the master branch.

- We move many of the guides from the docs to a dedicated "tutorials" section. We take the opportunity to add a tutorial on how to run edge there.
- We get CI to work with edge.
- We add a "reference" section to the docs, which is automatically generated with [sphinx-click](https://sphinx-click.readthedocs.io/).
- We make many minor improvements to the docs.
- In passing, we make a breaking change to the `settheme` command.


docs/tutorials.rst
```diff
@@ -0,0 +1,24 @@
...
+---------------------
+
+.. toctree::
+   :maxdepth: 2
+
+   tutorials/portainer
+   tutorials/proxy
+   tutorials/datamigration
+   tutorials/multiplatforms
+   tutorials/oldreleases
```
BbrSofiane, line 24: Should "Running Tutor with Podman" also be in this section?

docs/tutorials.rst
```diff
@@ -0,0 +1,24 @@
+Tutorials
```
BbrSofiane, line 1: What would be the requirements for a topic to move from shared knowledge in the forum to tutorial? 

It could be valuable to define that here. That way it would make it easier for people to propose PRs for the tutorial section.

docs/tutorials.rst
regisb: Yes, you're right.

docs/tutorials.rst
regisb: For me the big difference is that tutorials should be sufficiently stabilized to be added to the documentation. Maybe the right workflow could be:

1. Write a tutorial in the forums (in a new subcategory)
2. Get feedback from the community and adjust the tutorial
3. Once the tutorial has been battle-tested by a few users, add it to the docs.

BbrSofiane: Sounds good to me

docs/tutorials.rst
regisb: I attempted to actually run Tutor with Podman and I was sorely disappointed. The only reliable source of docs that I found concerning the integration with docker-compose is this blog post: https://www.redhat.com/sysadmin/podman-docker-compose
There are no other official docs :sweat: 

1. The instructions given in the blog post don't work out of the box. Launching the podman service failed altogether on Ubuntu 20.04 and 20.10. It worked on CentOS 8, but some parameters need to changed, such as the docker socket path.
2. After I got the podman service working, I managed to get an Open edX platform running with tutor, but with the root user. Then, containers complained that they could not write data to the bind-mounted volumes. I attempted to run as a non-root user, and discovered that the podman socket is only readable by root. This should explain why all commands from that blog post are prefixed by `sudo`.

Long story short, I was hoping to update the tutorial. Instead, I'll just move it for the sake of better organisation. For the life of me, I do not understand why some people would want to run Podman instead of Docker. Bad documentation is an immediate turn-off for me.

regisb:
This is ready for a new round of reviews. Note that the "edge" branch was renamed to "nightly".

ghassanmas:
Regarding the common usage of the subsution where  `$(tutor config printroot)/something/.etc`. I think it would fit better if it's used like `"$(tutor config printroot)"/something/..etc` because if for example there is **space** somewhere in the path of the printroot, it would fail unless wrapped in qoutation. 

Note: I tested that on both Mac and Linux based envs, but not sure how that would affect windows env.

PR #498, opened 2021-09-27
docs: clarify how to use custom ssl certificates

regisb:
I realized that the docs were very unclear, contradictory and misleading on
the topic of SSL/TLS termination and web proxies.

See: https://discuss.overhang.io/t/why-caddy-nginx/1952/10


docs/configuration.rst
```diff
@@ -183,7 +183,7 @@ SSL/TLS certificates for HTTPS access
 
 - ``ENABLE_HTTPS`` (default: ``false``)
 
-By activating this feature, a free SSL/TLS certificate from the `Let's Encrypt <https://letsencrypt.org/>`_ certificate authority will be created for your platform. With this feature, **your platform will no longer be accessible in HTTP**. Calls to http urls will be redirected to https url.
+When ``ENABLE_HTTPS`` is ``true``, the whole Open edX platform will be reconfigured to work with "https" URIs. Calls to "http" URIs will be redirected to "https". By default, SSL/TLS certificates will automatically be generated by Tutor (thanks to `Caddy <https://caddyserver.com/>`__) from the `Let's Encrypt <https://letsencrypt.org/>`_ certificate authority. With this feature, **your platform will no longer be accessible in HTTP**.
```
iamCristYe, line 186: rather than saying `your platform will no longer be accessible in HTTP`, I think `Calls to http urls will be redirected to https url` is better. the former makes me feel http://www.openedx.com would return a 403/500 error.

regisb: Good point. This sentence is already present in the first part of the paragraph, so I guess I can just remove the bold warning.

PR #502, opened 2021-09-29
Feature: new configuration setting PREVIEW_LMS_BASE

iamCristYe:
discussion here: https://discuss.overhang.io/t/new-settings-variable/1973/4

If my interpretation of https://github.com/overhangio/tutor/blob/926263deb31fab0fe28c1046fd6649e8be31fecd/tutor/config.py#L47 is correct, current users will not need to manually add the new variable for preview domain to work, as missing this config yaml will lead tutor to use the default value.


tutor/interactive.py
```diff
@@ -41,6 +41,7 @@ def ask_questions(config: Config, defaults: Config) -> None:
         dev_values: Config = {
             "LMS_HOST": "local.overhang.io",
             "CMS_HOST": "studio.local.overhang.io",
+            "PREVIEW_LMS_HOST": "preview.local.overhang.io",
```
regisb, line 44: I would rather avoid adding an extra question to the quickstart. I think that the default value is reasonable for most people, and we don't need to ask everyone about it.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Improvement] Add variable `PREVIEW_LMS_BASE` for custom preview domain.
```
regisb, line 7: Please replace "variable" by "configuration setting".

regisb (changes requested):

CHANGELOG.md
regisb: Also, this is actually a new feature, so you can replace "Improvement" by "Feature". Please also amend the git commit title accordingly.

regisb (changes requested):

regisb:
Can you please add the discussion link to the git commit message, such that people don't have to research this PR later?

iamCristYe:
@regisb Please check if it's good for merge now 😊

tutor/interactive.py
```diff
@@ -146,16 +155,6 @@ def ask_questions(config: Config, defaults: Config) -> None:
...
-    if run_for_prod:
-        ask_bool(
-            (
-                "Activate SSL/TLS certificates for HTTPS access? Important note:"
-                " this will NOT work in a development environment."
-            ),
-            "ENABLE_HTTPS",
-            config,
-            defaults,
-        )
```
regisb, line 158: This part of the file should not be modified.

regisb (changes requested):

regisb:
Thanks again for your PR @iamCristYe!

PR #505, opened 2021-10-04
feat: run all services as unprivileged containers

### LINKED ISSUE #323: Creating ‘openedx’ user takes too long when building openedx-dev image
As per the documentation, I ran the following command to build the openedx-dev image

`tutor images build openedx-dev`

OS Version is: 18.04.4 LTS
Tutor Version is: 3.12.3
I left it system to run overnight(almost 12 hours), yet the command was not successful. It was stuck at running the create-user.sh. Below is the log:
(I have used a workaround as specified [here](https://discuss.overhang.io/t/development-build-stuck-at-create-user/541/3), but this is a stop gap arrangement)
```
Building image docker.io/overhangio/openedx-dev:3.12.3
docker build -t docker.io/overhangio/openedx-dev:3.12.3 /home/machine/.local/share/tutor/env/build/openedx-dev --build-arg USERID=1000
Sending build context to Docker daemon  5.632kB
Step 1/11 : FROM docker.io/overhangio/openedx:3.12.3
 ---> 4e83d402b431
Step 2/11 : MAINTAINER Overhang.io <contact@overhang.io>
 ---> Using cache
 ---> ab86cbe8487e
Step 3/11 : RUN apt update &&     apt install -y vim telnet     && rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> 6cdbaa38645e
Step 4/11 : RUN pip install -r requirements/edx/development.txt
 ---> Using cache
 ---> 39e71c2bff12
Step 5/11 : RUN pip install ipdb==0.12.2 ipython==5.8.0
 ---> Using cache
 ---> 014cec4bd517
Step 6/11 : RUN rm -r /openedx/staticfiles &&     mkdir /openedx/staticfiles &&     openedx-assets webpack --env=dev
 ---> Using cache
 ---> 752e39197dfd
Step 7/11 : COPY ./bin /openedx/bin
 ---> Using cache
 ---> ccb0d4bf5d2a
Step 8/11 : RUN chmod a+x /openedx/bin/*
 ---> Using cache
 ---> 8fd152d4cbca
Step 9/11 : ARG USERID=1000
 ---> Using cache
 ---> dc04a982062b
Step 10/11 : RUN create-user.sh $USERID


 ---> Running in e6215f4a433c
Creating 'openedx' user with id 1000

```

regisb:
With this change, containers are no longer run as "root" but as unprivileged
users. This is necessary in some environments, notably some Kubernetes
clusters.

To make this possible, we need to manually fix bind-mounted volumes in
docker-compose. This is pretty much equivalent to the behaviour in Kubernetes,
where permissions are fixed at runtime if the volume owner is incorrect. Thus,
we have a consistent behaviour between docker-compose and Kubernetes.

We achieve this by bind-mounting some repos inside "*-permissions" services.
These services run as root user on docker-compose and will fix the required
permissions, as per build/permissions/setowner.sh These services simply do not
run on Kubernetes, where we don't rely on bind-mounted volumes. There, we make
use of Kubernete's built-in volume ownership feature.

With this change, we get rid of the "openedx-dev" Docker image, in the sense
that it no longer has its own Dockerfile. Instead, the dev image is now simply
a different target in the multi-layer openedx Docker image. This makes it much
faster to build the openedx-dev image.

Because we declare the APP_USER_ID in the dev/docker-compose.yml file, we need
to pass the user ID from the host there. The only way to achieve that is with a
tutor config variable. The downside of this approach is that the
dev/docker-compose.yml file is no longer portable from one machine to the next.
We consider that this is not such a big issue, as it affects the development
environment only.

We take this opportunity to replace the base image of the "forum" image. There
is now no need to re-install ruby inside the image. The total image size is
only decreased by 10%, but re-building the image is faster.

In order to run the smtp service as non-root, we switch from namshi/smtp to
devture/exim-relay. This change should be backward-compatible.

Note that the nginx container remains privileged. We could switch to
nginxinc/nginx-unprivileged, but it's probably not worth the effort, as we are
considering to get rid of the nginx container altogether.

Close #323.


regisb:
It should be noted that this PR was heavily influenced by this other one, by @menardorama https://github.com/overhangio/tutor/pull/364/

regisb:
Hey @thomas-skillup! I'm not sure I understand all of your comments. If we need to discuss this further I suggest we schedule a call. In particular, I am not familiar at all with `setfacl`, so I would appreciate more in-depth explanations.

Let me just emphasize that we need to address the following scenarios:

1. We must be able to run the "lms" and "cms" services in unprivileged mode, with `user != 0`, both locally and on Kubernetes. Thus, `gosu` and `sudo` are out of the question.
2. In development mode, the "lms" and "cms" containers must have write access to the shared written folders (/openedx/media and /openedx/data).
3. In development, when bind-mounting edx-platform, the repo file permissions must not be modified.

regisb:
For the record, this PR was modified to merge the changes in the "nightly" branch: that's because we don't want to merge this breaking change in the current release branch.

PR #508, opened 2021-10-14
feat: get rid of the nginx container and services

regisb:
Nginx and Caddy performed duplicate tasks. It was decided to get rid of
the nginx container, for simplification. This is a breaking change for
plugin developers. Also, applications that collect nginx logs will have
to be modified.

See: https://discuss.overhang.io/t/why-caddy-nginx/1952


regisb:
The meaning of the ``ENABLE_WEB_PROXY`` variable is whether or not a web proxy should be enabled _within tutor_. When true, Caddy performs as a web proxy. When false, something else must fill this role.

Does this make more sense?

iamCristYe:
Hmm... not really?

As in [docker-compose.prod.yml](https://github.com/overhangio/tutor/blob/580b85db091c69de9ca1f2ccbc31205b4c072364/tutor/templates/local/docker-compose.prod.yml), Caddy will always be there whether `ENABLE_WEB_PROXY` is true or not. Though I noted that in k8s config file, when `ENABLE_WEB_PROXY` is false, Caddy is indeed not running. 

I'm not so familiar with k8s, but at least in the docker world, it would be hard to redirect outside traffic to lms/cms container if they're not directly exposing a port, without the Caddy bridge.

Thus, I think the `WEB_PROXY` in `ENABLE_WEB_PROXY` shouldn't be Caddy, but the one in front of it.

What do you think?

regisb:
> As in docker-compose.prod.yml, Caddy will always be there whether ENABLE_WEB_PROXY is true or not. Though I noted that in k8s config file, when ENABLE_WEB_PROXY is false, Caddy is indeed not running.

Yes, Caddy will be running, but it will not expose port 443. Which means that it expects that another process will handle ssl termination.

> Though I noted that in k8s config file, when ENABLE_WEB_PROXY is false, Caddy is indeed not running.

This is a bug! I'll open a PR to fix this: https://github.com/overhangio/tutor/pull/516

> Besides, I really don't think user should be responsible to replicate a complex setting in https://github.com/overhangio/tutor/blob/580b85db091c69de9ca1f2ccbc31205b4c072364/tutor/templates/apps/caddy/Caddyfile if they just want to use their own cert in k8s.

There are many different ways to run a web proxy (different than Caddy) with custom certs in Kubernetes. I don't want to provide (and support) anything else than guidelines to

iamCristYe:
> > Besides, I really don't think user should be responsible to replicate a complex setting in https://github.com/overhangio/tutor/blob/580b85db091c69de9ca1f2ccbc31205b4c072364/tutor/templates/apps/caddy/Caddyfile if they just want to use their own cert in k8s.
> 
> There are many different ways to run a web proxy (different than Caddy) with custom certs in Kubernetes. I don't want to provide (and support) anything else than guidelines to

I said that when I don't know #516 is a bug. Now that Caddy will always be running when using either docker-compose or k8s, I feel even more confused: whether `ENABLE_WEB_PROXY` is true or not, the `WEB_PROXY`(Caddy) is always  `ENABLED`. I can understand the logic behind it though, I just hope other users wouldn't be confused 😄

PR #512, opened 2021-10-26
fix: bypass build to accelerate "local start"

regisb:
Previously, we were building all images every time we ran a "local start"
command. This was causing unnecessary rebuild. Here, instead, we make use of
the `docker-compose up --build`. This means that only the required images will
be rebuilt.


BbrSofiane (approved):
LGTM

PR #516, opened 2021-11-01
fix: always run Caddy on Kubernetes

regisb:
Caddy should always be running, even when ENABLE_WEB_PROXY is false.
It's the service that should not always be running.


regisb:

BbrSofiane (approved):
LGTM

PR #521, opened 2021-11-05
Update portainer.rst

ncresswell:
Update to latest ver of Portainer, that also now runs https by default.

### DIFF
--- a/docs/tutorials/portainer.rst
+++ b/docs/tutorials/portainer.rst
@@ -8,10 +8,10 @@ Portainer is a web UI for managing docker containers. It lets you view your enti
     docker run --rm \
         --volume=/var/run/docker.sock:/var/run/docker.sock \
         --volume=/tmp/portainer:/data \
-        -p 9000:9000 \
-        portainer/portainer:latest --bind=:9000
+        -p 9000:9000 -p 9443:9443 \
+        portainer/portainer-ce:latest
 
-You can then view the portainer UI at `http://localhost:9000 <http://localhost:9000>`_. You will be asked to define a password for the admin user. Then, select a "Local environment" to work on; hit "Connect" and select the "local" group to view all running containers.
+You can then view the portainer UI at `https://localhost:9443 <https://localhost:9443>`_. You will be asked to define a password for the admin user. Then, select the "local" group to view all running containers.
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/portainer.rst
```diff
@@ -8,10 +8,10 @@ Portainer is a web UI for managing docker containers. It lets you view your enti
     docker run --rm \
         --volume=/var/run/docker.sock:/var/run/docker.sock \
         --volume=/tmp/portainer:/data \
-        -p 9000:9000 \
-        portainer/portainer:latest --bind=:9000
+        -p 9000:9000 -p 9443:9443 \
+        portainer/portainer-ce:latest
 
-You can then view the portainer UI at `http://localhost:9000 <http://localhost:9000>`_. You will be asked to define a password for the admin user. Then, select a "Local environment" to work on; hit "Connect" and select the "local" group to view all running containers.
+You can then view the portainer UI at `https://localhost:9443 <https://localhost:9443>`_. You will be asked to define a password for the admin user. Then, select the "local" group to view all running containers.
```
regisb, line 14: On my computer, accessing https://localhost:9443 triggers a security alert: "The certificate is not trusted because it is self-signed." Which makes sense. Thus, I suggest that we do not recommend users to access Portainer in https. Also, we should not forward port 9443.

Actually, I also realized that with our recommended command, port 9000 would be exposed to the world. We should instead write `-p 127.0.0.1:9000:9000`.

regisb (changes requested):
Thanks for your PR! I was not aware that the base portainer image had changed.

docs/tutorials/portainer.rst
ncresswell: Self-signed SSL is far far safer than raw http, so i still recommend that people use https over http

regisb: I must admit that I'm not great at security; could you please explain what scenarios we would be avoiding if users were redirected to an https server with self-signed certificates on localhost? In my understanding, any attacker could spoof that domain with also self-signed certificates, right?

regisb:
I will update the docs to use portainer/portainer-ce instead of portainer/portainer. Feel free to re-open this PR if you'd like to add more changes.

PR #524, opened 2021-11-16
fix: remove trailing slashes in volume paths for docker-compose v2 compatibility

BbrSofiane:
Remove trailing slashes in volume paths for docker-compose v2 compatibility.

@overhangio/tutor-developers this is ready for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
+- [Bugfix] Remove trailing slashes in docker-compose files for [compatibility with docker-compose v2 in WSL](https://github.com/docker/compose/issues/8558).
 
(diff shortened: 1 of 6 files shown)


regisb (changes requested):
Looking good! Can you please add a line to the changelog to explain that this resolves a WSL issue? Also please add "close #522" to the commit message to automatically close the issue on merge.

BbrSofiane:
@regisb whoops yep I forgot the changelog!

iamCristYe:
better update https://github.com/overhangio/tutor/blob/master/docs/dev.rst#override-docker-compose-volumes as well?

regisb:
Yes @BbrSofiane I think that @iamCristYe is right, we should update the docs as well.

BbrSofiane:
Good spot @iamCristYe

regisb (approved):

Issue #450, opened 2021-06-14
The Open edX forum should be an optional feature


regisb:
**Is your feature request related to a problem? Please describe.**
Not everyone wants to run the Open edX forums on the LMS. Thus, forums should be moved to an officially-supported "forum" plugin.

**Describe the solution you'd like**

- Disable the forums by default in tutor-openedx
- Create an official forum plugin hosted on github.com/overhangio
- Make sure that the forum plugin is bundled with the `tutor` binary generated by GitHub Actions

**Additional context**
This issue was first reported here: https://discuss.overhang.io/t/disabling-the-forum-in-tutor/1599

ghassanmas:
I am intrested in reolsving this:
To make sure I am in the right direction here is my inteded to be workflow: 
For tutor 
- Remove everything relaeted the fourm (service and image)  from the standard tutor templates, congfiguration settings..etc
- Create a plugin in a separate repo, simliar to overhangio/tutor-mfe  
- When **enabled**, tutor shuold run as how it's running now by defualt, creating service/job, building fourm if in dev..etc

regisb:
Thanks for offering your help @ghassanmas! Your description of the work to be done is fairly accurate. It's certainly _not_ what I would consider an "easy" issue to tackle, especially if you are new to Tutor, but it's a great introduction to the Tutor internals and the plugin system. Also, I can provide advice as you make progress.

We will need to create the plugin before removing it from the Tutor core. I think it would be best to make the removal effective in Maple, and not before.

ghassanmas:
@regisb, following throught I have started with creating a draft for the plugin, with only the patches being pushed in the meantime, let's move the conversation there I guess, overhangio/tutor-forum/pull/1

regisb:
Closed by https://github.com/overhangio/tutor/pull/513

PR #526, opened 2021-11-17
Include the forum job only if RUN_FORUM is set

### LINKED ISSUE #525: Kubernetes jobs.yml template ignores RUN_FORUM
**Bug description**
When running `tutor k8s` we can disable the forum deployment and service by setting `RUN_FORUM` to `false`, but the corresponding job is added to `jobs.yml` unconditionally.

**How to reproduce**
* In a fresh Tutor environment, set `RUN_FORUM: False` in `tutor/config.yml`
* Run `tutor k8s quickstart`
* Observe that there is no mention of forum in `$TUTOR_ROOT/env/k8s/deployments.yml` or `services.yml`, but there is a job named `forum-job` in `jobs.yml`.

**Environment**
* Tutor 12.1.6

(The fix for this should be trivial; PR incoming.)

fghaas:
When running `tutor k8s` we can disable the forum deployment and service by setting `RUN_FORUM` to `false`, but the corresponding job is added to `jobs.yml` unconditionally.

Add a conditional to `tutor/templates/k8s/jobs.yml` so that the forum job definition is only included when `RUN_FORUM` is `true`.

Fixes #525.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Note: Breaking changes between versions are indicated by "💥".
 - [Feature] Add [tutor-richie](https://github.com/overhangio/tutor-richie) to the plugins that are bundled with the tutor binary.
 - [Improvement] Make `tutor plugins list` print plugins sorted by name.
 - [Improvement] Ignore Python plugins which cannot be loaded.
+- [Bugfix] When configured with `RUN_FORUM: false`, omit forum-related [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) from the manifests that `tutor k8s` generates. (#525)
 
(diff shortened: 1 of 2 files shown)


regisb (changes requested):
Good catch! This will be unnecessary soon, when [the forum is pulled out of tutor core](https://github.com/overhangio/tutor/issues/450), but in the meantime we need this fix.

Can you just add an entry to the changelog?

fghaas:
> Can you just add an entry to the changelog?
Sure, done.

regisb (approved):

regisb:
Thanks for the PR @fghaas!

fghaas:
Thanks for merging!

PR #529, opened 2021-11-18
Improvement: `settheme` now works with preview domain and `NGINX_HTTP_PORT`

iamCristYe:
https://discuss.overhang.io/t/settheme-not-applied-to-preview-site/2144

### DIFF
--- a/tutor/jobs.py
+++ b/tutor/jobs.py
@@ -147,4 +147,6 @@ def get_all_openedx_domains(config: Config) -> List[str]:
         get_typed(config, "LMS_HOST", str) + ":8000",
         get_typed(config, "CMS_HOST", str),
         get_typed(config, "CMS_HOST", str) + ":8001",
+        get_typed(config, "PREVIEW_LMS_HOST", str),
+        get_typed(config, "PREVIEW_LMS_HOST", str) + ":8000",
     ]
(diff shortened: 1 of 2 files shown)


tutor/jobs.py
```diff
@@ -144,7 +144,17 @@ def set_theme(theme_name: str, domain_names: List[str], runner: BaseJobRunner) -
 def get_all_openedx_domains(config: Config) -> List[str]:
     return [
         get_typed(config, "LMS_HOST", str),
+        get_typed(config, "LMS_HOST", str)
+        + ":"
+        + get_typed(config, "NGINX_HTTP_PORT", str),
```
regisb, line 149: I don't understand. Why do we need to append the nginx port here? When running a separate web proxy, we are not supposed to access the lms without passing through this proxy first.

regisb (changes requested):

tutor/jobs.py
iamCristYe: For the case like https://discuss.overhang.io/t/install-themes-not-working-even-indigo/2140 ? There's no harm adding this line I think?

regisb: If `NGINX_HTTP_PORT` is different than 80 (say: 81), then the above config will only work for http://lmshost:81. Yet, users are really not supposed to open the LMS from this URL. For instance, this host is not part of the Django ALLOWED_HOSTS.

So I think we must drop the NGINX_HTTP_PORT part here.

regisb (approved):

PR #535, opened 2021-11-25
feat: fail `images build/pull/push` commands on incorrect image arg

regisb:
Previously, `tutor images build doesnotexist` was a no-op. This was
counter-intuitive -- in particular, when users have a typo in their command.

Now, the command fails if the corresponding image was not found.

See discussion: https://discuss.overhang.io/t/return-something-when-tutor-images-build-contains-a-typo/2174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
+
+- 💥[Improvement] Fail on incorrect image name argument in `images build/pull/push/printtag` commands.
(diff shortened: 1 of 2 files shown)


regisb:
cc @iamCristYe who started the initial conversation.

iamCristYe:
There's a command `printtag` as well, would you consider do the same for that?

regisb:
Good catch @iamCristYe. I edited the PR.

regisb:

BbrSofiane (approved):
Looks good to me!

PR #536, opened 2021-11-25
fix: insecure http authentication in LMS

regisb:
When HTTPS was not enabled, it was not possible to login to the LMS in some browsers, such as Brave.

- The sessionid cookie was being set with SameSite=None.
- That's because we were setting the DCS_SESSION_COOKIE_SAMESITE but not the
  SESSION_COOKIE_SAMESITE setting.
- The DCS_* settings are ignored since edx-platform no longer makes use of
  django-cookies-samesite.

Close https://github.com/openedx/build-test-release-wg/issues/110

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,6 +2,7 @@
 
 Note: Breaking changes between versions are indicated by "💥".
 
+- [Improvement] Get rid of many Django deprecation warnings that were printed in stdout for every LMS/CMS command.
 - [Bugfix] Fix authentication to LMS when HTTPS is disabled.
 - [Bugfix] Make it possible for plugins to implement the "caddyfile" patch without relying on the "port" local variable.
 - 💥[Improvement] Move the Open edX forum to a [dedicated plugin](https://github.com/overhangio/tutor-forum/) (#450).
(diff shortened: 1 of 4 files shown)


regisb:
cc @BbrSofiane and @ghassanmas who mentioned that this bug was affecting you. Can you please test the current branch? If it works for you we will merge in the nightly branch.

ghassanmas:
@regisb Yes, it worked with the following configuration (I was able to log in).

```
CADDY_HTTP_PORT: 80
ENABLE_HTTPS: false
ENABLE_WEB_PROXY: true
```

But there was a buggy behaviour, (I speculate it's because of the browser because I was accessing the platform using HTTPS, and then using the new configuration I need to access with HTTP). However, trying to access the platform in **incognito** mode was a success. Also, I was able to access it in normal mode after I _**cleared my browser data**_

regisb:
@ghassanmas if you access your platform with https, **you must set `ENABLE_HTTPS=true`**. See this section of the docs: https://docs.tutor.overhang.io/configuration.html#ssl-tls-certificates-for-https-access

ghassanmas:
Yes, I am aware of that, I was deliberately accessing it with HTTP just to test this changes and it **worked**.
You can ignore my note/comment above, it was just cahce/browser issue.

BbrSofiane (approved):
It works for me too!

PR #537, opened 2021-11-26
refactor: clarify configuration management

regisb:
Previously, configuration management was very confusing because we kept mixing
"base" and "defaults" configuration:

- It was difficult to make the difference between core settings that were
  necessary (e.g: passwords) as opposed to others that could simply be
  defaulted to.
- The order of settings in config.yml mattered: config entries that depended on
  other needed to be defined later. As a consequence, Tutor was not compatible
  with Python 3.5, where dict entries are not sorted.

This is a major refactor but it should not (:crossed_fingers:) have backward incompatible changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ test-format: ## Run code formatting tests
 	black --check --diff $(BLACK_OPTS)
 
 test-lint: ## Run code linting tests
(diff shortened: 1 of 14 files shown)


regisb:
I will merge this so that everyone has a chance to test the refactored code prior to the maple release.

PR #539, opened 2021-11-29
Multiple improvements to the Caddy load balancer in Kubernetes

### LINKED ISSUE #532: Kubernetes LoadBalancer race conditions and other issues
Initial conversation: https://discuss.overhang.io/t/tutor-k8s-in-production-mode-is-this-a-dns-catch-22/2119/

In this conversations, several issues were raised:

1. A Kubernetes deployment requires a valid host name to run properly. Yet, we want to keep supporting minikube and local k8s cluster environments, as well as non-https deployments.
2. When https is enabled, a valid Caddy service must be ready before the platform is initialised. Otherwise, initialisation will fail. This is a sort of race condition in setting up the deployments and services.
3. When the platform is stopped, with `tutor k8s stop`, the LoadBalancer is deleted. As a consequence, when it is started again, its external IP will change, and DNS records will have to be updated.

We propose the following fixes:

- [ ] Somehow, find a way to run in non-production mode on a local cluster. If not, we should not be asking the user if they want to run in non-production mode during `tutor save -i`.
- [x] Provide an easy way to create a LoadBalancer object, such as `tutor k8s start caddy`. This will leave time to the end user to create the right DNS records.
- [x] Ensure that the Caddy deployment is up prior to platform initialisation. 
- [x] Do not delete the LoadBalancer on `tutor local stop`.

K8s deployment will It's ok if the k8s deployment is not exactly a 1-click installation. [As mentioned in the docs](https://docs.tutor.overhang.io/k8s.html), k8s deployments is aimed at operators who know what they are doing. There are so many k8s environments out there that we should not try to provide a one-size-fits-all deployment solution.

Note that at this point, I do not want Tutor to actively check that the DNS records are correct. This feature could be performed by a plugin, for example during initialisation.

cc @fghaas who initially started the conversation around these issues.

regisb:
See the changelog for a detailed review of changes.

Close #532, except for this item:

>  Somehow, find a way to run in non-production mode on a local cluster. If not, we should not be asking the user if they want to run in non-production mode during tutor save -i.

I think that we should stop asking the user whether they want to run in production when they run `tutor k8s quickstart`, but I'd like to merge #537 before we make this change. I'll open another PR later.

cc @fghaas who started the conversation around these issues.

### DIFF
--- a/docs/k8s.rst
+++ b/docs/k8s.rst
@@ -27,10 +27,22 @@ The Kubernetes cluster should have at least 4Gb of RAM on each node. When runnin
 .. image:: img/virtualbox-minikube-system.png
     :alt: Virtualbox memory settings for Minikube
 
-Ingress controller and SSL/TLS certificates
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Load Balancer and SSL/TLS certificates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-As of Tutor v11, it is no longer required to setup an Ingress controller to access your platform. Instead Caddy exposes a LoadBalancer service and SSL/TLS certificates are transparently generated at runtime.
+By default, Tutor deploys a `LoadBalancer <https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer>`__ service that exposes the Caddy deployment to the outside world. As in the local installation, this service is responsible for transparently generating SSL/TLS certificates at runtime. You will need to point your DNS records to this LoadBalancer object before the platform can work correctly. Thus, you should first start the Caddy load balancer, with::
+
+    tutor k8s start caddy
+
+Get the external IP of this services::
+
+    kubectl --namespace openedx get services/caddy
+
+Use this external IP to configure your DNS records. Once the DNS records are configured, you should verify that the Caddy container has properly generated the SSL/TLS certificates by checking the container logs::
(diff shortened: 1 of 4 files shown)


docs/k8s.rst
```diff
@@ -27,10 +27,22 @@ The Kubernetes cluster should have at least 4Gb of RAM on each node. When runnin
...
+
+Get the external IP of this services::
+
+    kubectl --namespace openedx get services/caddy
+
+Use this external IP to configure your DNS records. Once the DNS records are configured, you should verify that the Caddy container has properly generated the SSL/TLS certificates by checking the container logs::
+
+    tutor k8s logs -f caddy
+
+If, for some reason, you would like to deploy your own load balancer, you should set ``ENABLE_WEB_PROXY=false`` just like in the :ref:`local installation <web_proxy>`. Then, point your load balancer at the "caddy" service, which will be a `ClusterIP <https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types>`__.
```
fghaas, line 45: I'm not sure if I'm missing something but this looks like this patch is intended to be against `master`, and as far as I can tell `ENABLE_WEB_PROXY` currently exists only in `nightly`. So if this lands as-is in `master`, I think this would be a case of documentation contradicting code?

regisb: Yes, you are right. I attempted to land this in the master branch, but it appears that it will not be possible. I'll rebase my changes on top of the nightly branch.

PR #543, opened 2021-12-02
Fix arm64 image build - open-release/lilac.3

bradenmacdonald:
When I tried building the `openedx` image on an ARM64 Mac, this line of the `Dockerfile` failed:

https://github.com/overhangio/tutor/blob/19157d94bca8a4998271082045448a919bcd4761/tutor/templates/build/openedx/Dockerfile#L81-L82

[The error message was that `libgeos` could not be found while installing `shapely`.](https://stackoverflow.com/questions/19742406/could-not-find-library-geos-c-or-load-any-of-its-variants)

I am not sure why this would differ between arm64 and amd64 builds, but in any case this PR fixes the issue.

With this fix, `tutor images build openedx` works and is sufficient to build the LMS image for the `arm64/v8` architecture without any other changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Note: Breaking changes between versions are indicated by "💥".
 
+## Unreleased
+
+- [Bugfix] Fix building of the `openedx` image on ARM64 due to missing `libgeos-dev`
+
 ## v12.2.0 (2021-12-08)
 
 - [Bugfix] Fix incorrect "from" address in course bulk emails (see [pull request](https://github.com/edx/edx-platform/pull/29001)).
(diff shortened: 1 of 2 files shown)


regisb (approved):
Thanks for the fix. Can you please rebase your changes? Do you want to add a tutorial on top of this PR, or open another one? We can wait until you try out building images for the nightly branch.

bradenmacdonald:
Quick update: I got time to test nightly and it is _almost_ working (with this fix) but ran into some minor issue on one of the last steps. I'll post another update once I've figured it out.

Then once I've figured that out, I will suggest we merge this PR and I'll open a second PR with the full tutorial.

bradenmacdonald:
@regisb Sorry for the delay here. I've rebased this and it should be ready to merge whenever you'd like.

I will follow up with separate PRs for:

1. Fixing the nightly build (lilac+maple are fine with this fix, though I haven't tested MFEs yet. But `master` now requires mongodb to be running while a particular migration is applied, so it isn't working.)
2. Writing up a proper tutorial for the docs.

PR #548, opened 2021-12-15
Bugfix: A migration on nightly/master Open edX now requires MongoDB

bradenmacdonald:
In the latest master version of Open edX (post-Maple), some course data has moved from MongoDB to MySQL. There is [a data migration which automatically moves this data over](https://github.com/edx/edx-platform/blob/master/common/djangoapps/split_modulestore_django/migrations/0002_data_migration.py), but it currently causes Tutor nightly `tutor local init` to fail when running migrations, because MongoDB is not running.

Without this fix, `tutor local init` fails on Tutor nightly.

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,6 +2,7 @@
 
 Note: Breaking changes between versions are indicated by "💥".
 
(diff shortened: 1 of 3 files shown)


regisb (approved):
Nice, thanks! Looking forward to the day when we can get rid of mongodb entirely.

PR #554, opened 2022-01-03
fix: upgrade from Lilac on k8s

### LINKED ISSUE #551: 13.0.2 Upgrading a Kubernetes-managed installation from Tutor to Maple breaks Services
**Bug description**
Upgrading a Kubernetes-managed installation from Tutor to Maple breaks, because Kubernetes doesn't like the `nodePort` update to Services (the change that landed in 7c1e85ef4ba94cce1d597a1a3ea69cedbc2bde49).

**How to reproduce**
* Install Tutor 12.1.7, run `tutor k8s quickstart`
* Install Tutor 13.0.2, run `tutor k8s quickstart` again
* Observe that all Services break with:
   ```
   Service "mysql" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
   Service "redis" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
   ```

**Environment**
This is on Ubuntu Focal, against Kubernetes 1.18.20.

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T18:03:20Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.20", GitCommit:"1f3e19b7beb1cc0110255668c4238ed63dadb7ad", GitTreeState:"clean", BuildDate:"2021-06-16T12:51:17Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
<https://docs.tutor.overhang.io/k8s.html> doesn't cover Open edX release upgrades specifically, but <https://docs.tutor.overhang.io/install.html#upgrading> mentions re-running `tutor local quickstart`, so I am assuming that for Kubernetes, the same approach should work (with `tutor k8s quickstart`).

I understand that I can get around this with running `tutor k8s stop` while still on version 12, followed by `tutor k8s quickstart` on version 13 (because the persistent volumes persist, no data is lost this way), and people who ran Lilac on Tutor may want to do that when running the release upgrade. But the problem with that is that d8d0560b9e647d36bf3d596a559d41a6e01c06f6 only landed in version 13, meaning that doing so removes the Caddy service (and its external IP), so when people do that they will have to update their DNS records for the new IP external IP address they will receive. Meaning, the service interruption they incur may be longer than they expect.

Is there a way to add a follow-up change to 7c1e85ef4ba94cce1d597a1a3ea69cedbc2bde49, so that the Service change behaves nicely on a major release upgrade?

regisb:
When upgrading from Lilac, all services break with the following error:

    Service "***" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'

Upgrading deployments fails as well:

    Deployment.apps "***" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"openedx-********", "app.kubernetes.io/managed-by":"tutor", "app.kubernetes.io/name":"***", "app.kubernetes.io/part-of":"openedx"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

That's because deployments and services need to be deleted as part of the Maple
upgrade. So that's what we do as part of `tutor k8s upgrade --from=lilac`. And
we take the opportunity to:

1. Run upgrade as part of quickstart, when necessary.
2. Default to lilac during `tutor k8s upgrade`.

Close #551.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Fix ownership of mysql data, in particular when upgrading a Kubernetes cluster to Maple.
 - [Bugfix] Ensure that ``tutor k8s upgrade`` is run during ``tutor k8s quickstart``, when necessary.
 - [Bugfix] By default, upgrade from Lilac and not Koa during ``tutor k8s upgrade``.
 - [Bugfix] Fix upgrading from Lilac to Maple on Kubernetes by deleting deployments and services.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] During upgrade, make sure that environment is up-to-date prior to prompting to rebuild the custom images.
 - [Bugfix] Fix ownership of mysql data, in particular when upgrading a Kubernetes cluster to Maple.
 - [Bugfix] Ensure that ``tutor k8s upgrade`` is run during ``tutor k8s quickstart``, when necessary.
 - [Bugfix] By default, upgrade from Lilac and not Koa during ``tutor k8s upgrade``.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Ensure that ``tutor k8s upgrade`` is run during ``tutor k8s quickstart``, when necessary.
+- [Bugfix] By default, upgrade from Lilac and not Koa during ``tutor k8s upgrade``.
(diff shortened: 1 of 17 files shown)


regisb:

cc @fghaas

fghaas (commented):
I'm just back from the holiday break and will be catching up for a little while so I'm not sure when I'll be able to test this. But the approach looks entirely sound to me, thank you!

regisb:
I know that this PR was not reviewed. See my comment here: https://discuss.overhang.io/t/confusing-instructions-during-upgrade/2281/9 Sorry about that @overhangio/tutor-developers :-/

PR #567, opened 2022-01-14
draft: feat: move plugin CLIs under dev/local/k8s; let plugins run jobs

kdmccormick:
## Description

This is a potentially-overzealous yet backwards-compatible change to the Plugin API and Tutor Plugin CLI. Check out the commit message for the full description.

### Why?

In the tutor-coursegraph plugin I'm developing, I want to be able to [run a job against a service container from the plugin's custom CLI](https://github.com/kdmccormick/tutor-coursegraph/blob/6c327652c58b96a7ead075f91969ee5ebd167033/tutorcoursegraph/plugin.py#L64). As far as I could tell, there was no way to do that without this change.

If there's a way to do what I'm trying to achieve without this tutor PR, do enlighten me!

Related tutor-coursegraph issue: https://github.com/kdmccormick/tutor-coursegraph/issues/5

### Status

I'm not trying to get this reviewed and merged imminently; I'd like to go over the implementation with @regisb first.

Among other things, this PR still needs a couple unit tests and some documentation updates.

## Other information

Part of https://github.com/openedx/platform-roadmap/issues/11

## Testing instructions

Test instructions TBD.

I have manually tested the `dev` version of this but haven't tested out `local` or `k8s`.

### DIFF
--- a/tutor/commands/local.py
+++ b/tutor/commands/local.py
@@ -2,12 +2,13 @@ from typing import Optional
 
 import click
 
-from tutor import config as tutor_config
 from tutor import env as tutor_env
 from tutor import exceptions, fmt, utils
+from tutor import config as tutor_config
 from tutor.commands import compose
 from tutor.commands.config import save as config_save_command
 from tutor.commands.upgrade.local import upgrade_from
+from tutor.commands.plugins import add_plugin_commands
 from tutor.types import Config, get_typed
 
 
@@ -31,14 +32,14 @@ class LocalJobRunner(compose.ComposeJobRunner):
 
 # pylint: disable=too-few-public-methods
 class LocalContext(compose.BaseComposeContext):
-    def job_runner(self, config: Config) -> LocalJobRunner:
-        return LocalJobRunner(self.root, config)
+    def job_runner(self) -> LocalJobRunner:
+        return LocalJobRunner(self.root, self.config)
 
 
 @click.group(help="Run Open edX locally with docker-compose")
 @click.pass_context
 def local(context: click.Context) -> None:
-    context.obj = LocalContext(context.obj.root)
+    context.obj = LocalContext(context.obj.root, tutor_config.load(context.obj.root))
 
 
 @click.command(help="Configure and run Open edX from scratch")
@@ -112,7 +113,6 @@ Press enter when you are ready to continue"""
     click.echo(fmt.title("Database creation and migrations"))
     context.invoke(compose.init)
 
-    config = tutor_config.load(context.obj.root)
     fmt.echo_info(
         """The Open edX platform is now running in detached mode
 Your Open edX platform is ready and can be accessed at the following urls:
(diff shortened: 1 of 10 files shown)


tutor/commands/local.py
```diff
@@ -31,14 +32,14 @@ def __init__(self, root: str, config: Config):
...
-        return LocalJobRunner(self.root, config)
+    def job_runner(self) -> LocalJobRunner:
+        return LocalJobRunner(self.root, self.config)
 
 
 @click.group(help="Run Open edX locally with docker-compose")
 @click.pass_context
 def local(context: click.Context) -> None:
-    context.obj = LocalContext(context.obj.root)
+    context.obj = LocalContext(context.obj.root, tutor_config.load(context.obj.root))
```
regisb, line 42: This line will fail the first time someone runs `tutor local quickstart` with an empty project root.

regisb:
Hey Kyle! I believe that I understand the changes in this PR, but I'm not sure what new feature they are supposed to introduce?

Usually, when I need to run a specific job for a plugin, I just run `tutor local init --limit=myplugin`. This is overzealous, as it will apply all initialisation scripts, and not just a specific one. If we really need to run a small command frequently in one of the services, then there are two options:

1. Run a container that will run a cron job (or something similar). I'm really not a big fan of cron tasks, as they are frequently anti-patterns, but sometimes there is no way around it.
2. Add a convenient CLI utility to the plugin image, such that people can run `tutor local run myservice myservice-ctl ...`.

kdmccormick:
> Add a convenient CLI utility to the plugin image, such that people can run `tutor local run myservice myservice-ctl ....`

@regisb This almost gets me what I'm looking for, but I am still not sure how a plugin could expose a convenient CLI for running commands within a *different* service.

Taking this back to the specific CourseGraph use case: CourseGraph users need to be able to manually call a CMS management command. Users *could* type out the management command manually:
```
tutor local run cms ./manage.py cms dump_to_neo4j --courses ... --host localhost --user neo4j --password openedx ....etc
```
but I think it'd be a better experience if they could just type:
```
tutor local coursegraph refresh
```
or, as you suggested:
```
tutor local run cms refresh-coursegraph
```

But, to be able to run `tutor local run cms refresh-coursegraph`, the CourseGraph plugin would need to add the `refresh-coursegraph` script to the CMS container's filesystem, which I am not sure how to do.

(I guess this wouldn't be necessary if `./manage.py cms dump_to_neo4j` had a more convenient interface, would it? Perhaps scenarios like this would best be solved at the app level instead of the plugin level?)

regisb:
> (I guess this wouldn't be necessary if ./manage.py cms dump_to_neo4j had a more convenient interface, would it? Perhaps scenarios like this would best be solved at the app level instead of the plugin level?)

I think that this is the "right" solution. If a CLI tool does not prove convenient for end users, then that tool should be modified.

kdmccormick:
Makes sense to me.

PR #570, opened 2022-01-18
use LABEL instead of MAINTAINER

iamCristYe:
see https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


regisb:
Apparently this deprecation dates back from 6 years ago... https://github.com/moby/moby/pull/25466

regisb (approved):

PR #571, opened 2022-01-18
fix `tutor local importdemocourse`

iamCristYe:
Currently running `tutor local importdemocourse` would fail when the platform is not up. I'm a bit curious why we need redis though, as I'm not sure what role it plays.

### DIFF
--- a/tutor/templates/local/docker-compose.jobs.yml
+++ b/tutor/templates/local/docker-compose.jobs.yml
@@ -25,6 +25,6 @@ services:
         - ../apps/openedx/settings/lms:/openedx/edx-platform/lms/envs/tutor:ro
         - ../apps/openedx/settings/cms:/openedx/edx-platform/cms/envs/tutor:ro
         - ../apps/openedx/config:/openedx/config:ro
-      depends_on: {{ [("mysql", RUN_MYSQL)]|list_if }}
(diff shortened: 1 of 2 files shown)


tutor/templates/local/docker-compose.jobs.yml
```diff
@@ -25,6 +25,10 @@ services:
         - ../apps/openedx/settings/lms:/openedx/edx-platform/lms/envs/tutor:ro
         - ../apps/openedx/settings/cms:/openedx/edx-platform/cms/envs/tutor:ro
         - ../apps/openedx/config:/openedx/config:ro
-      depends_on: {{ [("mysql", RUN_MYSQL)]|list_if }}
+      depends_on:
+        {% if RUN_MYSQL %}- mysql{% endif %}
+        {% if RUN_ELASTICSEARCH %}- elasticsearch{% endif %}
+        {% if RUN_MONGODB %}- mongodb{% endif %}
+        {% if RUN_REDIS %}- redis{% endif %}
```
regisb, line 32: Please use `list_if`to list all dependent services. Otherwise, the docker-compose file will have an incorrect syntax for people who don't run any of these services.

regisb (changes requested):
Please also add a line to the changelog file.

regisb (approved):
thanks for your PR!

iamCristYe:
@regisb Any idea why we would need redis?

regisb:
No I don't. Why did you add it? Is the job triggering an error when redis is not there?

iamCristYe:
Yes, there's some error when Redis is not there

regisb:
The error will tell you why redis is necessary.

PR #575, opened 2022-01-20
fix: display themed favicon across all apps

regisb:
Previously, we were redirecting all /*favicon.ico requests to the default
favicon. This meant that the favicon might not necessarily be correctly themed,
most notably in MFEs. Here, we resolve this issue by redirecting to the
theme-agnostic theming/asset/* url. Also, we restrict the overly generic regexp
for favicon url matching. We verified that we did not miss any url by running
the following command on the demo server:

    tutor local logs caddy | grep --only-matching "host.*favicon.ico" | sort | uniq

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Display properly themed favicon.ico image in LMS, Studio and microfrontends.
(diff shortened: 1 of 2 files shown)


regisb:

ghassanmas (commented):
Would it be possible to define the `@favicon_matcher` only once on a higher level and then use it for both the CMS and the LMS

regisb:
> Would it be possible to define the @favicon_matcher only once on a higher level and then use it for both the CMS and the LMS

I'm not sure. I propose to investigate this if the pattern ever gets more complex one day.

PR #576, opened 2022-01-24
fix: lms authenticaton in development

regisb:
In development, it was no longer possible to authenticate to the lms. Ater
signing in, the session ID could not be dropped, and thus the user was not
signed in, although no error was logged -- just a warning in the browser
console.
This problem was caused by the fact that the SameSite policy was set to "None"
in development.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Fix authentication in development due to missing SameSite policy on session ID cookie.
 - [Bugfix] Display properly themed favicon.ico image in LMS, Studio and microfrontends.
 - [Bugfix] Fix "LazyStaticAbsoluteUrl is not JSON serializable" error when sending bulk emails.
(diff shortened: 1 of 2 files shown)


regisb:
This is ready for review @overhangio/tutor-developers (and rather important...)

ghassanmas:
I guess this was because `SameSite=None` requires `Secure` [ref](https://web.dev/samesite-cookies-explained/#samesitenone-must-be-secure) and in development it's `http` so setting cookie `Secure` won't be possible... 
I see the change rationale to explicitly set `SameSite=Lax` but I haven't tested this change.

mshukurlu:
@regisb I tried the new version of tutor, but no use the problem is remaining, I installed new patch via :

```
 ' sudo curl -L "https://github.com/overhangio/tutor/releases/download/v13.1.1/tutor-$(uname -s)_$(uname -m)" -o /usr/local/bin/tutor'
' sudo chmod 0755 /usr/local/bin/tutor' 
```
tutor --version output is 13.1.1
but the problem is remaining ..

mshukurlu:
@regisb after running tutor config save it worked thanks!

PR #574, opened 2022-01-19
docs: clarify container pruning during uninstallation

regisb:
Close #566.

PR #580, opened 2022-01-27
docs: fix typo in uninstallation process

iamCristYe:
I'm not sure if you've seen my comments in https://github.com/overhangio/tutor/pull/574, anyway, here's the PR


regisb (approved):
It's a little less explicit this way, but simpler, so I guess that's fine. Thanks for the PR!

PR #581, opened 2022-01-28
docs: Explain passing the Click context to a plugin command

fghaas:
Clarify how plugin authors can access the Click context, and from it the Tutor configuration, from the function implementing a plugin command.

Reference:
https://discuss.overhang.io/t/how-does-a-plugin-command-access-the-tutor-configuration/2372


regisb (approved):
Thanks for the PR!

PR #587, opened 2022-02-15
chore: upgrade mypy

regisb:
As a consequence, we need to resolve a typing issue.


regisb:

PR #593, opened 2022-02-21
Add `k8s apply` command

regisb:
This is convenient to check for k8s template validity.


regisb:

PR #594, opened 2022-02-21
chore: resolve "deprecated django-admin.py is deprecated" warning

regisb:
See: https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-4-0


regisb:

ghassanmas (approved):
All good to me, given as well this usage `dajngo-admin` is being supported by Django from >= 1.8.x, so it shuoldn't  even affect the usage of Django 2.X 

> Changed in Django 1.7:
Prior to Django 1.7, django-admin was only installed as django-admin.py.
[ref](https://docs.djangoproject.com/en/1.8/ref/django-admin/#django-admin-and-manage-py)

Issue #499, opened 2021-09-27
Plugin hooks are not supported by the yaml plugin API


regisb:
**Is your feature request related to a problem? Please describe.**

YAML-formatted Tutor plugins may not implement initialization hooks. The reason for that is that init hooks are actually templates with a specific path that are loaded at runtime. This is counter-intuitive and not very easy to grasp.

Note that the same problem applies to templates. It would make sense to be able to write custom templates in YAML plugins, too.

**Describe the solution you'd like**

Hooks should be declared as raw strings, just like patches. Thus, a plugin should be able to declare hooks as follows:

```
hooks = {
    "init": [
        {
            "service": "lms",
            "command": "..."
        }
    ]
}
```

**Describe alternatives you've considered**

It is still possible to implement hooks (and templates) with the Python API, but this is inconvenient. YAML plugins are more easily shared on the forums.

**Additional context**
See initial conversation: https://discuss.overhang.io/t/how-to-use-learning-mfe-on-tutor/1863/3

ghassanmas:
So if I get your solution right: Once implemented , will end up with two types of init hooks: a string (which refer to the directory where the hook is) or dict which means it's command to run. 

So let me know what you think of this flow to implement the solution: 
When looping over hooks of a plugin https://github.com/overhangio/tutor/blob/f2260006bd370584434b935f90e85e3dcee776c2/tutor/plugins.py#L131 
- We first check if the init hook is a string or object 
- If string just move forward with how it is, in order to not break already written plugins
- If dict then create a file hooks/<service-name>/init file 
- Write the command value in that file. 
- Make sure that file/diectory is then would be included or added to `template_root` ref: https://github.com/overhangio/tutor/blob/f2260006bd370584434b935f90e85e3dcee776c2/tutor/plugins.py#L59 

Is that a possible to implment the solution? if yes then I would move forward with creating a proof of concept.

regisb:
Thanks for your interest in this matter @ghassanmas! I think we should not stick with the "hooks as files" approach. Currently, we iterate over hook template files with `iter_plugin_hooks`, which itself yields from `iter_hooks`. This method yields a `(hook name, hook file path)` such that job runners run `run_job_from_template` using the hook path. This approach is uselessly complex, in my opinion. So we should refactor the code such that `iter_hooks` actually yields the unrendered content of the hook files.

Once this refactor is done, we should allow the definition of hooks as follows:

```
hooks = {
    "init": [
        {
            "name: "forum",
[...]

PR #599, opened 2022-03-04
Plugins v1

regisb:
This is a very large refactoring which aims at making Tutor both more
extendable and more generic. Historically, the Tutor plugin system was
designed as an ad-hoc solution to allow developers to modify their own
Open edX platforms without having to fork Tutor. The plugin API was
simple, but limited, because of its ad-hoc nature. As a consequence,
there were many things that plugin developers could not do, such as
extending different parts of the CLI or adding custom template filters.

Here, we refactor the whole codebase to make use of a generic plugin
system. This system was inspired by the Wordpress plugin API and the
Open edX "hooks and filters" API. The various components are added to a
small core thanks to a set of actions and filters. Actions are callback
functions that can be triggered at different points of the application
lifecycle. Filters are functions that modify some data. Both actions and
filters are collectively named as "hooks". Hooks can optionally be
created within a certain context, which makes it easier to keep track of
which application created which callback.

This new hooks system allows us to provide a Python API that developers
can use to extend their applications. The API reference is added to the
documentation, along with a new plugin development tutorial.

The plugin v0 API remains supported for backward compatibility of
existing plugins.

### Is this PR ready for comments?

Yes! If you are a plugin developer, we would like to hear what you have to say.

### Why is this PR so large?

Because I refactored the entire code base to make use of the v1 API. Tutor core now uses its own extension mechanism, in the spirit of [dogfooding](https://en.wikipedia.org/wiki/Eating_your_own_dog_food). We do that because the eventual long-term goal is to extract the extension API from Tutor in a different package, such that it can be used to deploy not just Open edX, but other software as well. (see [this issue](https://github.com/overhangio/2u-tutor-adoption/issues/10) for more information)

### Where should I start?

I'm glad you're interested :)

1. First, please have a look at the new plugin development tutorial. Basically, read all changes in the docs/ folder.
2. Have a look at the new hooks API, in the tutor/hooks/ folder. 
3. Have a look at the changes made in the commands module: all files in the tutor/commands/folder

I think that at this point you should already have plenty of questions and comments...

### When should this be merged?

All official plugins should be migrated to v1 prior before the Nutmeg release (June 9th 2022), so April 30th would be a good deadline.

### Will this PR break my plugin?

No. Existing plugins should keep working as expected. If they don't, then it's a bug. Please report them in the forum, as usual :)

### How long will the v0 API be supported? 

A long time. We want to leave time to all developers to migrate their plugins from v0 to v1. I don't anticipate that this will cause any issue, because under the hood the v0 plugin API already uses the v1 API.

### What can I do with v1 that I couldn't with v0?

Lots of stuff! This PR does not introduce many new features by itself, but I expect that in the future it will become very easy to make profound changes to how Tutor works using just plugins. Here's a sneak peek of what I have in mind:

- Open edX themes as plugins
- Plugin-extensible argument parsing, for instance to conveniently [automount](https://github.com/overhangio/2u-tutor-adoption/issues/43#issuecomment-1040447164) host folders.
- Extensible `tutor local` and `tutor k8s` CLI.
- Plugin debugging and troubleshooting, via another plugin _naturellement_.

## Done

- Do not load commands from plugins which are not enabled.
- Load enabled plugins once on start.
- Implement contexts for actions and filters, which allow us to keep track of
  the source of every hook.
- Migrate patches
- Migrate commands
- Migrate plugin detection
- Migrate templates_root
- Migrate config
- Migrate template environment globals and filters
- Migrate hooks to tasks
- Generate hook documentation
- Generate patch reference documentation

## TODO

- [x] Amend the API such that we can write `tutor.hooks.Filters.MY_FILTER.apply(...)`
- [x] Maybe rename `Action.on()` to `Action.add()`?
- [x] Make it clearer in the docs that actions and filters are different types of hooks.
- [x] Clarify in the docs that contexts is very much an internal concept that most plugin developers don't have to care about.
- [x] Don't talk about enable/disable in the code, but pick different terms.
- [x] Decide whether to keep action/filter `*args`. EDIT: we should keep them.


regisb:
This is ready for review @overhangio/tutor-developers. I understand this is a huge PR, and it requires some sort of explanation. We are currently planning a meeting with @kdmccormick so that everyone understands what's at stake here, and what is the technical solution that I've ended up implementing. We'll decide on a meeting date and then plugin developers will be invited to the meeting. Meanwhile, you can learn more about that project by having a look at this comment: https://github.com/overhangio/2u-tutor-adoption/issues/32#issuecomment-1031540558

docs/tutorials/plugin.rst
```diff
@@ -0,0 +1,275 @@
...
+
+If you have not already read :ref:`how_does_tutor_work` now would be a good time :-) Tutor uses templates to generate various files, such as settings, Dockerfiles, etc. These templates include ``{{ patch("patch-name") }}`` statements that allow plugins to insert arbitrary content in there. These patches are located at strategic locations. See :ref:`patches` for more information.
+
+Let's say that we would like to limit access to our brandh new Open edX platform. It is not ready for prime-time yet, so we want to prevent users from registering new accounts. There is a feature flag for that in the LMS: `FEATURES['ALLOW_PUBLIC_ACCOUNT_CREATION'] <https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['ALLOW_PUBLIC_ACCOUNT_CREATION']>`__. By default this flag is set to a true value, enabling anyone to create an account. In the following we'll set it to false.
```
bradenmacdonald, line 45: minor typo: `brandh`

docs/tutorials/plugin.rst
```diff
@@ -0,0 +1,275 @@
...
+This imports the ``hooks`` module from Tutor, which grants us access to ``hooks.filters`` and ``hooks.Filters`` (among other things).
+
+::
+
+    hooks.filters.add_item(
+        <filter>,
+        <item>
+    )
+
+This means "add <item> to the <filter>". In our case, we want to modify the LMS settings, both in production and development. The right patch for that is :patch:`openedx-lms-common-settings` and the corresponding filter is named ``hooks.Filters.ENV_PATCHES % "openedx-lms-common-settings"``. We add one item, which is a single Python-formatted line of code::
```
bradenmacdonald, line 69: Nit: I haven't seen the `%` operator in a while and actually forgot what this syntax does. I had to use the REPL to see what type of object `hooks.Filters.ENV_PATCHES` was before I understood this line 😅 

% strings aren't deprecated AFAIK but the python docs definitely encourage the newer `format` syntax. So I would suggest that something like this is a more modern approach:

```python
>>> hooks.Filters.env_patch = 'env:patches:{0}'.format
>>> hooks.Filters.env_patch("openedx-lms-common-settings")
'env:patches:openedx-lms-common-settings'
```
or even simply
```python
hooks.Filters.ENV_PATCH_PREFIX = "env:patches:"
hooks.Filters.ENV_PATCH_PREFIX + "openedx-lms-common-settings"
```
is more readable imho.

regisb:
If other plugin developers are interested in learning about the new plugin API, we'll be discussing it with Kyle & Rebecca today March 29th at 12:30 UTC: https://us02web.zoom.us/j/85637351768?pwd=Y1htdFlLNnRTZzd6VXQwNW45eTArUT09
FYI @kdmccormick @rgraber @fghaas @BbrSofiane

fghaas:
Thanks for the heads-up @regis! I'll *try* to make that but it comes at slightly short notice for me so I'm not sure if I'll be able to, hence I'm putting a few thoughts down here in writing. I hope that's OK.

In brief, I admire the rationale behind this, and I think it's a good approach. But, and this is completely separate from a discussion on the technical merits, to be honest I am a bit skeptical about the timing. 

Consider that since the December Open edX Maple release which made Tutor the default mode of deploying Open edX, a bunch of people would have been scrambling to achieve feature parity with edx-configuration, so that they can move their production platforms to Tutor. Since Tutor has always followed the policy of keeping its basic feature set simple, and exposing other features with plugins, that means that people have been busy writing and testing plugins since then ([we've written 7 so far](https://github.com/hastexo/?q=tutor-contrib&type=public)). Are you sure it's a good idea to introduce a new plugin API just now, less than 4 months after the Maple release, essentially forcing people to start updating/rewriting them? (Yes I understand you want to keep backward compatibility with the v0 API. But you wouldn't want to do that for long, see [xkcd927](https://xkcd.com/927/) 🙂). 

My point is just that after several years in limbo — with edx-configuration being "quietly" deprecated long before it officially was, no endorsed alternative coming out of edX, and there having been being months of back-and-forth until Tutor was finally adopted as the new community development platform — community Open edX deployers could be forgiven to desire some degree of stability in the deployment tools. So, is the v0 plugin API perhaps "good enough" for a few more months until people have completed their Maple/Tutor transitions, for a new plugin API to be introduced thereafter?

regisb:
I understand your concern Florian and I agree that the timing is not great. Several things are happening at once with Tutor: system administrators are learning to switch from the native installation and edX/2U is moving away from the Devstack. The latter project is straining the v0 plugin API beyond its capabilities, as we can see from these issues that were collected by Kyle: https://github.com/overhangio/2u-tutor-adoption/issues/

The goal of this new v1 API is to make it easier for developers to use Tutor and to customise their platforms with plugins. It would be counter-productive if it actually made people's life more more difficult.

My understanding is that to make this transition successful we need to work on three key aspects:

1. API stability (including the v0 API, for which we don't expect to set a deprecation date)
2. v1 features that are actually useful and productive
3. communication and pedagogy to help people transition from v0 to v1

We can cover items 1 and 2 by preserving support for the v0 API and resolving issues listed in the 2u-tutor-adoption project. To address item 3 we are planning to:

- organise a `tutor dev quickstart` tutorial at the conference: https://openedx2022.sched.com/event/zx6P/tutor-dev-quickstart I'm hoping to re-use the material in later live coding sessions (see below)
- improve the Tutor documentation
- organise live coding/tutorial streams on Twitch/Youtube: this is something that [I've played with](https://www.youtube.com/c/OverhangIO) in the past and I'm planning to do it more consistently now that I finally have an Internet connection that doesn't suck (yay fiber!)

I'll flip the question: you and your team have made several significant contributions to Tutor and I'd like you to keep doing that. How can I make your life easier and more productive?

>  I'll try to make that but it comes at slightly short notice for me so I'm not sure if I'll be able to

It's OK if you can't come; this is going to be a preliminary meeting, and I expect that there will be at least another one with Tutor maintainers before we merge this PR.

docs/tutorials/plugin.rst
regisb: Yes, you're right. I went with the `.format()` approach. But I also intend to add a "ENV_PATCH" filter that does not need to be string-formatted. Stay tuned!

kdmccormick (commented):
@regisb I have several critiques and suggested changes below, but please understand them in the context that I really like the new API and am overall very excited about this PR!

To really wrap my head around the implementation of API and clarify to myself what I feel should be changed, I made my own PR against your branch: https://github.com/kdmccormick/tutor/pull/5. I'm not proposing that you merge or even review my PR (there are several things wrong with it), but I'm linking it here in case it helps to have a working implementation of the changes I'm suggesting below

## Nomenclature

(This is the one I feel most strongly about.)

V1 certainly makes Tutor more modular and extensible. Having learned both the V0 and V1 within a couple months of each other, though, the V1 API feels more abstract and at times more mystical. I don't think that's bad in itself, but I think that in order to keep the V1 API approachable, we need to take extra care about the terms we are using, both in the API and in the implementation. Specifically, I think this PR should:
* Differentiate more clearly between "hooks" and "callbacks", in that "hooks" are just named extension points, whereas "callbacks" are the functions that are registered to listen to those callbacks.
* Clarify that a "filter" is a sub-type of "hook", and that the functions added to a filter are not filters themselves but are actually "filter callbacks".
* Clarify that an "action" is a sub-type of "hook", and that the functions added as handlers are not actions themselves but are actually "action callbacks".
* Avoid overloading "enable" and "disable". I think enable/disable should just mean adding or removing a plugin from the list in config.yml. To describe the process of adding or removing a plugin's callbacks, could we invent new terms, for example "activate" and "deactivate"?


## Verbosity

The current way of registering callbacks feels a little clunky to me:

```python
@hooks.actions.on(hooks.Actions.INSTALL_PLUGINS)
```

If we defined the constants in `Actions` as `Action` objects instead of strings, then we could register events in a more concise, object-oriented way (I've changed `on` to `handle` because it made more semantic sense to me in syntax I'm proposing, but I don't feel too strongly about it):
```python
@hooks.Actions.INSTALL_PLUGINS.handle()
```

Plugin-defined actions would look like this:
```python

# in myplugin/actions.py:
from tutor import hooks
MY_PLUGIN_ACTION = hooks.Action("myplugin:someaction")

# in some other module:
from myplugin.actions import MY_PLUGIN_ACTION
@MY_PLUGIN_ACTION.handle()
def _action_handler(...):
```

The same change could be made for `hooks.Filters`, with:

```python
hooks.filters.add_item(hooks.Filters.CLI_COMMANDS, command)
```
becoming:
```python
hooks.Filters.CLI_COMMANDS.add_item(command)
```

## Kwargs

I think there'd be a readability boost, both of Tutor core code and of plugin code, if actions only invoked their callbacks with keyword arguments. That is, instead of:
```python
hooks.Action("test-action").do(1)
```
it'd always be:
```python
hooks.Action("test-action").do(increment=1)
```
I propose the same for filter applications, excluding the obvious 'value' argument, eg:
```python
all_remote_images = hooks.Filter(filter_name).apply(
    all_remote_images, config=config
)
```

## Contexts

The idea of contexts is very elegant and clearly is a useful abstraction in the Tutor core 👍🏻 

Do we need to expose them in the Plugin API, though? I worry that plugin developers will not realize that they can basically ignore the documentation on Contexts for their intents and purposes. The more concepts there are in the API, the more difficult it will be for someone to jump into plugin development.

In my PR, I drop the Contextualized class and hide all the context-handling machinery in a private _HooksCallbackRegistry class. I still have a tutor.hooks.Contexts class, but that could easily be moved somewhere outside the `tutor.hooks` package, which seems to be the intended stable API for plugins.

regisb:
Hey @kdmccormick, thanks for that feedback. I mostly agree with you, so I amended the API such that objects in tutor.hooks.consts are actually Filter and Action objects. It was a bit tricky to do the same thing for templated names ("env:patches:{}") but I think that I reached an acceptable point.

I also agree with your comments regarding the docs, but I have not taken action yet. I'll do that when I return from my week off.

Regarding kwargs, I'm not quite sure how I feel about that yet, and I need to think about it.

kdmccormick:
Thanks @regisb , sounds good!

kdmccormick:
@regisb I like load/unload 👍🏻

regisb:
@kdmccormick I've made the requested changes, please let me know what you think. I've tried to improve the wording across the docs and the codebase. We can always improve it later, but we should be very careful with the hook constant names, which are not supposed to change with time (although I guess that we can take some liberties).

Regarding the suppression of hook `*args`: at first I agreed with you, but then when I attempted to get rid of them I realized something: if we make use only of keyword arguments, then we need to provide a default value for each one of them. And then, every hook needs to check whether the keyword argument is correctly defined or not. This adds too much complexity to my taste, so I decided to keep `*args`. We can discuss this further if you like.

docs/plugins/v0/legacy.rst
```diff
@@ -0,0 +1 @@
+.. warning:: The v0 plugin API is no longer the recommended way of developing plugins for Tutor, starting from Tutor v13.2.0. See our new :ref:`plugin creation tutorial <plugin_development_tutorial>` to learn more about the new v1 plugin API.
```
kdmccormick, line 1: I notice that the entrypoint for V1 plugins is currently `v1alpha`. I agree that it's reasonable to include "alpha" given the newness of the API. Should we hold off on dissuading usage of the V0 API until the V1 API isn't alpha anymore?

docs/reference/api.rst
```diff
@@ -0,0 +1,11 @@
+===
+API
```
kdmccormick, line 2: Can we make this heading more specific?

```suggestion
Hooks API
```
```suggestion
Python API
```

docs/reference/patches.rst
```diff
@@ -0,0 +1,329 @@
+.. _patches:
```
kdmccormick, line 1: Wow! I didn't even realize you included all the patches here until now. This is a great start to fully documenting all the patches. 🎉

tutor/hooks/consts.py
```diff
@@ -0,0 +1,266 @@
...
+"""
+# The Tutor plugin system is licensed under the terms of the Apache 2.0 license.
+__license__ = "Apache 2.0"
+
+from tutor import hooks
+
+__all__ = ["Actions", "Filters", "Contexts"]
+
+
+class Actions:
```
kdmccormick, line 13: If you take a look at [openedx-events](https://github.com/openedx/openedx-events/blob/main/openedx_events/learning/signals.py) (which seems like the Open edX app-level analogue to Tutor's actions), the event names are all adjectives, describing the state of the system in terms of something that has happened or is happening. For example, there is an event named `COURSE_ENROLLMENT_CREATED`, as opposed to `COURSE_ENROLL` or `COURSE_ENROLL_POST`. I like this because it reinforces what an event is: something happens in the core, and plugins can hook in some callbacks in order to respond to that thing.

Would you be willing to rename the actions under that framing? `CORE_READY` and `CORE_ROOT_READY` are great. For the others, I might suggest:
* `plugins:discover` -> `plugins:discovering`
* `plugin:load:{0}` -> `plugins:loading:{0}`
* `plugins:load-post` -> `plugins:loaded`

tutor/hooks/consts.py
```diff
@@ -0,0 +1,266 @@
...
+    #: :parameter list[str] names: list of plugin names.
+    LOAD_PLUGINS = hooks.actions.get("plugins:load")
+
+    #: Triggered after all plugins have been loaded.
+    #:
+    #: This action does not have any parameter.
+    LOAD_PLUGINS_POST = hooks.actions.get("plugins:load-post")
+
+
+class Filters:
```
kdmccormick, line 72: I like the names of all of these filters except `config:base` (see my comment about config in the plugin tutorial) and the `app:tasks:*...` filters. My litmus test here is: if I can look at the name of the filter and reasonably guess what it means and its type signature *without* looking at the docs, then it's a solid name.

I don't have any better suggestions yet... in fairness, the `app:tasks:*` filters are certainly the most complex ones. I need to submit this review soon so I can join a meeting, but I will stew on this and let you know if I have ideas :)

docs/plugins/v0/legacy.rst
fghaas: ... and also inject the word "new" between "developing" and "plugins".

That way it emphasizes that if you're starting a plugin from scratch, you might want to go with v1, but updates to existing plugins that use v0 are still worthwhile.

In contrast, I'd suggest removing the word "new" that precedes "v1 plugin API". It adds no additional value here, and new things become standard and then old quickly, at which point these little nits don't tend to get fixed down the road. That kind of word rot is easily averted by just dropping "new" here.

MaferMazu:
I'm starting to see this PR and at the moment I have the following questions:

- Will [yaml file plugins](https://docs.tutor.overhang.io/plugins/gettingstarted.html#yaml-file) still exist?
- I understand that there will also be some "simple" plugins (not packages) that will be .py files and with them you will also be able to add patches as mentioned in [tutorials/plugin](https://github.com/overhangio/tutor/blob/regisb/plugin-v1/docs/tutorials/plugin.rst), right?
- Will the structure of the plugins that are python packages change a lot?
- Is there a public example of a v1 plugin package?

docs/tutorials/plugin.rst
```diff
@@ -0,0 +1,261 @@
...
+Congratulations! You've created your first working plugin. As you can guess, you can add changes to other files by adding other similar patch statements to your plugin.
+
+Modifying configuration
+=======================
+
+In the previous section you've learned how to add custom content to the Tutor templates. Now we'll see how to modify the Tutor configuration. Configuration settings fall in three categories:
+
+1. "base" settings that need to be preserved in config.yml: for instance, password and secret keys.
+2. "default" settings that should only be stored in config.yml when they are modified by users. Most settings are in this category.
+3. "override" settings that modify the base Tutor configuration. These will be removed and restored to their default values when the plugin is disabled.
```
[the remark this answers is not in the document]
regisb: I'm OK with all of your suggestions, except for the renaming of "base" settings. While I agree that "base" is too generic, "required" is not appropriate either: after all, all settings are "required" to render environment templates. We need to find a word that says "unique to each user and thus must be persisted to disk". Maybe: "unique"? Or "user"?

docs/tutorials/plugin.rst
```diff
@@ -0,0 +1,261 @@
...
+
+
+Adding new templates
+====================
+
+If you are adding an extra application to your Open edX platform, there is a good chance that you will create a new Docker image with a custom Dockerfile. This new application will have its own settings and build assets, for instance. This means that you need to add new templates to the Tutor environment. To do that, we will create a new subfolder in our plugins folder::
+
+    $ mkdir -p "$(tutor plugins printroot)/templates/myplugin"
+
+Then we tell Tutor about this new template root thanks to the :py:data:`tutor.hooks.Filters.ENV_TEMPLATE_ROOTS` filter::
```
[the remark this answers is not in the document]
regisb: Yes, you're right.

docs/tutorials/plugin.rst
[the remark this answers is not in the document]
kdmccormick: That is fair! I like "unique" a lot. "user" is OK too, although I think it's slightly more ambiguous, since some software uses the term "user settings" just to refer to user overrides of the default settings.

tutor/hooks/consts.py
regisb: Good point. I also like the fact that this allows us to rename `VERB_NOUN` to `NOUN_LOADED` (e.g: `LOAD_PLUGINS` -> `PLUGINS_LOADED`) which allows us to better sort actions by namespace.

However I'm not a big fan of the "-ing" form. This leads to me to realize that the "plugins:discover" action is not really necessary, as it's only triggered on `CORE_READY`. So I think that I'm going to get rid of it. With this in mind, I'll do the following changes:

- plugins:discover -> replace by core:ready
- plugin:load:{0} -> plugins:loaded:{0} I've really racked my brains for this one. "Loaded" does not really reflect the event that is occurring. It's more like "we are going to load the plugin because it is requested/required/necessary". But I like none of these adjectives, so I've decided to stick with "loaded". It's probably ok if we modify the action name later, once we find a better name, as it's only used internally.
- plugin:load -> get rid of it
- plugins:load-post -> plugins:loaded

docs/tutorials/plugin.rst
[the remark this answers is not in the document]
regisb: We're going for "unique" then :)

tutor/hooks/consts.py
regisb: Hmmm I think I understand what's bothering you with tasks. Tasks designate both image building/pushing/etc. jobs and scripts that are run as part of platform initialisation:

    APP_TASKS_IMAGES_PULL = hooks.filters.get("app:tasks:images:pull")
    APP_TASKS_INIT = hooks.filters.get("app:tasks:init")

Also, the "app:" namespace does not really make sense because we never use it without the "tasks:" namespace.

The "task" prefix is also not appropriate, because the filters actually yield paths to templates that includes commands.

I suggest the following renaming:

- app:tasks:images:* -> images:*
- app:tasks:init -> commands:init
- app:tasks:pre-init -> commands:pre-init

regisb:
To answer your questions @MaferMazu:

> Will [yaml file plugins](https://docs.tutor.overhang.io/plugins/gettingstarted.html#yaml-file) still exist?

It will not be possible to create YAML plugins with the V1 API, but existing v0 plugins will remain supported for a long time.

> I understand that there will also be some "simple" plugins (not packages) that will be .py files and with them you will also be able to add patches as mentioned in [tutorials/plugin](https://github.com/overhangio/tutor/blob/regisb/plugin-v1/docs/tutorials/plugin.rst), right?

Yes, exactly. This will be the new "lightweight" way of creating simple plugins, similar to YAML plugins, but not subject to the same limitations ([this one](https://github.com/overhangio/tutor/issues/499) for example).

> Will the structure of the plugins that are python packages change a lot?

Not drastically, no. The `plugin.py` will be different, as well as the `setup.py` entrypoint, but I expect that the differences will be relatively straightforward to resolve.

> Is there a public example of a v1 plugin package?

Not yet. I'm working on upgrading the plugin cookiecutter to generate valid v1 plugins (see [this issue](https://github.com/overhangio/2u-tutor-adoption/issues/57)).

Regarding your comment @kdmccormick:

> Python 3 has a funky syntax to solve exactly this problem, if you were so inclined:

That's a bit too funky for me :) I had no idea that this syntax existed!

tutor/hooks/consts.py
kdmccormick: @regisb yes, I think you captured exactly what was bothering me!

I was in the process of typing a comment to suggest:
* app:tasks:images:* -> app:images:*
* app:tasks:init -> app:lifecycle:init-scripts
* app:tasks:pre-init -> app:lifecycle:pre-init-scripts

I like your suggestions as well.

The other thing I had on my mind was the type signatures of these filters. Some of them, especially the build filter, are pretty unwieldy to use:

```python
import typing as t
from tutor import hooks, types
@hooks.Filters.APP_TASK_IMAGES_BUILD.add()
def _add_myplugin_image_to_build(
    build_images: t.List[t.Tuple[str, t.Tuple[str, ...], str, t.List[str]]],
    config: types.Config,
) -> t.List[t.Tuple[str, t.Tuple[str, ...], str, t.List[str]]]:
    ...
```

In addition to modifying the names, friendlier type signature might make these easier to use:
```python
from tutor import hooks, types
@hooks.Filters.IMAGES_TO_BUILD.add()
def _add_myplugin_image_to_build(
    build_images: t.List[types.ImageToBuild], config: types.Config
) -> t.List[types.ImageToBuild]:
    ...
```

where `ImageToBuild` could just be a simple namedtuple or dataclass.

regisb: I must admit that I did not think about automated type checking for plugin developers... In an ideal world, I would like to create the `IMAGES_BUILD` filter as follows:

    IMAGES_BUILD: SomeFilterType = filters.get("...")

And then developers would write:

    IMAGES_BUILD.add_item(my_item)

and the type of `my_item` would automatically be checked against `SomeFilterType`. 

Would that resolve your issue? Do you know how to implement `SomeFilterType`? Hmmm I'm guessing that I should be using [generics](https://mypy.readthedocs.io/en/stable/generics.html)... At the moment I'm confused how to implement this as long as [this fix](https://github.com/python/mypy/issues/11833) was not added to a release. May I suggest that we wait before we type-annotate all filters and actions?

kdmccormick: Ack, that mypy bug is such a bummer! Otherwise, this would be a perfect use case for generics:

```python



# tutor/hooks/filters.py
T = t.TypeVar("T")
P = t.TypeVar("P", bound=tuple)
class Filter(Generic[T, P]):
    ....
    def get(name: str, return_type: type[T], param_types: type[P]): Filter[T, P]:
        ....
    def add(self) -> Callable[[CallableFilter[T, P]], CallableFilter[T, P]]:
        ....

# tutor/hooks/const.py
class ImageBuild(NamedTuple):
     service: str
     path: list[str]
     image: str
     args: list
class Filters:
    ....
    IMAGES_BUILD = filters.get("images:build", [list[ImageBuild], [Config]])

# myplugin.py
@hooks.Filters.IMAGES_BUILD.add()
def _add_myplugin_image_to_build(build_images, config):
    # Since IMAGES_BUILD is fully annotated, mypy should be able
    # to typecheck this function without the user having to annotate anything!
    ....  
```
Alas. I'm okay with holding off on annotations until we're able to absorb that mypy bugfix.

regisb:
I think I've exhausted the list of things to do here. I'll merge now :pirate_flag:

kdmccormick:
Woohoo! 🎉

Issue #601, opened 2022-03-06
SAML Plugin Example fails


CodeWithEmad:
**Bug description**
in the readme docs, the example for the SAML Plugin leads to a server error. 

```
Traceback (most recent call last):
File "lms/wsgi.py", line 19, in <module>
startup.run()
File "/openedx/edx-platform/./lms/startup.py", line 20, in run
django.setup()
File "/openedx/venv/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
apps.populate(settings.INSTALLED_APPS)
File "/openedx/venv/lib/python3.8/site-packages/django/apps/registry.py", line 114, in populate
app_config.import_models()
File "/openedx/venv/lib/python3.8/site-packages/django/apps/config.py", line 301, in import_models
self.models_module = import_module(models_module_name)
File "/opt/pyenv/versions/3.8.12/lib/python3.8/importlib/__init__.py", line 127, in import_module
return _bootstrap._gcd_import(name[level:], package, level)
File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
File "<frozen importlib._bootstrap>", line 991, in _find_and_load
File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
File "<frozen importlib._bootstrap_external>", line 843, in exec_module
File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
File "/openedx/edx-platform/common/djangoapps/third_party_auth/models.py", line 50, in <module>
_PSA_BACKENDS = {backend_class.name: backend_class for backend_class in _load_backend_classes()}
File "/openedx/edx-platform/common/djangoapps/third_party_auth/models.py", line 50, in <dictcomp>
_PSA_BACKENDS = {backend_class.name: backend_class for backend_class in _load_backend_classes()}
File "/openedx/edx-platform/common/djangoapps/third_party_auth/models.py", line 47, in _load_backend_classes
auth_class = module_member(class_path)
File "/openedx/venv/lib/python3.8/site-packages/social_core/utils.py", line 59, in module_member
module = import_module(mod)
File "/openedx/venv/lib/python3.8/site-packages/social_core/utils.py", line 53, in import_module
__import__(name)
ModuleNotFoundError: No module named 'third_party_auth' 
``` 

**How to reproduce**
use this plugin, save the configs and start.
```
name: saml
version: 0.1.0
patches:
  common-env-features: |
[...]

PR #604, opened 2022-03-06
No module named 'third_party_auth' issue solved

CodeWithEmad:
Here is the fix for https://github.com/overhangio/tutor/issues/601
I tested  `THIRD_PARTY_AUTH_BACKENDS = ["third_party_auth.saml.SAMLAuthBackend"]` and it's working. yay

### DIFF
--- a/docs/plugins/examples.rst
+++ b/docs/plugins/examples.rst
@@ -53,7 +53,7 @@ Enable SAML authentication
 
       openedx-lms-common-settings: |
         # saml special settings
-        AUTHENTICATION_BACKENDS += ["third_party_auth.saml.SAMLAuthBackend", "django.contrib.auth.backends.ModelBackend"]
+        THIRD_PARTY_AUTH_BACKENDS = ["third_party_auth.saml.SAMLAuthBackend"]
 
       openedx-auth: |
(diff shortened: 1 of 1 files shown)


docs/plugins/examples.rst
```diff
@@ -53,7 +53,7 @@ Enable SAML authentication
 
       openedx-lms-common-settings: |
         # saml special settings
-        AUTHENTICATION_BACKENDS += ["third_party_auth.saml.SAMLAuthBackend", "django.contrib.auth.backends.ModelBackend"]
+        THIRD_PARTY_AUTH_BACKENDS = ["third_party_auth.saml.SAMLAuthBackend"]
```
regisb, line 56: Are you quite sure that this is the right fix? Looking at the source code, the `THIRD_PARTY_AUTH_BACKENDS` setting is not read at all. I believe that the import error is caused by the fact that we should now specify the full import path: `common.djangoapps.third_party_auth.saml.SAMLAuthBackend`.

Can you please test again your plugin and my proposed change, see which one works?

regisb:
Closed in favor of https://github.com/overhangio/tutor/commit/2cf4e6dc3877aafd1b4c7b282275c7d6f467625e

PR #608, opened 2022-03-10
fix: dockerize on arm64

### LINKED ISSUE #591: arm64: `dockerize: Exec format error` on mysql init
**Bug description**

As part of `tutor local init` which suggests that the `dockerize` command is built for a foreign architecture and thus fails to run:

`docker-compose -f /home/ubuntu/.local/share/tutor/env/local/docker-compose.yml -f /home/ubuntu/.local/share/tutor/env/local/docker-compose.prod.yml --project-name tutor_local -f /home/ubuntu/.local/share/tutor/env/local/docker-compose.jobs.yml run --rm lms-job sh -e -c dockerize -wait tcp://mysql:3306 -timeout 20s`

**How to reproduce**
After fixing the former permission error #590 I ran into this problem.

**Environment**
Ubuntu 20.04 on ARM64 AWS instance

**Additional context**
I found the source:
https://github.com/overhangio/tutor/blob/156ef063ce9b536bb39d239587cd9536970d0004/tutor/templates/build/openedx/Dockerfile#L26
jwilder posted an armhf/aarch64 version but it's called armhf which is inconsistent with the kernel convention so we can try and add a --build-arg to docker to manually switch aarch64 to armhf and keep amd64 intact (uname -m/i/p).
This can be done as part of docker-compose.yml (build arguments).

Since I can live without amd64 I managed to get it working by manually changing amd64 to armhf in the respective openedx Dockerfile instruction.

regisb:
The version of dockerize that shipped with the "openedx" image was not
compatible with arm64. The original project is unmaintained, but there
is a fork that provides a version that is compatible with arm64.

Close #591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Bugfix] Fix dockerize on arm64 by switching to the [powerman/dockerize](https://github.com/powerman/dockerize) fork (#591).
 - [Bugfix] Fix "Unexpected args" error during service initialization on Kubernetes (#611).
 
 ## v13.1.6 (2022-03-15)
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -23,10 +23,12 @@ RUN $PYENV_ROOT/versions/$PYTHON_VERSION/bin/python -m venv /openedx/venv
 
 ###### Install Dockerize to wait for mysql DB availability
 FROM minimal as dockerize
-ARG DOCKERIZE_VERSION=v0.6.1
-RUN curl -L -o /tmp/dockerize.tar.gz https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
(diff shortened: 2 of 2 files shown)


yarons:
I'm not sure that the `dockerize` container is needed because now you are pulling the binary directly and there's no compilation involved, it's better to have a build step in the working docker rather than creating another one.
The only thing that we should pay attention to is that the `--from dockerize` is performed with a lower permissions UID so I had to download the file before the permission change and then run it.

regisb:
@yarons I do not understand your comment. 

1. Are you saying that the proposed change did not work for you? If it didn't work, what was the error?
2. Are you saying that it's better to build the dockerize binary in Tutor? Why?

yarons:
@regisb Oh no no, I'm suggesting an optimization, let me explain further:
Since the old dockerize had to be built from scratch with a dedicated go container you would need a go image as part of the multistage docker, the new dockerize requires no building so this step is redundant and you can just copy the binary directly from github, no additional step needed.

regisb:
@yarons I still do not understand. My proposed change does download the binary directly from Github. What are you suggesting _exactly_?

yarons:
The current process is:
One of the docker building steps is the Go docker, on top of it, the former dockerize would have been built.

The new process does not require building dockerize, so we don't need an additional docker step, we can just download the binary to the resulting docker directly.

PR #427, opened 2021-04-30
Add uwsgi buffer size to prevent its exceeding

wey-gu:
502 was observed when default size 4096 being hit

Close #426
This is equivalent to https://github.com/overhangio/tutor-ecommerce/commit/6df2c99362ee12239eb27b79455539411e4d479e

PR #612, opened 2022-03-16
feat: patch for openedx-dockerfile-final

danielrudn:
Building on the work done in https://github.com/overhangio/tutor/pull/427. This PR allows you to configure the uWSGI buffer size in `config.yml`

From the [uWSGI docs](https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html):

> If you start receiving “invalid request block size” in your logs, it could mean you need a bigger buffer. Increase it (up to 65535) with the buffer-size option.

One of our deployments uses SAML for authentication so the request headers and body for the authentication requests can get pretty verbose. We started receiving the "invalid request block size" error, so we bumped the buffer size to `32768`.

Does it make sense to bump the default from `8192` to `32768` (or something else)?

### DIFF
--- a/docs/reference/patches.rst
+++ b/docs/reference/patches.rst
@@ -266,6 +266,13 @@ Files: ``apps/openedx/settings/cms/development.py``, ``apps/openedx/settings/lms
 
 File: ``build/openedx/Dockerfile``
 
+.. patch:: openedx-dockerfile-final
+
+``openedx-dockerfile-final``
+============================
+
+File: ``build/openedx/Dockerfile``
+
 .. patch:: openedx-dockerfile-git-patches-default
 
 ``openedx-dockerfile-git-patches-default``
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 ## Unreleased
 
+- [Feature] Add patch to allow overriding final openedx docker image CMD
 - [Fix] Ignore Python plugins that cannot be loaded. (by @regisb)
 - [Improvement] Faster and more reliable builds with `npm clean-install` instead of `npm install`. (by @regisb. Thanks @ghassanmas!)
 - [Fix] Fix 500 error during studio login. (by @regisb)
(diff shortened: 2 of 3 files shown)


danielrudn:
Hey @regisb , sorry for the delay! 

I don't think the SAML use-case would be too common, and this seems like an edge case since we did not encounter it for other Open edX deployments using SAML.

I've added the `{{ patch("openedx-dockerfile-final") }}` to the end of the openedx Dockerfile, this plugin code seems to work for updating overriding the CMD with a higher buffer-size flag:

```python
from tutor import hooks

    (
        "openedx-dockerfile-final",
        """
CMD uwsgi \\
    --static-map /static=/openedx/staticfiles/ \\
    --static-map /media=/openedx/media/ \\
    --http 0.0.0.0:8000 \\
    --thunder-lock \\
    --single-interpreter \\
    --enable-threads \\
    --processes=${UWSGI_WORKERS:-2} \\
    --buffer-size=32768 \\
    --wsgi-file $SERVICE_VARIANT/wsgi.py
"""
    )
)
```

docs/reference/patches.rst
```diff
@@ -266,6 +266,13 @@ Files: ``apps/openedx/settings/cms/development.py``, ``apps/openedx/settings/lms
 
 File: ``build/openedx/Dockerfile``
 
+.. patch:: openedx-dockerfile-final
+
+``openedx-dockerfile-final``
+======================
```
regisb, line 272: This is going to cause an error when building docs (`make docs`). The `===` line should be just as long as the title above.

regisb (changes requested):
Fix the docs, squash your commits and we'll be good to go :)

erickhgm:
@danielrudn
Just to check, are you going to proceed with this feature?
It is useful to me, if it is no longer necessary for you, no worry. I can continue ...

danielrudn:
> @danielrudn Just to check, are you going to proceed with this feature? It is useful to me, if it is no longer necessary for you, no worry. I can continue ...

Yep, it should be good to go now.

regisb (approved):

PR #618, opened 2022-03-24
fix: nightly package version

regisb:
The version of the nightly python package should not include the "-nightly"
suffix. That's because when we `pip install -e` tutor plugins, pip also
installs the latest tutor release, as part of the requirements. This overrides
the local (nightly) installation of tutor.

See: https://app.slack.com/client/T02SNA1T6/C02V3GHE3UP

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ ABOUT = load_about()
 
 setup(
     name="tutor",
-    version=ABOUT["__version__"],
+    version=ABOUT["__version__"].split("-")[0],  # drop "-nightly" suffix if present
     url="https://docs.tutor.overhang.io/",
     project_urls={
         "Documentation": "https://docs.tutor.overhang.io/",


setup.py
```diff
@@ -40,7 +40,7 @@ def is_requirement(line: str) -> bool:
 
 setup(
     name="tutor",
-    version=ABOUT["__version__"],
+    version=ABOUT["__version__"].split("-")[0],
```
kdmccormick, line 43:
```suggestion
    version=ABOUT["__version__"].split("-")[0],  # Drop '-nightly' suffix from version if present.
```

kdmccormick (approved):
Tested with `__version_suffix__ = "nightly"`. Was able to install tutor-discovery without uninstalling my local Tutor clone 👍🏻 

Just one non-blocking suggestion for a comment.

PR #619, opened 2022-03-25
feat: Make the platform repository and NPM registry configurable from config.yml

fghaas:
Previously, the only way for Tutor users to use a fork of `edx-platform` or a custom NPM registry was to use build args during the image build.

This is suboptimal in the case of automatically building images from CI pipelines, which may want to auto-detect when an image needs to be rebuilt based on `config.yml` changes.

In addition, the `EDX_PLATFORM_VERSION` build argument can already be set via a corresponding `config.yml` parameter (`OPENEDX_COMMON_VERSION`), so it's reasonable to follow that precedent and also introduce `config.yml` parameters to correspond with the `EDX_PLATFORM_REPOSITORY` and `NPM_REGISTRY` build arguments.

Thus, introduce two new configuration parameters:

- `OPENEDX_PLATFORM_REPOSITORY`
- `OPENEDX_NPM_REGISTRY`

These parameters can now optionally be used instead of the aforementioned build args.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -79,7 +79,11 @@ You may want to pull/push images from/to a custom docker registry. For instance,
 Open edX customisation
 ~~~~~~~~~~~~~~~~~~~~~~
 
-- ``OPENEDX_COMMON_VERSION`` (default: ``"open-release/maple.2"``)
+- ``EDX_PLATFORM_REPOSITORY`` (default: ``"https://github.com/openedx/edx-platform.git"``)
+
+This defines the git repository from which you install Open edX platform code. If you run an Open edX fork with custom patches, set this to your own git repository. You may also override this configuration parameter at build time, by providing a ``--build-arg`` option.
+
+- ``OPENEDX_COMMON_VERSION`` (default: ``"open-release/maple.3"``)
 
 This defines the default version that will be pulled from all Open edX git repositories.
 
@@ -98,6 +102,11 @@ These two configuration parameters define which Redis database to use for Open e
 - ``OPENEDX_EXTRA_PIP_REQUIREMENTS`` (default: ``openedx-scorm-xblock<13.0.0,>=12.0.0``)
 
 This defines extra pip packages that are going to be installed for Open edX.
+
+- ``NPM_REGISTRY`` (default: ``"https://registry.npmjs.org/"``)
+
+This defines the registry from which you'll be pulling NPM packages when building Docker images. Like ``EDX_PLATFORM_REPOSITORY``, this can be overridden at build time with a ``--build-arg`` option.
+
 
 Vendor services
 ~~~~~~~~~~~~~~~
@@ -212,9 +221,9 @@ openedx Docker Image build arguments
 
 When building the "openedx" Docker image, it is possible to specify a few `arguments <https://docs.docker.com/engine/reference/builder/#arg>`__:
 
-- ``EDX_PLATFORM_REPOSITORY`` (default: ``"https://github.com/openedx/edx-platform.git"``)
+- ``EDX_PLATFORM_REPOSITORY`` (default: ``"{{ EDX_PLATFORM_REPOSITORY }}"``)
 - ``EDX_PLATFORM_VERSION`` (default: ``"{{ OPENEDX_COMMON_VERSION }}"``)
-- ``NPM_REGISTRY`` (default: ``"https://registry.npmjs.org/"``)
+- ``NPM_REGISTRY`` (default: ``"{{ NPM_REGISTRY }}"``)
 
 These arguments can be specified from the command line, `very much like Docker <https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg>`__. For instance::
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- [Improvement] In addition to the Docker build arguments `EDX_PLATFORM_REPOSITORY` and `NPM_REGISTRY`, also support two corresponding and identically-named `config.yml` values serving the same purpose.
+
 ## v13.1.10 (2022-04-11)
 
 - [Feature] Upgrade all applications to open-release/maple.3.
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -32,7 +32,7 @@ RUN dockerize_url="https://github.com/powerman/dockerize/releases/download/$DOCK
 
 ###### Checkout edx-platform code
 FROM minimal as code
-ARG EDX_PLATFORM_REPOSITORY=https://github.com/openedx/edx-platform.git
+ARG EDX_PLATFORM_REPOSITORY={{ EDX_PLATFORM_REPOSITORY }}
 ARG EDX_PLATFORM_VERSION={{ OPENEDX_COMMON_VERSION }}
 RUN mkdir -p /openedx/edx-platform && \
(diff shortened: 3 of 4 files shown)


fghaas:
@regisb: I'm not 100% sure if this properly counts as `feat` or as `refactor`; let me know if you want me to change the commit subject. :)

regisb (commented):
I must say that I don't feel great about this change. I hear your argument that there are already some values in config.yml that affect the build, but I'd rather avoid adding too many of them. My reasoning goes like this: when we set values in config.yml, we assume that new values will be taken into account the next time we run `tutor local quickstart`. But that's not the case with build-time config values, which require that we run `tutor images build ...`.

The good thing with build arguments is that they make it explicit that rebuild is required. If the only benefit of adding values to config.yml is to monitor changes in CI, then I think that it's not a strong enough reason.

Is there any way we can make your life easier without adding these extra plugins? The upcoming v1 plugin API (see #599) will make it possible for plugins to add custom parameters to `tutor images build ...` commands. Would that resolve your use case?

fghaas:
> I hear your argument that there are already some values in config.yml that affect the build, but I'd rather avoid adding too many of them.

I'd argue that that ship has sailed. Here are examples of config values that require image rebuilds if changed:

* `OPENEDX_COMMON_VERSION`
* `OPENEDX_EXTRA_PIP_REQUIREMENTS`
* `DOCKER_IMAGE_OPENEDX`
* any of the `DOCKER_IMAGE_*` parameters if they reference a custom registry (such as a Harbor registry, or the GitLab or GitHub container registry)
* any configuration change at all to the `mfe` plugin [if running on Kubernetes](https://github.com/overhangio/tutor-mfe#running-mfes-on-kubernetes),
* possibly any number of configuration settings added by a plugin, since [the plugin documentation](https://docs.tutor.overhang.io/plugins.html) does not discourage config values requiring image rebuilds.

> My reasoning goes like this: when we set values in config.yml, we assume that new values will be taken into account the next time we run `tutor local quickstart`. But that's not the case with build-time config values, which require that we run `tutor images build` ....

I don't follow that reasoning, based on our own Tutor experience. We use a custom registry, and we use a bunch of plugins. We also pip-install a bunch of extra XBlocks via `OPENEDX_EXTRA_PIP_REQUIREMENTS`. And we still want to follow the pattern of including version numbers in image tags, because that's the only reasonable way to move a Kubernetes deployment to a new image. With that, despite doing nothing out of the ordinary for a production Tutor deployment (I think), we already end up having to rebuild the images for *a lot* of `config.yml` changes; the idea that we would only have to do so when we're changing the build args doesn't really apply at all.

Also, since `OPENEDX_COMMON_VERSION` already exists in addition to the `EDX_PLATFORM_VERSION` build arg, I really don't see a good reason why the other two build args shouldn't also have `config.yml` equivelents.

> The upcoming v1 plugin API (see https://github.com/overhangio/tutor/pull/599)

I have thoughts on this. :) Is that PR OK to comment on?

regisb:
> I have thoughts on this. :) Is that PR OK to comment on?

I'll reply to your other (very relevant) comments later -- in the meantime, I just wanted to let you know that I would be meeting with Kyle today. See my comment here: https://github.com/overhangio/tutor/pull/599#issuecomment-1081458482

regisb:
> I'd argue that that ship has sailed. 

Your arguments are pretty convincing and I think that you are right. I'll add some comments to the PR.

docs/configuration.rst
```diff
@@ -79,6 +79,10 @@ You may want to pull/push images from/to a custom docker registry. For instance,
 Open edX customisation
 ~~~~~~~~~~~~~~~~~~~~~~
 
+- ``OPENEDX_PLATFORM_REPOSITORY`` (default: ``"https://github.com/openedx/edx-platform.git"``)
```
regisb, line 82: Can we rename this to `EDX_PLATFORM_REPOSITORY`? `OPENEDX_*` settings refer to values that are used across multiple Open edX apps.

docs/configuration.rst
```diff
@@ -79,6 +79,10 @@ You may want to pull/push images from/to a custom docker registry. For instance,
...
+
+This defines the git repository from which you install Open edX platform code. If you run an Open edX fork with custom patches, set this to your own git repository.
```
regisb, line 84: Can you you please highlight the fact that the setting will be overridden by the `--build-arg` argument? In reality we can say that this new setting does not completely define the edx-platform repo but just a reasonable default.

docs/configuration.rst
```diff
@@ -99,6 +103,11 @@ These two configuration parameters define which redis database to use for Open e
 
 This defines extra pip packages that are going to be installed for Open edX.
 
+- ``OPENEDX_NPM_REGISTRY`` (default: ``"https://registry.npmjs.org/"``)
+
+This defines the registry from which you'll be pulling NPM packages.
```
regisb, line 108: "... when building Docker images."

Also, we should highlight the fact that this setting can frequently be overridden by the `NPM_REGISTRY` build argument.

regisb (changes requested):

docs/configuration.rst
fghaas: I figured `OPENEDX_*` settings would also apply to anything that's relevant to building the `openedx` image, but sure I can rename it.

docs/configuration.rst
fghaas: And this should presumably also lose the `OPENEDX_` prefix, correct?

docs/configuration.rst
fghaas: We already say that under the subheading _openedx Docker Image build arguments,_ don't we?

docs/configuration.rst
fghaas: By the way, the whole subsection talks about building Docker images. If we judge we need the "... when building Docker images" remark here (i.e. it isn't evident from context), we should probably also insert it in the paragraph directly above, which discusses `OPENEDX_EXTRA_PIP_REQUIREMENTS`, to which the same consideration applies. What do you think?

docs/configuration.rst
fghaas: Done.

docs/configuration.rst
regisb: > And this should presumably also lose the OPENEDX_ prefix, correct?

Yes, you're right.

> By the way, the whole subsection talks about building Docker images. If we judge we need the "... when building Docker images" remark here (i.e. it isn't evident from context), we should probably also insert it in the paragraph directly above, which discusses OPENEDX_EXTRA_PIP_REQUIREMENTS, to which the same consideration applies. What do you think?

I guess that we should group all build-related configuration settings under a single section, which would include a paragraph explaining that setting these values requires rebuilding the openedx Docker image. But this is beyond the scope of your PR. We can tackle this later.

regisb (changes requested):

docs/configuration.rst
```diff
@@ -79,7 +79,11 @@ You may want to pull/push images from/to a custom docker registry. For instance,
 Open edX customisation
 ~~~~~~~~~~~~~~~~~~~~~~
 
-- ``OPENEDX_COMMON_VERSION`` (default: ``"open-release/maple.2"``)
+- ``EDX_PLATFORM_REPOSITORY`` (default: ``"https://github.com/openedx/edx-platform.git"``)
+
+This defines the git repository from which you install Open edX platform code. If you run an Open edX fork with custom patches, set this to your own git repository. You may also override this configuration parameter at build time, by providing a ``--build-arg`` option.
+
+- ``OPENEDX_COMMON_VERSION`` (default: ``"open-release/maple.3"``)
```
fghaas, line 86: Took the liberty to add this update, which was apparently missed in #634.

fghaas:
I believe I have addressed all the pending comments now; please re-review.

regisb (approved):
thanks for your patience!

PR #622, opened 2022-03-29
fix: Make courseware discovery and search conditional on Elasticsearch

fghaas:
Courseware discovery and search requires Elasticsearch. If Elasticsearch is not enabled, then the search and discovery UI
breaks. Thus, make the relevant UI features inherit the boolean value of `RUN_ELASTICSEARCH`, disabling these features if Elasticsearch is not being deployed.


regisb:
Hi @fghaas,
I think that there's some misunderstanding. Setting `RUN_ELASTICSEARCH=false` does not cause Open edX to completely get rid of Elasticsearch and its dependencies. It means that Elasticsearch will not be run by Tutor, and that system administrators are expected to run by themselves, in a separate cluster. Elasticsearch is a mandatory component of Open edX and there is no way to get rid of it in Tutor core.

I understand that the docs are not clear at all about this. Please feel free to open another GitHub issue.

fghaas:
> Elasticsearch is a mandatory component of Open edX and there is no way to get rid of it in Tutor core.

Uh, no it's not? :) Open edX runs fine without Elasticsearch if you disable the corresponding features.

Plus, for Elasticsearch specifically, Tutor doesn't really support an external cluster, does it? Unlike for MySQL, it doesn't have a way to pass in authentication credentials for Elasticsearch. That's why I figured that the Tutor expectation is either run Elasticsearch from Tutor, or don't run it at all. No?

So, I think there should either be configuration parameters that can enable and disable these features separately (and their default should be the value of `RUN_ELASTICSEARCH`), or they should just inherit `RUN_ELASTICSEARCH` directly. Note that the obvious remedy, namely setting these features from a YAML plugin, doesn't work particularly well here because of [the known `patches` limitation](https://discuss.overhang.io/t/plugins-and-common-lms-env-features-shouldnt-the-patch-invocations-be-at-the-end-of-the-features-dictionary/2430).

regisb:
> Uh, no it's not? :) Open edX runs fine without Elasticsearch if you disable the corresponding features.

Well, let's put it this way: Open edX _can_ run without Elasticsearch if you absolutely know what you are doing. But Tutor makes the assumption in several places that there is an Elasticsearch cluster running at `ELASTICSEARCH_HOST`. And the goal of this `RUN_ELASTICSEARCH` flag is just to decide whether the elasticsearch container should be run.

> Plus, for Elasticsearch specifically, Tutor doesn't really support an external cluster, does it? Unlike for MySQL, it doesn't have a way to pass in authentication credentials for Elasticsearch. That's why I figured that the Tutor expectation is either run Elasticsearch from Tutor, or don't run it at all. No?

Good point! Personally I never ran MySQL or MongoDb outside of Tutor, but some users do and they contributed changes to the Tutor configuration. That's how we got the credential settings. If we don't have the same settings for Elasticsearch that's because no one bothered to propose a change. That being said, I suspect that it's possible to pass HTTP credentials via the host, in the `http://username:password@...` format.

So while it's possible to provide a flag to globally disable Elasticsearch, it would require changes made to plugins that make use of Elasticsearch. Among others, there are: notes, forum, discovery. These plugins would become basically useless without Elasticsearch, so I'm not ready to provide that flag just yet.

fghaas:
OK, fair enough, but what if a user does not want to run search, notes, forum, or discovery, and thus has no need for Elasticsearch? Then their basic LMS should still not break, which means they'll need a way to disable the search feature, and Tutor doesn't really have that option right now. In other words, that would be the first option of the two that I described here:

> So, I think there should either be configuration parameters that can enable and disable these features separately (and their default should be the value of `RUN_ELASTICSEARCH`), or they should just inherit `RUN_ELASTICSEARCH` directly.

Should I update this PR to that effect?

fghaas:
> That being said, I suspect that it's possible to pass HTTP credentials via the host, in the `http://username:password@...` format.

Did a bit of cursory research and I don't think that actually works. Open edX uses `ELASTIC_SEARCH_CONFIG` whose format is documented [here](https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch), and it uses `host` and `port` dictionary keys which by that same documentation are incompatible with RFC-1738 formatted URLs that can contain a username and password.

So it appears to me that what Tutor can support for the time being is run its own Elasticsearch, or rely on an external Elasticsearch with no authentication at all. (I don't know how realistic the latter is.)

regisb:
I'd rather not have a new Tutor configuration setting for every Open edX feature flag. So I'm not in favour of adding `ENABLE_COURSE_DISCOVERY`, `ENABLE_COURSEWARE_SEARCH`, `ENABLE_DASHBOARD_SEARCH`. Creating these settings opens up the possibility of too many conflicting values. What happens for instance when users manually set `RUN_ELASTICSEARCH=false` and `ENABLE_COURSE_DISCOVERY=true`?

The edx-platform course search settings can be adjusted by a separate plugin. At the moment, we consider that Elasticsearch is a necessary component of Open edX. `RUN_ELASTICSEARCH` is for deciding whether Elasticsearch should be run locally or not -- although I acknowledge that running a remote cluster is a bit tricky, at the moment.

fghaas:
> What happens for instance when users manually set `RUN_ELASTICSEARCH=false` and `ENABLE_COURSE_DISCOVERY=true`?

Then they see HTTP 500s when trying to search. _Which is exactly what happens now if they just set `RUN_ELASTICSEARCH=false` and do nothing else._ (In other words, this patch cannot make things worse for users, but it _can_ improve things.)

> At the moment, we consider that Elasticsearch is a necessary component of Open edX.

I don't quite follow who "we" is in this context, because Open edX upstream does not agree with that statement. At least that's what I glean from the fact that those three feature flags all (sensibly, in my opinion) default to `False` upstream ([source](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/feature_flags/feature_flag_index.html)).

> The edx-platform course search settings can be adjusted by a separate plugin.

No they cannot, at least not in a way that isn't extremely confusing. Tutor hard-codes these settings in a way that is not transparently overrideable, [as we have discussed before](https://discuss.overhang.io/t/plugins-and-common-lms-env-features-shouldnt-the-patch-invocations-be-at-the-end-of-the-features-dictionary/2430). Yes, you can override the settings by having one value in `lms.json.env` and the opposite value in `production.py` and if you're intimately familiar with configuration settings precedence in Tutor and Open edX you'll know that `production.py` beats `lms.json.env` — but none of that is simple, straightforward, or intuitive.

I don't understand why it makes sense for Tutor to set a hard-coded, non-overrideable default that is the direct opposite of the Open edX default. Tutor is a deployment method. It should follow the upstream defaults and offer facilities to deviate from them, rather than second-guess upstream and introduce its own contrarian defaults. But where it does do the latter, it should at the very least have a facility to undo that second-guessing, and that's what I am trying to add with this patch.

regisb:
Let me try to recap my position:

- `RUN_ELASTICSEARCH` means "run elasticsearch locally", just like `RUN_MYSQL`, `RUN_MONGODB` and others. It does 
not mean "enable search across the platform". I do not want to change the meaning of this setting now.
- I consider that Elasticsearch is a mandatory component of Open edX. I might very well be wrong; but this assumption was made in many different places, including plugins and courseware/dashboard search. If this assumption is wrong, then it should be possible to correct it, but it will require extensive work, including the introduction of a `ENABLE_SEARCH` setting. I am ready to look at PRs that implement this feature but I don't have time to implement the changes myself.
- The approach of Tutor has always been to provide good defaults for most people and to make it possible for developers to change these defaults via plugins. In many cases Tutor overrides the Open edX defaults because, well, they are not great.
- It is my opinion that courseware/dashboard search are useful features for most people, even in development. It might not work for you, in which case I want to make it easy for you to override this behaviour via a plugin. If the current implementation makes it too difficult, I can move the features flags from lms.env.json and cms.env.json to the Python settings.
- However, what I cannot offer you is to override that default feature via a single setting. That's because I am not prepared to introduce a Tutor setting for every feature flag in edx-platform (there are just too many of them).

If we need to discuss this further, can I suggest that we schedule a synchronous meeting?

fghaas:
> `RUN_ELASTICSEARCH` means "run elasticsearch locally", just like `RUN_MYSQL`, `RUN_MONGODB` and others. It does
not mean "enable search across the platform". I do not want to change the meaning of this setting now.

I'm afraid that's a [straw man](https://en.wikipedia.org/wiki/Straw_man). :) My patch as is doesn't change the meaning of this setting either. It just makes three values configurable that Tutor hard-codes (to the opposite of the upstream default), and _as the default in these newly configurable values_ it uses something that makes a little more sense in a Tutor context.

Would you like the default to instead be `true`, rather than whatever `RUN_ELASTICSEARCH` is set to? I can update the PR to that effect.

> However, what I cannot offer you is to override that default feature via a single setting. That's because I am not prepared to introduce a Tutor setting for every feature flag in edx-platform (there are just too many of them).

I'd argue that this is an example of a [slippery slope fallacy](https://en.wikipedia.org/wiki/Slippery_slope): introducing 3 configurable values is _not_ guaranteed to lead to this outcome. And it's _also_ a straw man, because I didn't suggest introducing a Tutor setting for every feature flag in edx-platform. :)

There are 14 `FEATURES` flags that for some reason are being set, hard-coded, in `lms.env.json`, with no simple, intuitive, and straightforward way to override them. We're not talking about "every feature flag in edx-platform", we're talking _at a maximum_ about these 14. Of these, this patch covers 3.

> The approach of Tutor has always been to provide good defaults for most people and to make it possible for developers to change these defaults via plugins. 

But in this case it's not possible to change the default via a plugin, because these values are not really defaults. Instead, they are hard coded feature flags whose settings cannot be overridden in a simple, intuitive, and straightforward manner, and all that I am trying to do is make them overrideable in a simple, intuitive, and straightforward manner. :) Is that really so misguided?

regisb:
Why would I be making purely rhetorical arguments? What's my ulterior motive?

fghaas:
I'm not arguing that you're making rhetorical arguments, much rather that you're making thinkos. :) As in, from my point of view you're considering things to be true that I believe aren't, or things to logically follow from other things that I believe do not.

And, I'd much rather see these sorts of things resolved in Tutor itself, rather than user whole community descending into a cascade of forks and local patches again, like what happened with edx-configuration. That's why I'm so adamantly arguing in favor of making hard-coded defaults configurable.

BbrSofiane:
Hey guys, 

I'm chiming in to make sure I understand what the conversation is about.

If I understand correctly what you are saying @fghaas, you want to be able to turn the discovery features on/off as easily as possible and this PR does it by making it possible to do it from the `config.yml` file. Is that correct?

My understanding of @regisb view is that `config.yml` should not maintain a certain level of abstraction. We should avoid having settings value in `config.yml` that are simply duplicates of feature flags that are in `production.py` or `lms.env.json`. Hence, his suggestion of adding `ENABLE_SEARCH` as opposed to adding 3 different settings. 

My approach has been to create a plugin dedicated to managing the configuration that is specific to the way we run our Open edX instance (i.e. enable or disable specific features). So I don't have multiple plugins for multiple features but one that captures how we run our instances. Would this work for your use case? I only have one instance to run so this works for me but I can see how it can be limiting for managing multiple custom instances.

fghaas:
> If I understand correctly what you are saying @fghaas, you want to be able to turn the discovery features on/off as easily as possible and this PR does it by making it possible to do it from the `config.yml` file. Is that correct?

Yes.

> My understanding of @regisb view is that `config.yml` should not maintain a certain level of abstraction. [...] Hence, his suggestion of adding `ENABLE_SEARCH` as opposed to adding 3 different settings.

To be honest, introducing a Tutor-specific feature flag that toggles 3 Open edX feature flags at once *does* strike me as an additional abstraction, no matter how it's named — so this critique applies either way. :) And I still maintain that in its current incarnation, it is not really practical to run an Elasticsearch that Tutor is meant use, outside of Tutor. So `ENABLE_SEARCH` would always need to be set to the same value as `RUN_ELASTICSEARCH`, or at least default to it. 

> My approach has been to create a plugin dedicated to managing the configuration that is specific to the way we run our Open edX instance (i.e. enable or disable specific features). So I don't have multiple plugins for multiple features but one that captures how we run our instance.
Would this work for your use case? I only have one instance to run so this works for me but I can see how it can be limiting for managing multiple custom instances.

Yes that's one way to do it (basically having one plugin that manages all feature flags for a single instance), but (a) we find that to be pretty unwieldy once you want to support a bunch of different feature flags that you need on and off for different platforms, and (b) with YAML plugins no longer existing in plugins v1, I think that approach creates lots of issues in the v0 to v1 plugin API migration. So yes we've looked that that approach but no, we don't find it particularly practical.

regisb:
As discussed earlier, I am not ready to add three extra settings to the core Tutor core configuration. Anyone who wants to disable courseware discovery/search based on the value of `RUN_DISCOVERY` can use the following plugin:

```
from tutor import hooks

    (
        "openedx-lms-common-settings",
        """
FEATURES["ENABLE_COURSE_DISCOVERY"] = {{ RUN_DISCOVERY }}
FEATURES["ENABLE_COURSEWARE_SEARCH"] = {{ RUN_DISCOVERY }}
FEATURES["ENABLE_DASHBOARD_SEARCH"] = {{ RUN_DISCOVERY }}
"""
    )
)
```

Issue #43, opened 2018-07-10
Any idea how to delete courses?


DerekCrosson:
This is probably a silly question... How do I delete the demo course and my test courses?

regisb:
This is not a silly question, but it's a standard Open edX operation :-) https://openedx.atlassian.net/wiki/spaces/PLAT/pages/31687733/How+to+Delete+a+Course

PR #627, opened 2022-04-06
feat: introduce 'tutor dev quickstart' (also, some extra docs)

kdmccormick:
## Description

Fixes https://github.com/overhangio/2u-tutor-adoption/issues/58

There are three commits. The first two just add some supplementary doc updates, which I could remove from this changeset if they don't seem like a good idea. The third commit introduces `tutor dev quickstart` along with the most critical documentation updates.

## Implementation Notes

I wanted to tailor `tutor dev quickstart` to the development use case, so there are some differences between it and `tutor local quickstart`:
1. it starts dev containers instead of local ones (duh),
2. it builds the openedx-dev image, and
3. it assumes No for the the "Is this a production platform?" question,
4. it offers to import the demo course and create a supseruser.

Difference **1** is the only thing that _really_ needed to be changed, so if there are objections to **2-4** do let me know.

In order to implement differences **3-4** I needed some way of passing to `tutor config save` that fact that we're in a developer context. Instead of cluttering `tutor config save` with a new command-line option, I opted to expose an `is_dev()` method on the `Context` class, which is used to tell whether developer-mode assumptions should be made. Let me know if this seems like a bad idea.

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -5,37 +5,61 @@ Open edX development
 
 In addition to running Open edX in production, Tutor can be used for local development of Open edX. This means that it is possible to hack on Open edX without setting up a Virtual Machine. Essentially, this replaces the devstack provided by edX.
 
-The following commands assume you have previously launched a :ref:`local <local>` Open edX platform. If you have not done so already, you should run::
 
-    tutor local quickstart
+First-time setup
+----------------
 
-To run the platform in development mode, you **must** answer no ("n") to the question "Are you configuring a production platform".
+First, ensure you have already `installed Tutor <install.rst>`_ (for development against the named releases of Open edX) or `Tutor Nightly <nightly.rst>`_ (for development against Open edX's master branches).
 
-Note that the local.overhang.io `domain <https://dnschecker.org/#A/local.overhang.io>`__ and its `subdomains <https://dnschecker.org/#CNAME/studio.local.overhang.io>`__ all point to 127.0.0.1. This is just a domain name that was set up to conveniently access a locally running Open edX platform.
+Then, launch the developer platform setup process::
 
-Once the local platform has been configured, you should stop it so that it does not interfere with the development environment::
+    tutor dev quickstart
 
-    tutor local stop
+This will perform several tasks for you. It will:
 
-Finally, you should build the ``openedx-dev`` docker image::
+* stop any existing locally-running Tutor containers,
 
-    tutor dev dc build lms
+* disable HTTPS,
 
-This ``openedx-dev`` development image differs from the ``openedx`` production image:
+* set your ``LMS_HOST`` to `local.overhang.io <http://local.overhang.io>`_ (a convenience domain that simply `points at 127.0.0.1 <https://dnschecker.org/#A/local.overhang.io>`_),
 
-- The user that runs inside the container has the same UID as the user on the host, to avoid permission problems inside mounted volumes (and in particular in the edx-platform repository).
-- Additional python and system requirements are installed for convenient debugging: `ipython <https://ipython.org/>`__, `ipdb <https://pypi.org/project/ipdb/>`__, vim, telnet.
-- The edx-platform `development requirements <https://github.com/openedx/edx-platform/blob/open-release/maple.master/requirements/edx/development.in>`__ are installed.
+* prompt for a platform details (with suitable defaults),
 
-Since the ``openedx-dev`` is based upon the ``openedx`` docker image, it should be re-built every time the ``openedx`` docker image is modified.
+* build an ``openedx-dev`` image, which is based ``openedx`` production image but is `specialized for developer usage`_,
 
-Run a local development webserver
----------------------------------
+* start LMS, CMS, supporting services, and any plugged-in services,
 
-::
+* ensure databases are created and migrated, and
 
-    tutor dev runserver lms # Access the lms at http://local.overhang.io:8000
-    tutor dev runserver cms # Access the cms at http://studio.local.overhang.io:8001
+* run service initialization scripts, such as service user creation and Waffle configuration.
+
+Once setup is complete, the platform will be running in the background:
+
+* LMS will be accessible at `http://local.overhang.io:8000 <http://local.overhang.io:8000>`_.
+* CMS will be accessible at `http://studio.local.overhang.io:8001 <http://studio.local.overhang.io:8001>`_.
+* Plugged-in services should be accessible at their documented URLs.
+
+
+Stopping the platform
+---------------------
+
+To bring down your platform's containers, simply run::
+
+  tutor dev stop
+
+
+Starting the platform back up
+-----------------------------
+
+Once you have used ``quickstart`` once, you can start the platform in the future with the lighter-weight ``start`` command, which brings up containers but does not perform any initialization tasks::
+
+  tutor dev start     # Run platform in the same terminal ("attached")
+  tutor dev start -d  # Or, run platform the in the background ("detached")
+
+Nonetheless, ``quickstart`` is idempotent, so it is always safe to run it again in the future without risk to your data. In fact, you may find it useful to use this command as a one-stop-shop for pulling images, running migrations, initializing new plugins you have enabled, and/or executing any new initialization steps that may have been introduced since you set up Tutor::
+
+  tutor dev quickstart --pullimages
+
 
 Running arbitrary commands
 --------------------------
@@ -55,6 +79,26 @@ You can then import edx-platform and django modules and execute python code.
 To collect assets, you can use the ``openedx-assets`` command that ships with Tutor::
 
     tutor dev run lms openedx-assets build --env=dev
+
+
+.. _specialized for developer usage: 
+
+Rebuilding the openedx-dev image
+--------------------------------
+
+The ``openedx-dev`` Docker image is based on the same ``openedx`` image used by ``tutor local ...`` to run LMS and CMS. However, it has a few differences to make it more convenient for developers:
+
+- The user that runs inside the container has the same UID as the user on the host, to avoid permission problems inside mounted volumes (and in particular in the edx-platform repository).
+
+- Additional Python and system requirements are installed for convenient debugging: `ipython <https://ipython.org/>`__, `ipdb <https://pypi.org/project/ipdb/>`__, vim, telnet.
+
+- The edx-platform `development requirements <https://github.com/openedx/edx-platform/blob/open-release/maple.master/requirements/edx/development.in>`__ are installed.
+
+
+If you are using a custom ``openedx`` image, then you will need to rebuild ``openedx-dev`` every time you modify ``openedx``. To so, run::
+
+    tutor dev dc build lms
+
 
 .. _bind_mounts:
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -36,8 +36,8 @@ To learn more about Tutor, watch this 7-minute lightning talk that was made at t
 
 .. youtube:: Oqc7c-3qFc4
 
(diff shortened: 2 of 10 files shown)


README.rst
```diff
@@ -63,10 +63,21 @@ Features
 Quickstart
 ----------
 
+For Open edX deployers
+~~~~~~~~~~~~~~~~~~~~~~
+
```
regisb, line 68: I'm not a big fan of making the quickstart instructions more complex. They are purposely very short; that's because there are many people who don't care about running Tutor in dev mode. So I suggest getting rid of this commit.

docs/dev.rst
```diff
@@ -5,37 +5,70 @@ Open edX development
...
-    tutor dev dc build lms
+* stop any existing locally-running Tutor containers,
 
-This ``openedx-dev`` development image differs from the ``openedx`` production image:
+* disable HTTPS,
 
-- The user that runs inside the container has the same UID as the user on the host, to avoid permission problems inside mounted volumes (and in particular in the edx-platform repository).
-- Additional python and system requirements are installed for convenient debugging: `ipython <https://ipython.org/>`__, `ipdb <https://pypi.org/project/ipdb/>`__, vim, telnet.
-- The edx-platform `development requirements <https://github.com/openedx/edx-platform/blob/open-release/maple.master/requirements/edx/development.in>`__ are installed.
+* set your ``LMS_HOST`` to `local.overhang.io`_ (a conveninece domain that simply `points at 127.0.0.1`_),
```
regisb, line 26: "convenience"

docs/dev.rst
```diff
@@ -5,37 +5,70 @@ Open edX development
...
+
+* prompt for a platform details (with suitable defaults),
+
+* offer to import the demo course,
```
regisb, line 30: I'd rather not do that -- and also not the superuser creation below. Provisioning should be done manually by the end user.

docs/dev.rst
```diff
@@ -5,37 +5,70 @@ Open edX development
...
+* run service initialization scripts, such as service user creation and Waffle configuration.
+
+Once setup is complete, the platform will be running in the background:
+
+* LMS will be accessible at `http://local.overhang.io:8000 <http://local.overhang.io:8000>`_.
+* CMS will be accessible at `http://studio.local.overhang.io:8001 <http://studio.local.overhang.io:8001>`_.
+* Plugged-in services should be accessible at their documented URLs.
+
+.. _local.overhang.io: http://local.overhang.io
+.. _points at 127.0.0.1: https://dnschecker.org/#A/local.overhang.io
```
regisb, line 49: Please use inline links:

```
`some text <http://example.com>`__
```

docs/dev.rst
```diff
@@ -56,6 +89,26 @@ To collect assets, you can use the ``openedx-assets`` command that ships with Tu
...
+.. _specialized for developer usage: 
+
+Rebuilding the openedx-dev image
+--------------------------------
+
+The ``openedx-dev`` Docker image is based on the same ``openedx`` image used by ``tutor local ...`` to run LMS and CMS. However, it has a few differences to make it more convenient for developers:
+
+- The user that runs inside the container has the same UID as the user on the host, to avoid permission problems inside mounted volumes (and in particular in the edx-platform repository).
+
+- Additional python and system requirements are installed for convenient debugging: `ipython <https://ipython.org/>`__, `ipdb <https://pypi.org/project/ipdb/>`__, vim, telnet.
```
regisb, line 102: "Python"

docs/intro.rst
```diff
@@ -101,6 +101,34 @@ You can now take advantage of the Tutor-powered CLI (item #3) to bootstrap your
 
 Under the hood, Tutor simply runs ``docker-compose`` and ``docker`` commands to launch your platform. These commands are printed in the standard output, such that you are free to replicate the same behaviour by simply copying/pasting the same commands.
 
+How do I navigate Tutor's command-line interface?
+-------------------------------------------------
```
regisb, line 105: I guess this section should include a link to the CLI reference docs, somewhere.

tutor/commands/config.py
```diff
@@ -48,7 +48,9 @@ def save(
     unset_vars: List[str],
     env_only: bool,
 ) -> None:
-    config = interactive_config.load_user_config(context.root, interactive=interactive)
+    config = interactive_config.load_user_config(
+        context.root, interactive=interactive, dev_context=context.is_dev()
+    )
```
regisb, line 53: I'd rather avoid adding new methods to the context objects. Instead, I suggest to get rid of the `interactive.load_user_config` function and to add a `run_for_prod=None` argument to `interactive.ask_questions`. 

```
def ask_questions(config: Config, run_for_prod=None):
    if run_for_prod is None:
        # ask "Are you configuring a production platform?"
    ...
```

Each `quickstart` method will make a call to `ask_questions` instead of `load_user_config`:

```
def quickstart(...):
    config = tutor_config.load_minimal(root)
    if interactive:
        # run_for_prod will be:
        # None in local
        # False in dev
        # True in k8s
        interactive.ask_questions(config, run_for_prod=...)
```

regisb (changes requested):

docs/dev.rst
```diff
@@ -5,37 +5,70 @@ Open edX development
 
 In addition to running Open edX in production, Tutor can be used for local development of Open edX. This means that it is possible to hack on Open edX without setting up a Virtual Machine. Essentially, this replaces the devstack provided by edX.
 
-The following commands assume you have previously launched a :ref:`local <local>` Open edX platform. If you have not done so already, you should run::
+.. warning::
 
-    tutor local quickstart
+  Do not run ``tutor dev ...`` commands on a production system. They can modify your configuration and database in ways that are nonsensical or insecure for production use.
```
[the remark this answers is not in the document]
kdmccormick: @regisb I felt the need to add this warning because `tutor dev quickstart` would do things you'd likely not want on a production instance:
1. import the demo course
2. create a superuser
3. automatically set `LMS_ROOT=local.overhang.io` and `ENABLE_HTTPS=false`

You've asked me to revert (1) and (2) so those aren't concerns any more.

Should I revert (3) as well? That is, should `tutor dev quickstart` prompt the user with "Is this a production instance? [y/n]" instead of assuming "no" like it does in my current changeset?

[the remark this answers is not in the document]
regisb: (3) is great, let's keep it! It makes configuration much simpler and less error-prone. But doing (3) will not dramatically affect the platform, changing HTTPS/host settings is a routine task for many people. So we can get rid of the warning.

docs/dev.rst
```diff
@@ -5,37 +5,70 @@ Open edX development
...
+* Plugged-in services should be accessible at their documented URLs.
+
+.. _local.overhang.io: http://local.overhang.io
+.. _points at 127.0.0.1: https://dnschecker.org/#A/local.overhang.io
+
+
+Stopping the platform
+---------------------
+
+To bring down your platrform's containers, simply run::
```
regisb, line 55: nit: "platform"

bradenmacdonald:
@kdmccormick @regisb What do you think about `tutor dev quickstart` offering an interactive option to use (or even create) a copy of `edx-platform` on the user's host computer? e.g. the old devstack used to clone edx-platform for you, and I think that for Open edX developers, having an editable version of the platform is important. (I would actually guess that most people who want to use tutor nightly are also wanting to have an editable/dev version of edx-platform).

However, currently the instructions for doing so are [complicated](https://docs.tutor.overhang.io/dev.html#sharing-directories-with-containers) (and not referenced on the [Nightly page](https://docs.tutor.overhang.io/tutorials/nightly.html)). So I was thinking something like a prompt which says: "Do you have a local checkout of the edx-platform source (master branch) that you want to use with Tutor Nightly? If so, enter the full path to it [e.g. /home/regis/dev/edx-platform] or press enter to skip:"

And then if you enter a path, it can [create the `docker-compose.override.yml`](https://docs.tutor.overhang.io/dev.html#override-docker-compose-volumes) so that everything "just works".

README.rst
kdmccormick: That makes sense, I'll remove this.

Still, I think the readme should have some link to point Open edX developers in the right direction. Currently, the [Open edX development page](https://docs.tutor.overhang.io/dev.html) is 3+ clicks away. How would you feel about linking right to that page, either in the first sentence ("...both for production and **local development**.") or via a new bullet point in the Features list?

docs/dev.rst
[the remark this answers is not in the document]
kdmccormick: Sounds good!

docs/dev.rst
kdmccormick: OK 👍🏻

tutor/commands/config.py
kdmccormick: Much cleaner.

In the existing implementation, quickstart's call to `config.save` causes these lines to get executed:
```python
if not env_only:
    tutor_config.save_config_file(context.root, config)
# Reload configuration, without version checking
config = tutor_config.load_full(context.root)
env.save(context.root, config)
```
Since we're not calling `config.save` any more, should I also copy the equivalent of those lines in to `local.quickstart` and `dev.quickstart`?

kdmccormick:
@bradenmacdonald That's a really interesting idea. I agree with your point that mounting code locally is currently too complicated. We have [work in flight](https://github.com/overhangio/2u-tutor-adoption/issues/43) to address it, with a proposal syntax like:
```
$ tutor dev start -d -m /path/to/my/edx-platform -m /path/to/another/ida
```

So, I am torn on your suggestion. On one hand, it would allow developers to "set and forget" repository mounts, which would certainly lower the start-up effort for new users. On the other hand, running services from built-in container code by-default is one of the reasons Tutor is less resource-intensive and (IMO) more reliable than Devstack (which host-mounts code for every single service), so encouraging developers to explicitly mount in host directories as-needed might be better. @regisb , what do you think?

docs/intro.rst
kdmccormick: Added

tutor/commands/config.py
kdmccormick: Update: I did this. This block appears in each `quickstart`:
```python
    click.echo(fmt.title("Interactive platform configuration"))
    if not non_interactive:
        interactive_config.ask_questions(config, run_for_prod=...)
    tutor_config.save_config_file(context.obj.root, config)
    config = tutor_config.load_full(context.obj.root)
```
Kinda feels like it should be factored out.

bradenmacdonald:
@kdmccormick I see, that does make a lot of sense. Sorry I missed the existing discussion; we can follow up in #43.

regisb (approved):
Can I just say that I love all the changes that we are making?

tutor/commands/dev.py
```diff
@@ -41,6 +43,58 @@ def dev(context: click.Context) -> None:
...
+
+    click.echo(fmt.title("Stopping any existing platform"))
+    context.invoke(compose.stop)
+
+    if pullimages:
+        click.echo(fmt.title("Docker image updates"))
+        context.invoke(compose.dc_command, command="pull")
+
+    click.echo(fmt.title("Building Docker image for LMS and CMS development"))
+    context.invoke(compose.dc_command, command="build", args=["lms"])
```
regisb, line 77: @kdmccormick I'd like to come back to this PR. It seems to me that it's unnecessary to build the lms image in a separate command. The image should be re-built in the next start command anyway, because we launch a `docker compose up --build` command.
Am I missing something? If not, I'd like to remove this line, because it's causing unnecessary cache checks.

kdmccormick: @regisb When I added this, I don't think I realized that `dev start` used the compose `--build` flag. So, yes, rip it out!

PR #635, opened 2022-04-11
wip: upgrade to Nutmeg

regisb:


gabor-boros:
Hey @regisb what's left to merge this PR? 
cc: @keithgg @mtyaka

regisb:
We will only merge this PR once Nutmeg is released. Otherwise all users who run the Tutor master branch will be migrated to Nutmeg...

gabor-boros:
I see that part. My question was mostly about literally what's left. 

For example, tutor has no deployments defined for celerybeat while the openedx configuration repo has and available for nutmeg.

regisb:
Right. As far as I know, this PR has all the necessary changes for Nutmeg. I might be wrong though, and I'll be happy to pull in required changes, for instance those that are identified as necessary while running the [test plan](https://docs.google.com/spreadsheets/d/1A3LYVp9ivOLL1pj7eYeyel9zcxHM8BdQ_gVRpzMJqbU/edit?pli=1#gid=536490455).

I was not aware that Celerybeat was required in Nutmeg. What is it used for?

gabor-boros:
Thank you for the reply.

Not required for nutmeg, but optional. That can be setup as a plug-in too, just wanted to see if you plan to add deployment for that or not ☺️

regisb:
I'd rather not add an extra container using as much memory as the LMS, that would be too much for most users -- and I'm not sure what would be the benefit.

gabor-boros:
@regisb although I don't want to go down the rabbit hole: Why would it have to use the same amount of resources?

regisb:
According to `docker stats` the `lms-worker` container uses more memory than the `lms` in idle mode:

```
8ea3fe82d755   tutor_local_lms_1             0.08%     293.4MiB / 19.48GiB   1.47%     11.8kB / 3.21kB   77.6MB / 0B      10
31dc4af9b47d   tutor_local_lms-worker_1      0.58%     656.4MiB / 19.48GiB   3.29%     35.3kB / 32.7kB   78.4MB / 0B      5
```

I wouldn't know precisely why the worker uses more memory... I suspect it's because the lms-worker needs to import parts of the cms code. (most of the memory comes from imports, as far as I can remember)

I would expect that any extra beat worker would use the same amount of memory as the lms-worker.

PR #637, opened 2022-04-12
feat: run `tutor k8s exec ...` commands with multiple args

### LINKED ISSUE #636: Error while executing commands inside service pods as per the documentation
**Bug description**
The tutor docs [recommends](https://github.com/overhangio/tutor/blame/master/docs/k8s.rst#L115-L122) the following command to run a python shell in lms container:
`tutor k8s exec lms ./manage.py lms shell`

However there are two issues with running the command as is :

1. Tutor expects the `k8s exec` cli commands to be of the form `tutor k8s exec [OPTIONS] SERVICE COMMAND` and thus throws the error `Error: Got unexpected extra arguments (lms shell)`. To run the service commands we need to add them to quotes such as `tutor k8s exec lms "./manage.py lms shell"`. Perhaps the docs should be modified to indicate that.

3. By default the sevice commands runs using the development settings, and thus throws the below error since it cannot find the debug toolbar installed:
image

Perhaps something like [BASE_OPENEDX_COMMAND](https://github.com/overhangio/tutor/blob/baf7d03cf801a398a5aa48ea80af275df418e215/tutor/jobs.py#L6) could be used with `k8s exec` to set the correct settings.
Or, the docs could simply guide the user to set the correct settings in the form of:
`tutor k8s exec lms "DJANGO_SETTINGS_MODULE=lms.envs.tutor.production ./manage.py lms shell"`

**How to reproduce**
1. Spin up a tutor k8s deployment with `tutor k8s quickstart`
2. Run `tutor k8s exec lms ./manage.py lms shell` to try to get to the lms shell.

**Environment**
Error encountered using tutor version 13.1.8

**Additional context**

regisb:
Previously, the `k8s exec` command did not support unknown "--options". This
made it impossible to launch, say, a django shell in the lms container.

While implementing this feature we saw an opportunity to simplify the way jobs
are handled in the k8s commands.

Close #636.
Another related issue is: https://github.com/overhangio/2u-tutor-adoption/issues/52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Note: Breaking changes between versions are indicated by "💥".
 
 ## Unreleased
 
+- 💥[Improvement] Make it possible to run `tutor k8s exec <command with multiple arguments>` (#636). As a consequence, it is no longer possible to run quoted commands: `tutor k8s exec "<some command>"`. Instead, you should remove the quotes: `tutor k8s exec <some command>`.
+
 ## v13.1.11 (2022-04-12)
 
 - [Security] Apply SAML security fix.
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -8,7 +8,7 @@ from tutor import config as tutor_config
 from tutor import env as tutor_env
 from tutor import exceptions, fmt, jobs, serialize, utils
 from tutor.commands.config import save as config_save_command
-from tutor.commands.context import Context
+from tutor.commands.context import BaseJobContext
 from tutor.commands.upgrade.k8s import upgrade_from
 from tutor.types import Config, get_typed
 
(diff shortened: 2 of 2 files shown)


regisb:
this is ready for review @overhangio/tutor-developers. cc @kdmccormick.

kdmccormick (approved):
LGTM, with a couple notes:

* The workaround to this issue that I had been using was to to wrap the command and its args in quotes, e.g. `tutor k8s exec "ls -l"`. This PR "breaks" that workaround (which is totally fine with me), but consider marking it a breaking change in the changelog. Up to you.
* [#636](https://github.com/overhangio/tutor/issues/636) raises two issues. This PR solves one of them, but the second issue (`DJANGO_DEBUG_TOOLBAR` not being set) still exists, and is a problem for `tutor [dev|local] exec` as well. I think we should open a new issue to track it. I had proposed a potential solution [in this comment](https://github.com/overhangio/2u-tutor-adoption/issues/26#issuecomment-1073440295).

regisb:
Good points @kdmccormick. I modified the changelog to indicate a breaking change.

Indeed, this in itself does not resolve the `DJANGO_DEBUG_TOOLBAR` error raised in #636. I'll comment there.

PR #639, opened 2022-04-13
fix: Correct EDX_PLATFORM_VERSION default for local edx-platform forks

fghaas:
PR #619 set the `EDX_PLATFORM_VERSION` build arg's default to `OPENEDX_COMMON_VERSION`. While this works fine for setting a non-default branch to run edx code from (say, `master`), it may break if the user sets `OPENEDX_COMMON_VERSION` to a branch or tag name that does not exist upstream in repositories *other than* `EDX_PLATFORM_REPOSITORY`.

Thus, introduce a separate configuration parameter, `EDX_PLATFORM_VERSION`, to match the build arg of the same name. Set its default to `OPENEDX_COMMON_VERSION`.

This way, the user can deploy an arbitrarily-named fork of `edx-platform`, while retaining the default  `OPENEDX_COMMON_VERSION` (like, for example `open-release/maple.3`) for everything else.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -87,6 +87,10 @@ This defines the git repository from which you install Open edX platform code. I
 
 This defines the default version that will be pulled from all Open edX git repositories.
 
+- ``EDX_PLATFORM_VERSION`` (default: the value of ``OPENEDX_COMMON_VERSION``)
+
+This defines the version that will be pulled from just the Open edX platform git repositories. You may also override this configuration parameter at build time, by providing a ``--build-arg`` option.
+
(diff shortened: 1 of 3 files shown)


fghaas:
@regisb Sorry about this thinko. I don't think this needs an urgent release — it would only affect people that have previously overridden `OPENEDX_COMMON_VERSION`, and only in a way that I think would not have worked before #619 either, so this possibility strikes me as merely theoretical. At any rate though, this should fix things.

I'll amend this patch shortly with a doc update.

regisb (approved):

PR #643, opened 2022-04-19
feat: add --mount option to local/dev

regisb:
The `--mount` and `--automount` options are available both with `tutor local`
and `tutor dev` commands. They allow users to easily bind-mount containers from
the host to containers. Yes, I know, we already provide that possibility with
the `bindmount` command and the `--volume=/path/` option. But these suffer from
the following drawbacks:

- They are difficult to understand.
- The "bindmount" command name does not make much sense.
- It's not convenient to mount an arbitrary folder from the host to multiple
  containers, such as the many lms/cms containers (web apps, celery workers and
  job runners).

To address this situation, we now recommend to make use of --mount and
--automount options.

1. `--mount=service1[,service2,...]:/host/path:/container/path`: manually mount
   `/host/path` to `/container/path` in container "service1" (and "service2").
2. `--automount=/host/path`: use the new v1 plugin API to discover plugins that
   will detect this option and select the right containers in which to bind-mount
   volumes. This is really nifty...

Close https://github.com/overhangio/2u-tutor-adoption/issues/43


regisb:
cc @kdmccormick

docs/dev.rst
```diff
@@ -156,37 +192,31 @@ You are then free to bind-mount any directory to any container. For instance, to
...
-    tutor dev bindmount lms /openedx/edx-platform
-    tutor dev runserver --volume=/openedx/edx-platform lms
-
-    # Mount from an arbitrary directory
-    tutor dev runserver --volume=/path/to/edx-platform:/openedx/edx-platform lms
+Following the instructions :ref:`above <bind_mounts>` on how to bind-mount directories from the host above, you may mount your own `edx-platform <https://github.com/openedx/edx-platform/>`__ fork in your containers by running::
 
-    # Add your own volumes to $(tutor config printroot)/env/dev/docker-compose.override.yml
-    tutor dev runserver lms
+    tutor dev runserver --mount=/path/to/edx-platform lms
```
kdmccormick, line 210: Since the linked bind-mount instructions use `start`, let's use it here too.
```suggestion
    tutor dev start -d --mount=/path/to/edx-platform lms
```

docs/dev.rst
```diff
@@ -156,37 +192,31 @@ You are then free to bind-mount any directory to any container. For instance, to
...
 
-If you choose any but the first solution above, you will have to make sure that your fork works with Tutor.
+But to achieve that, you will have to make sure that your fork works with Tutor.
 
-First of all, you should make sure that you are working off the ``open-release/maple.2`` tag. See the :ref:`fork edx-platform section <edx_platform_fork>` for more information.
+First of all, you should make sure that you are working off the ``open-release/maple.3`` tag. See the :ref:`fork edx-platform section <edx_platform_fork>` for more information.
 
 Then, you should run the following commands::
```
kdmccormick, line 216: Too bad users still need to run four extra commands after mounting edx-platform. I don't know if we can address this before the conference, but it's on my mind as something we could improve.

Just thinking out loud (not a change request for this PR): If we upstreamed the openedx-assets script into edx-platform behind `make static`, and then made Makefile targets for Python and JS dev requirements, this could become:
```
tutor dev run -m /path/to/edx-platform make requirements-dev static
```

docs/dev.rst
```diff
@@ -107,9 +107,43 @@ Sharing directories with containers
...
+.. _mount_option:
+
+Bind-mount volumes with ``--mount``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``run``, ``runserver``, ``init`` and ``start`` subcommands of ``tutor dev`` and ``tutor local`` support the ``-m/--mount`` option (see :option:`tutor dev start -m`) which can take two different forms. The first is explicit::
+
+    tutor dev start --mount=lms:/path/to/edx-platform:/openedx/edx-platform lms
+
+And the second is implicit::
```
kdmccormick, line 119: I like how you structured this whole section. Very informative and easy to follow.

tutor/commands/compose.py
```diff
@@ -60,23 +93,87 @@ def job_runner(self, config: Config) -> ComposeJobRunner:
         raise NotImplementedError
 
 
+class MountParam(click.ParamType):
+    """
+    Parser for --mount arguments of the form "service1[,service2,...]:/host/path:/container/path".
+    """
+
+    name = "mount"
+    MountType = t.Tuple[str, str, str]
```
kdmccormick, line 102: What about just `Mount`? I would expect a type called `MountType` to mean something like an enum of the different types of mounts.

kdmccormick (approved):
Looking good other than a few comments, and works great! Just make sure to update your commit message since `--mount/--automount` has been combined into just `--mount`.

I assume you'll handle the `tutor dev bindmount` replacement in another PR?

docs/dev.rst
```diff
@@ -197,9 +227,9 @@ Then, you should run the following commands::
     # Rebuild static assets
     openedx-assets build --env=dev
 
-To debug a local edx-platform repository, add a ``import ipdb; ipdb.set_trace()`` breakpoint anywhere in your code and run::
+After running all these commands, your edx-platform repository will be ready for local development. To debug a local edx-platform repository, you can then add a ``import ipdb; ipdb.set_trace()`` breakpoint anywhere in your code and run::
 
-    tutor dev runserver [--volume=...] lms
+    tutor dev runserver --mount=/path/to/edx-platform lms
```
kdmccormick, line 232: Hm, thinking forward to the `runserver` deprecation... without `runserver`, how would folks interact with pdb?

Do we need a new `attach` command?
```
tutor dev start -d --mount=/path/to/edx-platform lms
tutor dev attach lms
```

docs/dev.rst
regisb: What would be a good command name for preparing the edx-platform repo? I'm thinking of creating a `tutor dev tasks` subcommand under which we would move `createuser` and `importdemocourse`:

    tutor dev tasks importdemocourse

Of course, the list of tasks would be the result of a filter, making it possible for plugins to add custom commands. We could then write something like:

    tutor dev tasks -m /path/to/edx-platform prepare-edx-platform

regisb:
@overhangio/tutor-developers you may want to have a look at this, it's really cool :)

tutor/commands/compose.py
regisb: Oops I missed that comment, sorry about that...

I picked the "Type" suffix because it's not a class of which we want to create instances. But I'm not bent on it.

docs/dev.rst
kdmccormick: @regisb I like the `tasks` subcommand idea! I wonder if it would be elegant to bring `init` (and `pre-init`) within the `tasks` abstraction as well . What would you think about:
* Making `tutor MODE init` an alias to `tutor MODE tasks init`
* Renaming the filter "commands:init" to "tasks:init", and namespacing new tasks under "tasks:*"
* Supporting `tasks --limit`, with the same semantics as `init --limit`

Anyway, in terms of `prepare-edx-platform`, what if we just called the task something generic like `prepare-repo`? That way, the discovery plugin could hook in commands to prepare a mounted course-discovery repo , and so on for other plugins.

tutor/commands/compose.py
kdmccormick: No worries. I'm not bent on it either.

docs/dev.rst
regisb: I'm not sure whether replacing the init commands by a single task would work -- at least with what I had in mind for the `tasks` subcommand. In my understanding, a "task" would be a command to be run in a single container. More importantly, subsequent command line arguments passed to `tutor local tasks ...` would be passed along to the script. For instance, this would allow the `createuser` script to support `--superuser --staff --password=...` arguments. We could have a single task trigger multiple commands in different containers (such as `init`) but I'm not sure whether this approach would play well with command line arguments.

kdmccormick: Okay, got it, I was imaging something more involved. I misread:
> the list of tasks would be the result of a filter, making it possible for plugins to add custom commands.

as something like:
> the list of commands run by a task would be the result of a filter, making it possible for plugins to add custom commands [to tasks]

What advantage would `tutor dev tasks` have over just putting scripts in the /openedx/bin folder (like `openedx-assets`), and then invoking them with `tutor dev run`? Is it that plugins can't currently add to /openedx/bin for existing images?

regisb: > Is it that plugins can't currently add to /openedx/bin for existing images?

Yes, exactly. Adding stuff there requires rebuilding the image, which is inconvenient. We could also bind-mount a script folder, but this is inconvenient on Kubernetes.

kdmccormick: Got it. I'm in favor, then!

In the long term, I'd prefer if we could upstream any generally-useful scripts into the app repos instead of writing them into plugins as tasks, but I know that's often easier said than done...

PR #645, opened 2022-04-21
feat: add `dev/local copyfrom` commands

regisb:
`copyfrom` copies data from a container to the local filesystem. It's similar
to bindmount, but less clunky, and more intuitive. Also, it plays along great
with `--mount`. Eventually we'll just get rid of the `bindmount` command and
the `--volume` option.

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -139,10 +139,23 @@ So, when should you *not* be using the implicit form? That would be when Tutor d
 
 .. note:: Remember to setup your edx-platform repository for development! See :ref:`edx_platform_dev_env`.
 
+Copy files from containers to the local filesystem
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, you may want to modify some of the files inside a container for which you don't have a copy on the host. A typical example is when you want to troubleshoot a Python dependency that is installed inside the application virtual environment. In such cases, you want to first copy the contents of the virtual environment from the container to the local filesystem. To that end, Tutor provides the ``tutor dev copyfrom`` command. First, copy the contents of the container folder to the local filesystem::
+
+    tutor dev copyfrom lms /openedx/venv ~
+
+Then, bind-mount that folder back in the container with the ``--mount`` option (described :ref:`above <mount_option>`)::
+
+    tutor dev start --mount lms:~/venv:/openedx/venv lms
+
+You can then edit the files in ``~/venv`` on your local filesystem and see the changes live in your container.
+
 Bind-mount from the "volumes/" directory
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. warning:: Bind-mounting volumes with the ``bindmount`` command is no longer the default, recommended way of bind-mounting volumes from the host. Instead, see the :ref:`mount option <mount_option>`.
+.. warning:: Bind-mounting volumes with the ``bindmount`` command is no longer the default, recommended way of bind-mounting volumes from the host. Instead, see the :ref:`mount option <mount_option>` and the ``tutor dev/local copyfrom`` commands.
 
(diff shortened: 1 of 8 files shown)


regisb:
This is ready for review @overhangio/tutor-developers. cc @kdmccormick.

docs/dev.rst
```diff
@@ -139,10 +139,23 @@ So, when should you *not* be using the implicit form? That would be when Tutor d
...
+
+    tutor dev start --mount lms:~/venv:/openedx/venv lms
+
+You can then edit the files in ``~/venv`` on your local filesystem and see the changes live in your container.
+
 Bind-mount from the "volumes/" directory
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. warning:: Bind-mounting volumes with the ``bindmount`` command is no longer the default, recommended way of bind-mounting volumes from the host. Instead, see the :ref:`mount option <mount_option>`.
+.. warning:: Bind-mounting volumes with the ``bindmount`` command is no longer the default, recommended way of bind-mounting volumes from the host. Instead, see the :ref:`mount option <mount_option>` and the ``tutor dev/local copyfrom`` commands.
```
kdmccormick, line 158: Consider making the `bindmount` command print a similar warning.

docs/dev.rst
```diff
@@ -139,10 +139,23 @@ So, when should you *not* be using the implicit form? That would be when Tutor d
...
+
+Sometimes, you may want to modify some of the files inside a container for which you don't have a copy on the host. A typical example is when you want to troubleshoot a Python dependency that is installed inside the application virtual environment. In such cases, you want to first copy the contents of the virtual environment from the container to the local filesystem. To that end, Tutor provides the ``tutor dev copyfrom`` command. First, copy the contents of the container folder to the local filesystem::
+
+    tutor dev copyfrom lms /openedx/venv ~
+
+Then, bind-mount that folder back in the container with the ``--mount`` option (described :ref:`above <mount_option>`)::
+
+    tutor dev start --mount lms:~/venv:/openedx/venv lms
+
+You can then edit the files in ``~/venv`` on your local filesystem and see the changes live in your container.
```
kdmccormick, line 153: This is awesome!

tutor/commands/compose.py
```diff
@@ -343,6 +341,39 @@ def bindmount_command(context: BaseComposeContext, service: str, path: str) -> N
...
+@click.command(
+    name="copyfrom",
+    help="Copy files/folders from a container directory to the local filesystem.",
+)
+@click.argument("service")
+@click.argument("container_path")
+@click.argument(
+    "host_path",
+    type=click.Path(exists=True, dir_okay=True, file_okay=False, resolve_path=True),
+)
```
kdmccormick, line 353: In a posix shell, the semantics of `cp -r SRC DST` are:
1. if `$DST` exists, copy `$SRC` to `$DST/$(basename $SRC)`
2. else, copy `$SRC` directly to `$DST`

(aside, I had to look these rules up because they're so natural to me that I never actually thought about them :)

Your implementation does (1), but folks used to posix's `cp` might expect/want (2) to also work:
```
$ tutor dev copyfrom lms /openedx/venv edx-platform-venv
Usage: tutor dev copyfrom [OPTIONS] SERVICE CONTAINER_PATH HOST_PATH
Try 'tutor dev copyfrom --help' for help.

Error: Invalid value for 'HOST_PATH': Directory 'edx-platform-venv' does not exist.
```

If you'd rather leave it as-is, no worries. This works well enough:
```
$ tutor dev copyfrom lms /openedx/venv . && mv venv edx-platform-venv
```

kdmccormick (approved):
Nice! Two suggestions, non-blocking.

tutor/commands/compose.py
regisb: I pulled my hair a little but I found an acceptable solution to implement (2). Please see the revised implementation.

kdmccormick (approved):

tutor/commands/compose.py
kdmccormick: Looks great.

PR #646, opened 2022-04-21
feat: `local stop` on `dev start` (and vice versa)

regisb:
Running `local start` while a dev platform is still running is a common sourse
of mistakes. Here we introduce a new action to automatically stop local and dev
projects whenever a project with a different name is started.

### DIFF
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -24,6 +24,12 @@ class ComposeJobRunner(jobs.BaseComposeJobRunner):
         """
         Run docker-compose with the right yml files.
         """
+        if "start" in command or "up" in command or "restart" in command:
+            # Note that we don't trigger the action on "run". That's because we
+            # don't want to trigger the action for every initialization script.
+            hooks.Actions.COMPOSE_PROJECT_STARTED.do(
+                self.root, self.config, self.project_name
+            )
         self.__update_docker_compose_tmp()
         args = []
         for docker_compose_path in self.docker_compose_files:
--- a/tutor/commands/local.py
+++ b/tutor/commands/local.py
@@ -4,7 +4,7 @@ import click
 
 from tutor import config as tutor_config
 from tutor import env as tutor_env
-from tutor import exceptions, fmt
+from tutor import exceptions, fmt, hooks
 from tutor import interactive as interactive_config
 from tutor import utils
 from tutor.commands import compose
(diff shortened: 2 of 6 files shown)


tutor/commands/compose.py
```diff
@@ -7,12 +7,10 @@
 from tutor import bindmounts
 from tutor import config as tutor_config
 from tutor import env as tutor_env
-from tutor import fmt, jobs, utils
-from tutor import serialize
+from tutor import fmt, hooks, jobs, serialize, utils
```
regisb, line 10: Import order changes are the result of running `make isort` and you can safely ignore them.

kdmccormick:
Resolves https://github.com/overhangio/2u-tutor-adoption/issues/27

tutor/commands/local.py
```diff
@@ -165,6 +166,17 @@ def upgrade(context: click.Context, from_release: t.Optional[str]) -> None:
     context.invoke(config_save_command)
 
 
+@hooks.Actions.COMPOSE_PROJECT_STARTED.add()
+def _stop_on_dev_start(root: str, config: Config, project_name: str) -> None:
+    """
+    Stop the dev platform as soon as a platform with a different project name is
```
kdmccormick, line 172:
```suggestion
    Stop the local platform as soon as a platform with a different project name is
```

kdmccormick (approved):
Love this, works like a charm. I think plugin devs will find COMPOSE_PROJECT_STARTED to be a really useful hook as well.

Thinking beyond this PR... I also encounter this same problem when switching between Tutor nightly and master. What would you think about stopping all Nightly platforms when starting a stable platform, and vice versa?

regisb:
> Thinking beyond this PR... I also encounter this same problem when switching between Tutor nightly and master. What would you think about stopping all Nightly platforms when starting a stable platform, and vice versa?

I really don't know how to implement this in a clean way... The problem is that tutor "non-nightly" has no concept of "nightly". Also,  the tutor `__app__` can be overridden by environment variables, such that making sure that stopping any non-nightly platform is a slightly chaotic process...
*In theory*, the nightly platform can run just fine next to a local platform. They run in different project roots, don't share any data and don't have the same project name. The only source of conflict would be the exposed web proxy ports (80, 443).
My problem is that I don't know how to pass action callbacks from one installation of tutor to the next.

regisb:
Also, note that I modified the criterion for triggering the `COMPOSE_PROJECT_STARTED` action: it is no longer triggered on "run". That's because I don't want to call `docker-compose stop` for every init script.

tutor/hooks/consts.py
```diff
@@ -25,6 +25,13 @@ def your_action():
             # Do stuff here
     """
 
+    #: Triggered whenever a docker-compose start/run/restart command is executed.
```
kdmccormick, line 28:
```suggestion
    #: Triggered whenever a docker-compose start or restart command is executed.
```

kdmccormick (approved):
> Also, the tutor `__app__` can be overridden by environment variables

Ah, I was thinking you could hard-code a list of the possible app names `["tutor-nightly", "tutor"]`. But if `__app__` be arbitrarily overridden, then nevermind :)

> Also, note that I modified the criterion for triggering the COMPOSE_PROJECT_STARTED action: it is no longer triggered on "run". That's because I don't want to call docker-compose stop for every init script.

Good call.

PR #647, opened 2022-04-21
fix: `--mount` should accept service names with hyphens

kdmccormick:
I didn't include a changelog entry, as this fixes a feature that hasn't yet been released.

Also included, in a second commit: a drive by fix to the `runserver` deprecation warning.


kdmccormick:
@regisb This is a quick fix to `--mount` when you have a chance to review.

regisb (approved):

PR #648, opened 2022-04-21
Ensure edx-platform tests can run without fiddling with environment variables

kdmccormick:
## Description

See the commit message for details.
Blocked by https://github.com/openedx/edx-platform/pull/30298
Resolves https://github.com/overhangio/2u-tutor-adoption/issues/48

## Considerations 

This PR (and the linked edx-platform PR) would succeed at getting pytest to default to `[lms|cms].envs.tests`. However, I haven't come up with an elegant way of getting pytest to default to `[lms,cms].envs.tutor.test`, which adds this change to the upstream test settings:
```python
# Fix MongoDb connection credentials
DOC_STORE_CONFIG["user"] = None
DOC_STORE_CONFIG["password"] = None
```
In trying to find out why we need to change DOC_STORE_CONFIG connection credentials, git blame leads me to https://github.com/overhangio/tutor/commit/8d803fb08b13e2e20249b82c034483e61ec22795. @regisb , do you remember why you why need to change these credentials? Do you think it's possible we could work around this some other way, allowing us to use the upstream test settings?

## Screenshot of changed docs section

image

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -81,7 +81,7 @@ To collect assets, you can use the ``openedx-assets`` command that ships with Tu
     tutor dev run lms openedx-assets build --env=dev
 
 
-.. _specialized for developer usage: 
+.. _specialized for developer usage:
 
 Rebuilding the openedx-dev image
 --------------------------------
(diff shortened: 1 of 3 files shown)


kdmccormick:
@regisb and @bradenmacdonald  , here's my first stab at getting `pytest` to work. Let me know what you think of this PR and of the linked edx-platform PR.

docs/dev.rst
```diff
@@ -255,26 +255,30 @@ You should then run the development server as usual, with ``start``. Every chang
...
-    export DJANGO_SETTINGS_MODULE=cms.envs.tutor.test
-    pytest cms
+    tutor dev run cms bash
+
+As before, simply invoke pytest. Tests under the cms/ source tree will automatically use CMS settings. For tests under the shared source trees, you will need to tell pytest to look in the cms/ directory for configuration by providing ``--rootdir`` option::
+
+    pytest cms                                                            # Run tests for the entire CMS source tree.
+    pytest cms/djangoapps/coursegraph                                     # Run tests in a single CMS app.
+    pytest --rootdir=cms openedx common                                   # Run tests for both shared source trees.
+    pytest --rootdir=cms openedx/core/djangoapps/content/course_overviews # Run tests for a single shared app.
```
regisb, line 280: This contradicts the instruction from line 268. I'm confused: do we need to specify the rootdir when running shared app tests?

regisb:
> In trying to find out why we need to change DOC_STORE_CONFIG connection credentials, git blame leads me to https://github.com/overhangio/tutor/commit/8d803fb08b13e2e20249b82c034483e61ec22795. @regisb , do you remember why you why need to change these credentials?

I can't remember, unfortunately. Did you manage to run the unit tests without these statements?

(fun fact: I opened https://github.com/overhangio/tutor/commit/8d803fb08b13e2e20249b82c034483e61ec22795 in a different tab in my browser and started reviewing it, incorrectly believing it was actually your PR, and thinking wow this is horrible there is no way we are making these changes...)

docs/dev.rst
kdmccormick: @regisb Right, sorry, it's a little convoluted.

By default, ALL tests will use `lms.envs.test` (because of edx-platform's setup.cfg) EXCEPT the tests under `cms/`, which will use `cms.envs.test` (because of edx-platform's cms/pytest.ini).

This is generally reasonable, except that sometimes folks will want to run tests for the "shared" directories (openedx/ and common/) in a CMS context. The easiest way to make this happen is to tell pytest to look at the cms/ directory for a pytest.ini, hence `--rootdir=cms`.

As I think more about this, I really wish we could just have the behavior of:
* tests in lms/ always use `lms.envs.test`,
* tests in cms/ always use `cms.envs.test`,
* tests in shared directories use `lms.envs.test` when in the LMS container, and use `cms.envs.test` when in the CMS container.

Orchestrating that will take some hair-pulling. For the conference, I think we'll just have to accept the status quo. The current instructions, while a little annoying, certainly work for the vast majority of edx-platform tests :)

kdmccormick:
Closing for now per https://github.com/overhangio/tutor/pull/648#discussion_r857029417

PR #650, opened 2022-04-23
docs: Update ARM64 tutorial to reference new plugin

bradenmacdonald:
This updates the documentation to point to https://github.com/open-craft/tutor-contrib-arm64 as the easiest way to get Tutor working on ARM64 systems for now. It also recommends using MySQL 8 instead of MariaDB for newer platform versions that are compatible.

### DIFF
--- a/docs/tutorials/arm64.rst
+++ b/docs/tutorials/arm64.rst
@@ -1,22 +1,36 @@
 .. _arm64:
 
+==================================
 Running Tutor on ARM-based systems
 ==================================
 
 Tutor can be used on ARM64 systems, although support for that platform is currently experimental.
 
-There are generally two ways to run Tutor on an ARM system - using qemu to run x86_64 images using emulation or running native ARM images. Since emulation can be quite slow, this Tutorial will focus on using native images where possible.
-
-There are currently no official ARM64 images provided for Tutor, but Tutor makes it easy to build them yourself.
-
-Building the images
--------------------
-
-Start by :ref:`installing <install>` Tutor and its dependencies (e.g. Docker) onto your system.
+In any case, start by :ref:`installing <install>` Tutor and its dependencies (e.g. Docker) onto your system.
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/arm64.rst
```diff
@@ -1,22 +1,36 @@
...
 
-There are currently no official ARM64 images provided for Tutor, but Tutor makes it easy to build them yourself.
+.. note:: For Open edX developers, if you want to use the :ref:`nightly <nightly>` version of Tutor to "run master", install Tutor using git and check out the ``nightly`` branch of Tutor at this point. See the :ref:`nightly documentation <nightly>` for details.
 
-Building the images
--------------------
+Then, follow one of the options in this tutorial:
 
-Start by :ref:`installing <install>` Tutor and its dependencies (e.g. Docker) onto your system.
+The easy way: tutor-contrib-arm64
```
regisb, line 15: Can you commit to maintaining this plugin in the future? If not, I'd rather not recommend to use it in the official docs and only mention it in the conference tutorial.

bradenmacdonald: Well, how often do you think I would need to build and publish updated ARM64 images on our Docker Hub? If it's not any more frequently than weekly, sure I can commit to maintaining this plugin. I do hope that eventually it won't be needed, and you can find a solution to publish cross-platform official builds, but for now I'm happy to help in this way.

regisb: If we want to keep these instructions in the docs for the long term, then we need to address these two scenarios (and possibly other):
- people should be able to run `tutor local quickstart` as soon as a new release comes out. Thus, the image build should be triggered after every tutor release.
- past version of Tutor and Open edX should keep working; thus, images need to be tagged and not regularly overwritten. 

I appreciate your offer to run a CI to build and maintain these images; but I'm worried that having non-pinned images is going to lead to hard-to-debug problems, where people show up on the forum with incoherent Tutor and Open edX versions. I'm not sure how to best resolve this problem. I guess that I should be running a multi-platform build, but before that can happen I'd rather emphasise that the non-pinned image is a temporary one.

bradenmacdonald: Yes, I certainly agree with your assessment of the problem. Given that, I'm fine to just try out this plugin + images for the conference and we can keep thinking about what a longer term solution will look like. Feel free to close this PR for now.

PR #655, opened 2022-04-30
Learning MFE issue with certificates

CodeWithEmad:
as mentioned [here](https://github.com/overhangio/tutor-mfe/issues/45), after migrating to MFEs, if PersistentGradesEnabledFlag is not enabled, after the certification creation request, an error will happen and it makes the pod restart.
this fix enables the PERSISTENT_GRADES_ENABLED_FOR_ALL_TESTS feature for lms. much much cleaner than creating a record that is not reversible.


CodeWithEmad:
doesn't it need to add anything to documentation or changelogs?

PR #659, opened 2022-05-10
feat:added event routing backend plugin as dependency

ziafazal:
Added event routing backend plugin as dependency


ziafazal:
@regisb Could you please review this PR?

regisb:
@ziafazal I need some further information. Do you think that all Open edX users should have edx-event-routing-backends installed by default on their platform? Why?

ziafazal:
@regisb No I don't think all Open edX users should have edx-event-routing-backends installed. I just want to install it for testing nutmeg on https://nutmeg.demo.overhang.io/. So it would be just a local dependency for https://nutmeg.demo.overhang.io/ instance.

regisb:
In that case, I suggest that you close this PR and instead add the edx-event-routing-backends just in the demo platform, very much in the spirit of [this PR](https://github.com/overhangio/nutmeg-demo/pull/11).

ziafazal:
Closing this PR in favour of [this](https://github.com/overhangio/nutmeg-demo/pull/13) one.

PR #666, opened 2022-05-16
feat: pinned nodejs requirements with `npm ci`

regisb:
Contrary to what we might expect, `npm install` does not install pinned
requirements from a project's package-lock.json. That's the responsibility of
`npm ci`: https://docs.npmjs.com/cli/v8/commands/npm-ci

Running `npm ci` is also *much* faster than `npm install`, so that's a huge win.

See this issue for reference: https://github.com/openedx/frontend-wg/issues/100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


regisb:
This is ready for review @overhangio/tutor-developers. Cc @ghassanmas.

ghassanmas (commented):
This will be applied for building the image in production and development right? if yes
I think for development using `npm install` would be a batter option (unless developer in development mode is not concnered about node_modules)

PR #669, opened 2022-05-19
fix: avoid double-rendering mounts to docker-compose.tmp.yml

kdmccormick:
In certain code paths, such as in `tutor local quickstart`,
`process_mount_points` is called more than once in the same process,
causing mounts to be added to `COMPOSE_LOCAL[_JOBS]_TMP` redundantly.
As a result, docker-compose[.jobs].tmp.yml was occasionally being
rendered with duplicate volume specifiers. Some versions of Docker
Compose ignored this; other versions warned or threw an error.

In order to make `process_mount_points` tolerant to being called
multiple times, we wrap its volume-adding callbacks within a new
hooks context. This allows us to clear said hooks context every
time `process_mount_points` is called, essentially making the
function idempotent.

### DIFF
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -465,7 +465,7 @@ def dc_command(context: BaseComposeContext, command: str, args: t.List[str]) -> 
     context.job_runner(config).docker_compose(command, *volume_args, *non_volume_args)
 
 
-def process_mount_arguments(mounts: t.Tuple[t.List[MountParam.MountType]]) -> None:
+def process_mount_arguments(mounts: t.Tuple[t.List[MountParam.MountType], ...]) -> None:
     """
     Process --mount arguments.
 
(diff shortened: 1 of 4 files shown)


tutor/commands/compose.py
```diff
@@ -465,7 +465,7 @@ def dc_command(context: BaseComposeContext, command: str, args: t.List[str]) ->
     context.job_runner(config).docker_compose(command, *volume_args, *non_volume_args)
 
 
-def process_mount_arguments(mounts: t.Tuple[t.List[MountParam.MountType]]) -> None:
+def process_mount_arguments(mounts: t.Tuple[t.List[MountParam.MountType], ...]) -> None:
```
regisb, line 468: Why do we need to modify the function signature?

regisb (commented):
If I understand correctly, the problem is that during quickstart we are making multiple calls to `process_mount_arguments` and this is causing the same callbacks to be associated multiple times to the `COMPOSE_LOCAL_TMP` and `COMPOSE_LOCAL_JOBS_TMP` filters. Your solution works, but I think that we should also address this underlying issue. The `process_mount_arguments` function is a bit hackish (mea culpa), and it should be improved.

I propose the following solution:
1. we introduce the `COMPOSE_LOCAL_CLI_MOUNTS` and `COMPOSE_LOCAL_JOBS_CLI_MOUNTS` filters. 
2. when we call `process_mount_arguments` we override the return values of these two filters.
3. the `_add_mounts_to_docker_compose_tmp` and `_add_mounts_to_docker_compose_jobs_tmp` functions are taken out of the `process_mount_arguments` function. When they are called, they read their values from the two filters we created.

Do you see what I mean? The filter names are not great and could certainly be improved.

regisb:
Sorry for the delayed review @kdmccormick. It would be fantastic if you could update this PR with the new hooks. I agree that my proposed filter names are not great. But I think that we cannot use `COMPOSE_VOLUMES` or `COMPOSE_JOBS_VOLUMES`. The reason for that is that the `process_mount_arguments` is going to override the values of these filters. Thus, if another plugin tries to add callbacks to these filters, they will be ignored.

Let me suggest an alternative implementation with contexts. With this alternative implementation we will not have to create new filters. It goes like this:

```
# Clear any previously defined filter created in this context
hooks.Filters.COMPOSE_LOCAL_TMP.clear(context=hooks.Contexts.COMPOSE_LOCAL_CLI)

# Recreate filter callbacks
with hooks.Contexts.COMPOSE_LOCAL_CLI.enter():
    @hooks.Filters.COMPOSE_LOCAL_TMP.add()
    def _add_mounts_to_docker_compose_tmp(...):
        ...
    @hooks.Filters.COMPOSE_LOCAL_JOBS_TMP.add()
    def _add_mounts_to_docker_compose_jobs_tmp(...):
        ...
```

What do you think?

kdmccormick:
Thanks for your patience. Looks good to me -- I'll apply those changes soon.

regisb (approved):
Thanks for your patience!

PR #673, opened 2022-05-23
docs: complete uninstalling of tutor

iamCristYe:
I noticed `pip uninstall -y tutor` will not uninstall the plugins, so I made this PR. I know it's ugly, but I don't find any other way of doing it. Let me know if there are better choices 😊

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -167,3 +167,9 @@ Finally, uninstall Tutor itself::
 
     # If you downloaded the tutor binary
     sudo rm /usr/local/bin/tutor
+
+    # Optionally, you may want to remove Tutor plugins installed.
+    # You can get a list of the installed plugins:
+    pip freeze | grep tutor
+    # You can then remove them using the following command:
+    pip uninstall <plugin-name>


docs/install.rst
```diff
@@ -163,7 +163,18 @@ Then, delete all data associated with your Open edX platform::
 Finally, uninstall Tutor itself::
 
     # If you installed tutor from source
-    pip uninstall tutor
+    pip uninstall -y tutor
+    pip uninstall -y tutor-android
```
BbrSofiane, line 167: How about?
```
pip uninstall -y <plugin_name>
```

Or are those specific plugins installed?

iamCristYe: Those specified in https://github.com/overhangio/tutor/blob/master/setup.py#L62 and https://github.com/overhangio/tutor/blob/master/requirements/plugins.txt will be installed when installing using the recommended approach `pip install tutor[full]`

regisb: I'd rather not give the full list of plugins here, it's just too cumbersome. Moreover, it's not _strictly_ necessary to uninstall plugins to "get rid" of tutor. Thus, I suggest that we give generic instructions on how to uninstall plugins. Something along those lines:

> List all Tutor plugins
> 
>     $ pip freeze | grep tutor
> 
> Then, uninstall every plugin:
> 
>     $ pip uninstall <tutor plugin name>

regisb (approved):

PR #674, opened 2022-05-24
fix: ignore plugins that cannot be loaded

regisb:
When running multiple concurrent versions of a plugin there are sometimes
version conflicts that prevent the plugin from being loaded. Prior to v1, Tutor
was correctly ignoring plugins that could not be loaded. During the transition
to v1 we lost that feature because we only captured TutorErrors.


regisb:

regisb:
I need to merge this quickly to fix my local installation.

Issue #518, opened 2021-11-02
`settheme` fails when LMS_HOSTNAME or CMS_HOSTNAME is over 50 characters


mafonso:
**Bug description**

The new implementation of `settheme` has an issue when the domain name is very long (>50 chars) 

**How to reproduce**

Assign `LMS_HOSTNAME` or `CMS_HOSTNAME` a long domain name that is over 45 chars.  

**Environment**

Tutor version 12.1.5 running in Kubernetes (docker image)

**Additional context**

in `tutor/jobs.py`

```
site, _ = Site.objects.get_or_create(domain='{domain_name}')
if not site.name:
    site.name = '{domain_name}'
    site.save()
site.themes.all().delete()
site.themes.create(theme_dir_name='{theme_name}')
```

site.name is being assigned from the domain name, which is a problem on itself, as they have different length limits on the Django sites data model.  Domain is limited to 100 chars, and site.name is limited to 50.

We had a use case where the CMS FQDN amounted to 48 chars, but then the port was appended to the domain name made it go over the 50 limit.

```
def get_all_openedx_domains(config: Config) -> List[str]:
    return [
        get_typed(config, "LMS_HOST", str),
        get_typed(config, "LMS_HOST", str) + ":8000",
        get_typed(config, "CMS_HOST", str),
        get_typed(config, "CMS_HOST", str) + ":8001",
    ]
```

regisb:
To reproduce this issue, run:

```
$ tutor local settheme -d aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.com indigo
...
Error: Command failed with status 1: docker-compose -f /home/regis/.local/share/tutor/env/local/docker-compose.yml -f /home/regis/.local/share/tutor/env/local/docker-compose.prod.yml --project-name tutor_local -f /home/regis/.local/share/tutor/env/local/docker-compose.jobs.yml run --rm lms-job sh -e -c '
export DJANGO_SETTINGS_MODULE=$SERVICE_VARIANT.envs.$SETTINGS
echo "Loading settings $DJANGO_SETTINGS_MODULE"
./manage.py lms shell -c "from django.contrib.sites.models import Site
print('"'"'Assigning theme indigo to aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.com...'"'"')
[...]

PR #676, opened 2022-05-26
fix: trim name for site configuration before saving

navinkarkera:
This MR fixes an issue which occurs when instance domain names exceed 50 characters in length by trimming name to 50 characters before saving.

Error log:
```console
Assigning theme edx-simple-theme to studio.alongtestname.development.opencraft.hosting.
--
File "/openedx/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
return self.cursor.execute(sql, params)
File "/openedx/venv/lib/python3.8/site-packages/django/db/backends/mysql/base.py", line 73, in execute
return self.cursor.execute(query, args)
File "/openedx/venv/lib/python3.8/site-packages/MySQLdb/cursors.py", line 206, in execute
res = self._query(query)
File "/openedx/venv/lib/python3.8/site-packages/MySQLdb/cursors.py", line 319, in _query
db.query(q)
File "/openedx/venv/lib/python3.8/site-packages/MySQLdb/connections.py", line 259, in query
_mysql.connection.query(self, query)
MySQLdb._exceptions.DataError: (1406, "Data too long for column 'name' at row 1")

```

Openedx is utilizing [django's sites framework](https://docs.djangoproject.com/en/4.0/ref/contrib/sites/#module-django.contrib.sites) to store domain names in the database.
The `max_length=50` specified for the `name` column in the [Site model](https://github.com/django/django/blob/main/django/contrib/sites/models.py#L87) is causing this failure.


keithgg:
[~navinkarkera] I agree with @Alec4r assessment. We should rather raise an error message than mangle the user's input.

Can we check if the domain is within the length limits and if not, exit nicely so that the user knows what the problem is?

erickhgm:
`Name` is just a description. I don't why we have only 50 chars to it.

**Documentation**

[class models.Site](https://docs.djangoproject.com/en/4.0/ref/contrib/sites/#django.contrib.sites.models.Site)
A model for storing the domain and name attributes of a website.

> [domain](https://docs.djangoproject.com/en/4.0/ref/contrib/sites/#django.contrib.sites.models.Site.domain)
> The fully qualified domain name associated with the website. For example, www.example.com.
> 
> [name](https://docs.djangoproject.com/en/4.0/ref/contrib/sites/#django.contrib.sites.models.Site.name)
> A human-readable “verbose” name for the website.

**Code**
```
class Site(models.Model):
    domain = models.CharField(
        _("domain name"),
        max_length=100,
        validators=[_simple_domain_name_validator],
        unique=True,
    )
    name = models.CharField(_("display name"), max_length=50)
```

navinkarkera:
>Name is just a description. I don't know why we have only 50 chars to it.

Yes, it should be set to a descriptive name for site instead saving domain again.

tutor/jobs.py
```diff
@@ -126,11 +126,18 @@ def set_theme(
         return
     python_code = "from django.contrib.sites.models import Site"
     for domain_name in domain_names:
+        if len(domain_name) > 50:
```
keithgg, line 129: Instead of 50. Use `Site._meta.get_field('name').max_length`

Shouldn't this be part of the python

keithgg (approved):
LGTM :+1:

tutor/jobs.py
navinkarkera: @keithgg This will require us to have django installed locally.

keithgg: Ah, is it possible to rather move it to the `python_code` block instead?

navinkarkera: To print warnings we are using `fmt.echo_alert` which is part of tutor and not edx-platform, which is the reason I had to add it outside.

keithgg: Cool. Let's leave it as is. :+1:

tutor/jobs.py
```diff
@@ -126,11 +126,18 @@ def set_theme(
...
+            fmt.echo_alert(
+                f"The domain {domain_name} is too long to be used for site name, "
+                "truncating to 50 characters. Please update the site name using admin."
+            )
```
regisb, line 133: I think it's great that you added a warning, but we should modify the wording. It's not really a big deal if the site displayed name is incorrect, so we should not suggest to modify it in the admin. Instead, I propose the following:

> Assigning a theme to a site with a long (> 50 characters) domain name. The displayed site name will be truncated to 50 characters.

regisb (changes requested):
Almost there! Can you also please add "Close #518" to the git commit message to reference the corresponding issue?

tutor/jobs.py
navinkarkera: Thanks @regisb . Updated the commit with your suggestion.

regisb (approved):

regisb:
Thanks for your PR @navinkarkera!

PR #685, opened 2022-06-07
`Docker Compose` v2 compatibility fix #641

### LINKED ISSUE #641: Docker compose should upgrade to v2 (`docker compose`)
Docker-compose v2 is now available, stable and should be compatible with `docker-compose`: https://docs.docker.com/compose/cli-command/
Thus, we should no longer call `docker-compose` but `docker compose`. I'm not entirely sure how we can allow older versions of Docker to keep running.

If we require a recent version of docker-compose, we should take this opportunity to get rid of the `version: 3.7` statement at the top of the docker-compose.yml file: https://docs.docker.com/compose/compose-file/#version-top-level-element

Migrating to v2 will allow us to make us of [`docker compose cp`](https://docs.docker.com/engine/reference/commandline/compose_cp/).

iamCristYe:
for your review @regisb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 - [Bugfix] Fix smtp server port in `cms.yml` which was causing email sending failures in the Studio. (by @regisb)
 - [Improvement] Use `git am` instead of `cherry-pick` to simplify patching process.
+- [Improvement] Tutor is now compatible with Docker Compose subcommand.
 
 ## v14.0.1 (2022-06-13)
 
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -5,6 +5,9 @@ Local deployment
 
 This method is for deploying Open edX locally on a single server, where docker images are orchestrated with `docker-compose <https://docs.docker.com/compose/overview/>`_.
 
+.. note::
+    Tutor is compatible with the ``docker compose`` subcommand. However, this support is still in beta and we're not sure it will behave the same as the previous ``docker-compose`` command. So ``docker-compose`` will be preferred, unless you set an environment variable ``TUTOR_USE_COMPOSE_SUBCOMMAND`` to enforce using ``docker compose``.
+
 .. _tutor_root:
 
 In the following, environment and data files will be generated in a user-specific project folder which will be referred to as the "**project root**". On Linux, the default project root is ``~/.local/share/tutor``. An alternative project root can be defined by passing the ``--root=...`` option to the ``tutor`` command, or defining the ``TUTOR_ROOT=...`` environment variable::
--- a/tutor/utils.py
+++ b/tutor/utils.py
@@ -1,4 +1,5 @@
 import base64
+from functools import lru_cache
 import json
 import os
 import random
@@ -162,12 +163,25 @@ def docker(*command: str) -> int:
     return execute("docker", *command)
 
 
+@lru_cache
+def _docker_compose_command() -> Tuple[str, ...]:
+    """
+    A helper function to determine which program to call when running docker compose
+    """
+    if os.environ.get("TUTOR_USE_COMPOSE_SUBCOMMAND") is not None:
+        return ("docker", "compose")
+    if shutil.which("docker-compose") is not None:
+        return ("docker-compose",)
+    if shutil.which("docker") is not None:
+        if subprocess.run(["docker", "compose"], capture_output=True).returncode == 0:
+            return ("docker", "compose")
+    raise exceptions.TutorError(
+        "docker-compose is not installed. Please follow instructions from https://docs.docker.com/compose/install/"
+    )
+
+
 def docker_compose(*command: str) -> int:
-    if shutil.which("docker-compose") is None:
-        raise exceptions.TutorError(
-            "docker-compose is not installed. Please follow instructions from https://docs.docker.com/compose/install/"
-        )
-    return execute("docker-compose", *command)
+    return execute(*_docker_compose_command(), *command)
 
 
 def kubectl(*command: str) -> int:


Alec4r:
Change looks good for me, maybe we can use guard clauses instead of if-elif-else. @regisb

tutor/utils.py
```diff
@@ -163,11 +163,19 @@ def docker(*command: str) -> int:
 
 
 def docker_compose(*command: str) -> int:
-    if shutil.which("docker-compose") is None:
+    """
+    Fallback to docker compose when docker-compose is not available
+    """
+    if shutil.which("docker-compose") is not None:
+        return execute("docker-compose", *command)
+    elif "is not a docker command" not in str(
```
regisb, line 171: I understand that you are doing this to make sure that docker-compose v2 is installed, but we should not run this command for every call to `docker-compose`. 

Also, I'd like to be able to force the use of docker-compose v2 by defining an environment variable; this would be useful for me to troubleshoot v2 issues.

So, here's what I propose:

- If `TUTOR_USE_DOCKER_COMPOSE_V2` is in `os.environ`, use `docker compose`.
- Else, if "docker-compose" is present, use it.
- Else, check for `docker compose version` and cache the result with [lru_cache](https://docs.python.org/3.8/library/functools.html#functools.lru_cache). If present, use `docker compose`. (don't check for the presence of "is not a docker command", but check the result `returncode` instead)
- Else, raise TutorError.

Also, please add an entry to the changelog, following: https://docs.tutor.overhang.io/tutor.html#contributing

regisb: > Why is an env variable preferred over a variable in config.yml

config.yml files are meant to be portable from one server to the next, despite the fact that they may not all run the same docker-compose version.

iamCristYe:
For your reference, I use the name `TUTOR_USE_COMPOSE_PLUGIN` because technically compose v2 can be installed in standalone mode.

tutor/utils.py
```diff
@@ -162,12 +163,29 @@ def docker(*command: str) -> int:
...
+
 def docker_compose(*command: str) -> int:
-    if shutil.which("docker-compose") is None:
-        raise exceptions.TutorError(
-            "docker-compose is not installed. Please follow instructions from https://docs.docker.com/compose/install/"
-        )
-    return execute("docker-compose", *command)
+    if _which_docker_compose() == "docker":
+        return execute("docker", "compose", *command)
+    if _which_docker_compose() == "docker-compose":
```
regisb, line 184: There are a few problems with your implementation:

1. Multiple calls are made to `_which_docker_compose`. Even though the results are lru-cached, it's inelegant.
2. The name of the `TUTOR_USE_COMPOSE_PLUGIN` environment variable is not really adequate. I understand that you refer to the Docker compose plugin, but people will mistake it with Tutor plugins (I did). I suggest instead ` TUTOR_USE_DOCKER_COMPOSE_V2`.
3. The string returned by `_which_docker_compose` is rather arbitrary and makes it more difficult to read the code.

I suggest that you move back the code of the `_which_docker_compose` back inline into the `docker_compose` function. You should only create a separate lru-cached function for fetching the `docker compose version`, which should return either a `str` or None.

regisb: > I agree with you that the PLUGIN in the name seems confusing, but I really don't like to use V2 to refer to using docker compose as I've been using docker-compose v2 for a long time. So I propose TUTOR_USE_DC_AS_SUBCOMMAND.

"DC" is not easy to understand out of context. I counter-propose `TUTOR_USE_COMPOSE_SUBCOMMAND`.

> But the case is os.environ.get("TUTOR_USE_DC_AS_SUBCOMMAND") and shutil.which("docker-compose") gets computed over and over again. Wouldn't it be better to cache them?

Yes, you are right. Thus, I propose:

```
@lru_cache
def _docker_compose_command():
    if ...:
        return ("docker-compose",)
    if ...:
        return ("docker", "compose")
    raise ...

def docker_compose(*command):
    return execute(*_docker_compose_command(), *command)
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Every user-facing change should have an entry in this changelog. Please respect
 
 - [Bugfix] Fix smtp server port in `cms.yml` which was causing email sending failures in the Studio. (by @regisb)
 - [Improvement] Use `git am` instead of `cherry-pick` to simplify patching process.
+- [Improvement] Tutor is now compatibe with Docker Compose subcommand.
```
regisb, line 23: nit: compatibe -> compatible.

tutor/utils.py
```diff
@@ -162,12 +163,25 @@ def docker(*command: str) -> int:
     return execute("docker", *command)
 
 
+@lru_cache
+def _docker_compose_command() -> str:
+    """
+    A helper function to determine which program to call when running docker compose
+    """
+    if os.environ.get("TUTOR_USE_COMPOSE_SUBCOMMAND") is not None:
```
regisb, line 171: Could you add a mention of this environment variable in the docs? (in local.rst for instance) Please make it clear that Docker Compose v2 is still in test mode, and we are not quite sure yet whether we are fully compatible.

regisb (changes requested):

tutor/utils.py
```diff
@@ -162,12 +163,25 @@ def docker(*command: str) -> int:
     return execute("docker", *command)
 
 
+@lru_cache
+def _docker_compose_command() -> str:
```
regisb, line 167: The return type should be `t.Tuple[str, ...]` (see failing type tests).

docs/local.rst
```diff
@@ -5,6 +5,9 @@ Local deployment
 
 This method is for deploying Open edX locally on a single server, where docker images are orchestrated with `docker-compose <https://docs.docker.com/compose/overview/>`_.
 
+.. note::
+    Tutor is compatible with the ``docker compose`` subcommand. However, this support is still in beta and we're not sure it will behave the same as the previous ``docker-compose`` command. So ``docker-compose`` will be preferred, unless you set an environment variable ``TUTOR_USE_COMPOSE_SUBCOMMAND`` to enforce using ``docker compose``.
```
regisb, line 9: Awesome :ok_hand:

regisb (approved):
Thanks for your time on this PR @iamCristYe!

PR #687, opened 2022-06-07
chore: change default of catalog settings

ghassanmas:
This partialy fixes #openedx/build-test-release-wg/issues/164
 The flag would make it possible to hide courses in the /courses
 page when ENABLE_COURSE_DISCOVERY is true.
 For more context check the forum thread:
 https://discuss.openedx.org/t/request-for-feedbak-breaking-changes-for-course-visibility/7341

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_lms.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_lms.py
@@ -21,6 +21,11 @@ OAUTH_ENFORCE_SECURE = False
 # Email settings
 DEFAULT_EMAIL_LOGO_URL = LMS_ROOT_URL + "/theming/asset/images/logo.png"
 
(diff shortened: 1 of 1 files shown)


regisb:
I'm not in favour or merging this right now for the following reasons:

- This is a breaking change that would require a major version bump.
- Nutmeg will be released in 2 days and it already include the fix: https://github.com/overhangio/tutor/pull/635/files#diff-c7b8fd3e439b9c293fbec7493376cee7fbccf4c641b113b77d68167f9585d07a
- In Maple, this setting can be easily enabled via a plugin.

ghassanmas:
Alright!, I didn't know you already added it in nutmeg, I was rushing to add it as per last BTR meeting...So should this be closed or labeled WIP? as to keep for reference... and IMHO this better to be thought through or resolved upstream.

Issue #689, opened 2022-06-11
Migration needed before upgrade_from_maple


ibrmora:
**Bug description**
When you upgrade from Maple to Nutmeg using command
`tutor local upgrade --from=maple`

the result

`MySQLdb._exceptions.ProgrammingError: (1146, "Table 'openedx.user_tours_usertour' doesn't exist")
ERROR: 1
Error: Command failed with status 1: docker-compose -f /root/.local/share/tutor/env/local/docker-compose.yml -f /root/.local/share/tutor/env/local/docker-compose.prod.yml -f /root/.local/share/tutor/env/local/docker-compose.tmp.yml --project-name tutor_local run --rm lms sh -e -c ./manage.py lms backpopulate_user_tours`

to avoid the error run migration on both 

`tutor local run lms ./manage.py lms migrate`
`tutor local run cms ./manage.py cms migrate`

then run  `tutor local upgrade --from=maple`  again

or add migration to _upgrade_from_maple_ function
https://github.com/overhangio/tutor/blob/v14.0.0/tutor/commands/upgrade/local.py#L102

**Environment**

update from 
tutor, version 13.2.3      open-release/maple.3
to
tutor, version 14.0.0      open-release/nutmeg.1

insad:
I mentioned this already here https://discuss.openedx.org/t/tutor-upgrade-to-nutmeg-error-because-of-uncreated-tables/7452

regisb:
This is resolved in v14.0.1.

PR #691, opened 2022-06-12
fix: upgrade_from_maple issue #689

ibrmora:
fix: Migration needed before upgrade_from_maple issue #689


ibrmora:
Another table need to migrate.

PR #693, opened 2022-06-16
[Improvement] Use `git am` instead of `cherry-pick` to simplify patching process

iamCristYe:
When reviewing Dockerfile recently, I noticed that the git apply patch lines (where I learned the GitHub .diff and .patch trick) were replaced by cherry-pick. I noticed from the [commit](https://github.com/overhangio/tutor/commit/ca3aabedce250ef5a34c9f89a6545d984cdef30c) that this is for a better git working dir. However, apparently fetching the whole repo (even with the `depth` option) is not elegant. After some research, I found that `git am` is a better choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


regisb (approved):
This is vastly superior, thanks for the PR.

PR #699, opened 2022-06-30
fix: replica set error from pymongo 3.10 -> 3.12 upgrade

ormsbee:
The pymongo dependency for edx-platform was updated (3.10.1 to 3.12.3)
in https://github.com/openedx/edx-platform/pull/30569

This caused the following error when running the edx-platform database
migration split_modulestore_django.0002_data_migration  as part of
`tutor dev quickstart`:

```
  pymongo.errors.ServerSelectionTimeoutError: client is configured to
  connect to a replica set named '' but this node belongs to a set named
  'None', Timeout: 30s, Topology Description: <TopologyDescription id:
  62bdbaf182687350acf1aeec, topology_type: Single, servers:
  [<ServerDescription ('mongodb', 27017) server_type: Unknown, rtt:
   None, error=ConfigurationError("client is configured to connect to a
   replica set named '' but this node belongs to a set named 'None'")>]>
```

This commit explicitly sets replicaSet to None to indicate that it's a
standalone MongoDB instance.


ormsbee:
(Actually, going to hold on this because while it did get me past the migration, I should wait until I have a fully functioning local tutor dev env.)

ormsbee:
Ah, okay, so tutor's `development.py` would set the MongoDB parameters, but that file imports everything from `devstack.py` and that's where `CONTENTSTORE['DOC_STORE_CONFIG']['replicaSet'] = None` is getting set when there's no value for `CONTENTSTORE['DOC_STORE_CONFIG']`, and thus explosion.

But then, why didn't that change break devstack too?

regisb:
The error you are facing is because Tutor sets `CONTENTSTORE: null` in lms.yml: 
https://github.com/overhangio/tutor/blob/b8f773a5efde67a769301aecf8434b8c19bbea93/tutor/templates/apps/openedx/config/partials/auth.yml#L4
https://github.com/overhangio/tutor/blob/b8f773a5efde67a769301aecf8434b8c19bbea93/tutor/templates/apps/openedx/config/lms.env.yml#L86
Here's the original commit which introduced this null value: https://github.com/overhangio/tutor/commit/f32eb089404bcc4ffd8669e54f797c5f34bda1b8
I'm not sure whether you can remove this entry from auth.yml. Maybe you can try?

In any case,

ormsbee:
Thanks @regisb!

ormsbee:
@regisb: That worked! Though I don't really understand what taking that out might break...?

ormsbee:
I'll add it to the PR anyhow.

ormsbee:
@regisb: Does this look okay to you?

ormsbee:
Ah, forgot the CHANGELOG policy in this repo. Going to add that now...

ormsbee:
Okay, now I think it's ready for review. 😄 Thank you for your patience as I stumble through my first Tutor contribution.

aht007:
> Ah, okay, so tutor's `development.py` would set the MongoDB parameters with the templated value altered in this PR, but that file imports everything from `devstack.py` and that's where `CONTENTSTORE['DOC_STORE_CONFIG']['replicaSet'] = None` is getting set when there's no value for `CONTENTSTORE['DOC_STORE_CONFIG']`, and thus explosion.
> 
> But then, why didn't that change break devstack too?

This change didn't cause any issues because in the pymongo upgrade PR we already tackled this in [configuration](https://github.com/openedx/edx-platform/pull/30569/files#diff-ede4abf6f02ac511ae5f1ec419fef6ee86c958f5a06b2be41bd8702a591c5e63R240)
I didn't know if it would cause any issue with tutor else I would have already given a heads up.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Every user-facing change should have an entry in this changelog. Please respect
 -->
 
 ## Unreleased
+- [Bugfix] Fix MongoDB replica set connection error resulting from edx-platform's pymongo (3.10.1 -> 3.12.3) upgrade ([edx-platform#30569](https://github.com/openedx/edx-platform/pull/30569)). (by @ormsbee)
```
regisb, line 20: Can you please move that line to the CHANGELOG-nightly.md file? In the nightly branch we store changes there, otherwise there are too many conflicts when we merge the master branch.

tutor/templates/apps/openedx/config/partials/auth.yml
```diff
@@ -1,7 +1,6 @@
 SECRET_KEY: "{{ OPENEDX_SECRET_KEY }}"
 AWS_ACCESS_KEY_ID: "{{ OPENEDX_AWS_ACCESS_KEY }}"
 AWS_SECRET_ACCESS_KEY: "{{ OPENEDX_AWS_SECRET_ACCESS_KEY }}"
-CONTENTSTORE: null
```
regisb, line 4: For the record, this line is no longer necessary in cms/envs/production.py since this commit: https://github.com/openedx/edx-platform//commit/7d2a8340c5e997bf4ca43ca537a91e8f0962df1f#diff-aa2b311852d707265f0324225f3d9178bed527d71c84cd296de4de6a8d2c190dL388

regisb (changes requested):

CHANGELOG.md
ormsbee: Done.

regisb (approved):
It's cool to merge your first PR here Dave :)

PR #700, opened 2022-07-04
fix: build openedx-dev image when host user is root

regisb:
Sometimes, the host user is root: this may happen when tutor is run with
"sudo" (which is not recommended) or on Windows. In such cases, building
the image should not fail, but default to a reasonable user. Also, when
we pass an invalid APP_USER_ID as a build arg, then we should fail with
an explicit message.

See this conversation:
https://discuss.overhang.io/t/problem-with-dev-image-build-useradd-uid-0-is-not-unique/2406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Every user-facing change should have an entry in this changelog. Please respect 
 -->
 
 ## Unreleased
+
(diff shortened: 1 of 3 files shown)


regisb:

kdmccormick (approved):

PR #705, opened 2022-07-13
Fix: Issue 702: tutor k8s start fails with no PatchStrategicMerge

edazzocaisser:
Fix: tutor k8s start will succeed when `k8s-override` and `kustomization-patches-strategic-merge` aren't specified.

https://github.com/overhangio/tutor/issues/704

Starting with version 14.0.2, every time we tried to run `tutor k8s start`, we would receive the following error:

```
Run tutor k8s start
kubectl get namespaces openedx-prod
NAME           STATUS   AGE
openedx-prod   Active   49d
Namespace already exists: skipping creation.
kubectl apply --kustomize /home/runner/.local/share/tutor/env --wait --selector app.kubernetes.io/component=volume
error: trouble configuring builtin PatchStrategicMergeTransformer with config: `
paths:
- k8s/override.yml
`: patch appears to be empty; files=[k8s/override.yml], Patch=
Error: Command failed with status 1: kubectl apply --kustomize /home/runner/.local/share/tutor/env --wait --selector app.kubernetes.io/component=volume
Error: Process completed with exit code 1.
```

Since we don't specify a `k8s-override` resulting in an empty `k8s/override.yml`, which is unsupported by kustomize. No `kustomization-patches-strategic-merge` also results in an empty `patchesStrategicMerge`.

I opened an issue for this bug, but since it seemed like a straightforward fix, I just made the changes.

PR #709, opened 2022-07-21
fix: k8s wait for deployments instead of pods to determine readiness

keithgg:
## Description

Currently tutor waits for k8s pods to be ready by running the `kubectl wait` command like this:

`kubectl wait --namespace openedx --selector=app.kubernetes.io/name=elasticsearch --for=condition=ContainersReady --timeout=600s pod`

This works for the most part, however it fails if there are existing pods not in the correct state.

For example, if a the Elastic Search pod is Evicted, you can end up with a pod list that looks list this if the nodes were experiencing memory pressure.

```
elasticsearch-78dcff9c9d-69lqp   0/1     Evicted     0          13h
elasticsearch-78dcff9c9d-clp66   0/1     OOMKilled   0          16h
elasticsearch-78dcff9c9d-g28bp   0/1     Evicted     0          13h
elasticsearch-78dcff9c9d-hngsd   1/1     Running     0          13h
elasticsearch-78dcff9c9d-jx2n7   0/1     Evicted     0          13h
elasticsearch-78dcff9c9d-nfc4f   0/1     Evicted     0          13h
elasticsearch-78dcff9c9d-pth9t   0/1     Evicted     0          13h
```

Even though there is still valid running pod, the `kubectl wait` condition will fail and time out, resulting in a failed deploy.

Instead of waiting for pods, I propose instead we should wait for `deployments` as they become `Available` as soon as the pods are ready. So far, I've found that it fixes the above issue with no downside.

### Verify that everything works normally

Run `tutor k8s quickstart`. The behaviour should be unchanged except `kubectl wait` command mentioning deployments instead of pods.

### Verify the fix

I haven't quite figured out the best way to create extra pods in a "broken" state so this is the best I can do.

We start 20 replicas of the elasticsearch pod and then almost immediately scale down to 1. If you get the pods list (`kubectl get pods -nopenedx`) it'll show a bunch of pods in the `Terminating` state.

- `kubectl scale deployment elasticsearch -nopenedx --replicas=20; sleep 1; kubectl scale deployment elasticsearch -nopenedx --replicas=1`: Create ElasticSearch pods where almost all will be in `Terminating` status
- `kubectl wait --namespace openedx --selector=app.kubernetes.io/name=elasticsearch --for=condition=Available=True --timeout=600s deployment` Will exit immediately as there will be 1 instance corresponding to the number required for the deployment.
- `kubectl wait --namespace openedx --selector=app.kubernetes.io/name=elasticsearch --for=condition=ContainersReady --timeout=600s pod` will not until the other pods are terminated.

## Other
- I've included the changes made in #705 in this PR for easier testing, I'll remove them if approved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 - [Fix] `tutor dev quickstart` would fail under certain versions of docker-compose due to a bug in the logic that handled volume mounting. (by @kdmccormick)
 - [Bugfix] The `tutor k8s start` command will succeed even when `k8s-override` and `kustomization-patches-strategic-merge` are not specified. (by @edazzocaisser)
+- [Fix] `kubectl wait` checks deployments instead of pods as it could hang indefinitely if there are extra pods in a broken state. (by @keithgg)
 
 ## v14.0.3 (2022-07-09)
 
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -332,10 +332,10 @@ def delete(context: K8sContext, yes: bool) -> None:
 def init(context: K8sContext, limit: Optional[str]) -> None:
     config = tutor_config.load(context.root)
     runner = context.job_runner(config)
-    wait_for_pod_ready(config, "caddy")
+    wait_for_deployment_ready(config, "caddy")
     for name in ["elasticsearch", "mysql", "mongodb"]:
         if tutor_config.is_service_activated(config, name):
-            wait_for_pod_ready(config, name)
+            wait_for_deployment_ready(config, name)
     jobs.initialise(runner, limit_to=limit)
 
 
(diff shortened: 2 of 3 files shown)


tutor/templates/kustomization.yml
```diff
@@ -59,8 +59,12 @@ configMapGenerator:
         app.kubernetes.io/name: redis
 {{ patch("kustomization-configmapgenerator") }}
 
+{%- if patch("k8s-override") or patch("kustomization-patches-strategic-merge") %}
```
gabor-boros, line 62: @keithgg I may don't get it, but why are these ifs are needed?

keithgg: @gabor-boros these changes are part of https://github.com/overhangio/tutor/pull/705. If you don't have those patches by default, you'll get an error.

regisb (commented):
LGTM! Can you please reformat your code (`make format`) and rebase your changes? Then we'll be ready to merge.

keithgg:
Thank you @regisb! I've rebased and formatted the code.

regisb (changes requested):
One more thing: could you add a line to the CHANGELOG.md file? See: https://docs.tutor.overhang.io/tutor.html#contributing

keithgg:
No problem @regisb. Changelog updated.

regisb (approved):
Fantastic, thanks!

PR #712, opened 2022-07-25
build: run a subset of test suite under Python 3.7

kdmccormick:
## Description

This PR changes the CI suite to run `make test-lint` and `make test-unit` for both the oldest-supported (3.7) and latest-supported (3.9) Python versions.

We still only run the remaining tests on the latest-supported version (3.9):

* `test-format`, `test-docs`, and `test-pythonpackage`, because end users will not care whether these work on 3.7.
* `test-types`, because type annotations in 3.7 are missing a lot helpful features.

## Example

With this change, [`shlex.join`'s incompatibility with 3.7](https://github.com/overhangio/tutor/pull/708#discussion_r928912499) would have been caught by CI:

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,25 @@ on:
     branches: [ master, nightly ]
 
 jobs:
-  tests:
+  # Run a subset of the test suite using the oldest-supported Python version.
+  tests-compat:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


kdmccormick:
@regisb this is ready for your consideration. I know build speed and billing might be factors here, so I understand if you'd rather not.

regisb:
Yeah, if we were quite thorough we would test Tutor with multiple versions of Python, but I don't want to keep committers waiting for too long. Also in most cases it's a waste of CPU cycles and the earth is burning.

I'll close this for now, let's re-open it later if we find out that backward compatibility is an important issue.

PR #719, opened 2022-08-30
build: prepare Dockerfile for common/lib removal

kdmccormick:
Soon, running:

  pip install -r ./requirements/edx/base.txt

in edx-platform will no longer install the local
project (that is, `-e .`). To prepare for that change,
we add the line:

  pip install -e .

to the Dockerfile. This is backwards-compatible.
More details:
https://openedx.atlassian.net/browse/BOM-2575?focusedCommentId=613181


regisb (approved):
I'm looking forward to this! It's really cool to see changes made in the upstream repos to make life easier for Tutor users.

PR #720, opened 2022-09-08
depr: richie plugin is no longer official

regisb:
The richie plugin was transferred to the openfun GitHub org. As a consequence, it is no longer part of the list of official plugins.

See: https://github.com/overhangio/tutor-contrib-richie/pull/5


BbrSofiane (approved):
👍🏿

PR #725, opened 2022-09-30
fix: edx-platform requirement installation

regisb:
The local requirements files does not exist since local requirements were all removed from the edx-platform repo. As a consequence, the nightly build was broken.


regisb:
FYI @kdmccormick @Carlos-Muniz

PR #727, opened 2022-09-30
Remove warning about building tutor-mfe on ARM systems

michaelwheeler:
It looks like this was resolved in https://github.com/overhangio/tutor-mfe/pull/42


regisb (approved):

regisb:
I'm not able to confirm that tutor works correctly on arm64, but since no one is complaining about the mfe plugin, then I guess that we can remove the warning from the docs.

PR #731, opened 2022-10-06
WIP strongly-typed hooks

regisb:
This PR aims at achieving two things:

1. Refine the generic type of actions and filters. This is possible now that the upstream mypy bugs have been resolved.
2. Implement strongly-typed named actions and filters in consts.py, to make sure that they are always called with the right arguments. Unfortunately, this does not work, yet.

I'm opening this PR just for the sake of leaving a trace of my research. Unfortunately the proposed implementation does not work in Python < 3.10, so we'll have to wait until Python 3.9 EOL to get this to work... in [October 2025 ](https://endoflife.date/python) :crying_cat_face:

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -26,7 +26,7 @@ kubernetes==18.20.0
     # via -r requirements/base.in
 markupsafe==2.0.1
     # via jinja2
-mypy==0.942
+mypy==0.982
     # via -r requirements/base.in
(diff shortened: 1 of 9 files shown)


regisb:
FYI @kdmccormick.

kdmccormick:
I created https://github.com/overhangio/2u-tutor-adoption/issues/121 to track this.

PR #733, opened 2022-10-10
feat: strongly typed hooks

regisb:
Now that the mypy bugs have been resolved, we are able to define more precisely and cleanly the types of Actions and Filters.

Moreover, can now strongly type named actions and hooks (in consts.py). With such a strong typing, we get early alerts of hooks called with incorrect arguments, which is nothing short of awesome :)

This change breaks the hooks API by removing the `context=...` argument. The reason for that is that we cannot insert arbitrary arguments between `P.args, P.kwargs`: https://peps.python.org/pep-0612/#the-components-of-a-paramspec

> A function declared as def inner(a: A, b: B, *args: P.args, **kwargs:
> P.kwargs) -> R has type Callable[Concatenate[A, B, P], R]. Placing
> keyword-only parameters between the *args and **kwargs is forbidden.

Getting the documentation to build in nitpicky mode is quite difficult... We need to add `nitpick_ignore` to the docs conf.py, otherwise sphinx complains about many missing class references. This, despite upgrading almost all doc requirements (except docutils).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 ## Unreleased
 
+- 💥[Feature] Strong typing of action and filter hooks: this allows us to detect incorrect calls to `actions.add` or `filters.add` early. Strong typing forces us to break the `do` and `apply` API by removing the `context` named argument. Developers should replace `do(context=...)` by `do_from_context(..., )` (and similar for `apply`).
+
 ## v14.1.2 (2022-11-02)
 
 - [Security] Fix edx-platform XSS vulnerability on "next" parameter. (by @regisb)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,13 +23,22 @@ extensions = []
 templates_path = ["_templates"]
 source_suffix = ".rst"
 master_doc = "index"
-language = None
+language = "en"
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 pygments_style = None
 
 # Autodocumentation of modules
 extensions.append("sphinx.ext.autodoc")
 autodoc_typehints = "description"
+# For the life of me I can't get the docs to compile in nitpicky mode without these
+# ignore statements. You are most welcome to try and remove them.
+nitpick_ignore = [
+    ("py:class", "Config"),
+    ("py:class", "click.Command"),
+    ("py:class", "tutor.hooks.filters.P"),
+    ("py:class", "tutor.hooks.filters.T"),
+    ("py:class", "tutor.hooks.actions.P"),
+]
 
 # -- Sphinx-Click configuration
 # https://sphinx-click.readthedocs.io/
--- a/tutor/env.py
+++ b/tutor/env.py
@@ -42,13 +42,13 @@ def _prepare_environment() -> None:
             ("long_to_base64", utils.long_to_base64),
             ("random_string", utils.random_string),
             ("reverse_host", utils.reverse_host),
+            ("rsa_import_key", utils.rsa_import_key),
             ("rsa_private_key", utils.rsa_private_key),
         ],
     )
     # Template variables
     hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
         [
-            ("rsa_import_key", utils.rsa_import_key),
             ("HOST_USER_ID", utils.get_user_id()),
             ("TUTOR_APP", __app__.replace("-", "_")),
             ("TUTOR_VERSION", __version__),
@@ -77,9 +77,7 @@ class Renderer:
         self.environment = JinjaEnvironment(self.template_roots)
 
         # Filters
-        plugin_filters: t.Iterator[
-            t.Tuple[str, JinjaFilter]
-        ] = hooks.Filters.ENV_TEMPLATE_FILTERS.iterate()
+        plugin_filters = hooks.Filters.ENV_TEMPLATE_FILTERS.iterate()
         for name, func in plugin_filters:
             if name in self.environment.filters:
                 fmt.echo_alert(f"Found conflicting template filters named '{name}'")
@@ -87,10 +85,7 @@ class Renderer:
         self.environment.filters["walk_templates"] = self.walk_templates
 
         # Globals
-        plugin_globals: t.Iterator[
-            t.Tuple[str, JinjaFilter]
-        ] = hooks.Filters.ENV_TEMPLATE_VARIABLES.iterate()
-        for name, value in plugin_globals:
+        for name, value in hooks.Filters.ENV_TEMPLATE_VARIABLES.iterate():
             if name in self.environment.globals:
                 fmt.echo_alert(f"Found conflicting template variables named '{name}'")
             self.environment.globals[name] = value
@@ -220,12 +215,10 @@ def is_rendered(path: str) -> bool:
     If the path matches an include pattern, it is rendered. If not and it matches an
     ignore pattern, it is not rendered. By default, all files are rendered.
     """
-    include_patterns: t.Iterator[str] = hooks.Filters.ENV_PATTERNS_INCLUDE.iterate()
-    for include_pattern in include_patterns:
+    for include_pattern in hooks.Filters.ENV_PATTERNS_INCLUDE.iterate():
         if re.match(include_pattern, path):
             return True
-    ignore_patterns: t.Iterator[str] = hooks.Filters.ENV_PATTERNS_IGNORE.iterate()
-    for ignore_pattern in ignore_patterns:
+    for ignore_pattern in hooks.Filters.ENV_PATTERNS_IGNORE.iterate():
         if re.match(ignore_pattern, path):
             return False
     return True
@@ -255,10 +248,7 @@ def save(root: str, config: Config) -> None:
     Save the full environment, including version information.
     """
     root_env = pathjoin(root)
-    targets: t.Iterator[
-        t.Tuple[str, str]
-    ] = hooks.Filters.ENV_TEMPLATE_TARGETS.iterate()
-    for src, dst in targets:
+    for src, dst in hooks.Filters.ENV_TEMPLATE_TARGETS.iterate():
         save_all_from(src, os.path.join(root_env, dst), config)
 
     upgrade_obsolete(root)
@@ -458,8 +448,8 @@ def _delete_plugin_templates(plugin: str, root: str, _config: Config) -> None:
     """
     Delete plugin env files on unload.
     """
-    targets: t.Iterator[t.Tuple[str, str]] = hooks.Filters.ENV_TEMPLATE_TARGETS.iterate(
-        context=hooks.Contexts.APP(plugin).name
+    targets = hooks.Filters.ENV_TEMPLATE_TARGETS.iterate_from_context(
+        hooks.Contexts.APP(plugin).name
     )
     for src, dst in targets:
         path = pathjoin(root, dst.replace("/", os.sep), src.replace("/", os.sep))
--- a/tutor/hooks/actions.py
+++ b/tutor/hooks/actions.py
@@ -4,19 +4,22 @@ __license__ = "Apache 2.0"
 import sys
 import typing as t
 
+from typing_extensions import ParamSpec
+
 from .contexts import Contextualized
 
-# Similarly to CallableFilter, it should be possible to refine the definition of
-# CallableAction in the future.
-CallableAction = t.Callable[..., None]
+P = ParamSpec("P")
+# Similarly to CallableFilter, it should be possible to create a CallableAction alias in
+# the future.
+# CallableAction = t.Callable[P, None]
 
 DEFAULT_PRIORITY = 10
 
 
-class ActionCallback(Contextualized):
+class ActionCallback(Contextualized, t.Generic[P]):
     def __init__(
         self,
-        func: CallableAction,
+        func: t.Callable[P, None],
         priority: t.Optional[int] = None,
     ):
         super().__init__()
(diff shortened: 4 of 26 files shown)


regisb:
FYI @kdmccormick this PR is for solving https://github.com/overhangio/2u-tutor-adoption/issues/121

It was more difficult than I thought to implement this... In particular, getting the docs to compile was challenging. I'd appreciate it if a sphinx/autodoc guru could have a look at my `conf.py` to try and get rid of the `nitpick_ignore` entries.

kdmccormick:
Nice. After I put [this plugin](https://github.com/overhangio/2u-tutor-adoption/issues/77#issuecomment-1276217825) up for review, I'll take a look at this.

tutor/hooks/consts.py
```diff
@@ -171,21 +187,27 @@ def your_filter(items):
     #:      rendered at runtime with the user configuration. Thus, the image tag could
     #:      be ``"{{ DOCKER_REGISTRY }}/myimage:{{ TUTOR_VERSION }}"``.
     #:    - ``args`` is a list of arguments that will be passed to ``docker build ...``.
-    #: :parameter dict config: user configuration.
-    IMAGES_BUILD = filters.get("images:build")
+    #: :parameter Config config: user configuration.
+    IMAGES_BUILD: "Filter[t.List[t.Tuple[str, t.Tuple[str, ...], str, t.Tuple[str, ...]]], [Config]]" = filters.get(
```
kdmccormick, line 191: We could put `from __future__ import annotations` at the top of this module, which enables deferred annotation evaluation for Python 3.7+ and would allow us to drop these quotation marks.
```suggestion
    IMAGES_BUILD: Filter[t.List[t.Tuple[str, t.Tuple[str, ...], str, t.Tuple[str, ...]]], [Config]] = filters.get(
```
Even better, given that we're requiring Python 3.9+ now for type checking, we could start subscripting types directly instead of using the `typing` module:
```suggestion
    IMAGES_BUILD: Filter[list[tuple[str, tuple[str, ...], str, tuple[str, ...]]], [Config]]" = filters.get(
```
I'm also happy make that second refactoring in a separate PR, since it'd be a pretty big diff.

tutor/hooks/consts.py
```diff
@@ -146,15 +160,17 @@ def your_filter(items):
     #: :parameter str name: basename of the host-mounted folder. In the example above,
     #:   this is "edx-platform". When implementing this filter you should check this name to
     #:   conditionnally add mounts.
-    COMPOSE_MOUNTS = filters.get("compose:mounts")
+    COMPOSE_MOUNTS: "Filter[t.List[t.Tuple[str, str]], [str]]" = filters.get(
+        "compose:mounts"
+    )
```
kdmccormick, line 165: I think using NamedTuples here could really improve the readability of some of these definitions. It would be backwards-compatible, since NamedTuples are can be used as ordinary tuples.

```suggestion
    class ComposeMount(t.NamedTuple):
        service: str
        path: str

    COMPOSE_MOUNTS: Filter[list[ComposeMount], [str]] = filters.get(
        "compose:mounts"
    )
```

Could be a future PR, of course.

kdmccormick (commented):
This is great! Here are my thoughts just from looking through the diff.

I'll do another pass once I manage to fix my Python 3.9 installation and can run mypy & the docs build.

docs/conf.py
```diff
@@ -23,13 +23,27 @@
...
+    ("py:class", "Action"),
+    ("py:class", "ActionTemplate"),
+    ("py:class", "Filter"),
+    ("py:class", "t.List"),
+    ("py:class", "t.Tuple"),
+    ("py:class", "t.Any"),
+    ("py:class", "tutor.hooks.filters.P"),
+    ("py:class", "tutor.hooks.filters.T"),
+    ("py:class", "tutor.hooks.actions.P"),
+]
```
kdmccormick, line 46: @regisb I banged my head against this for a while:

* With the PR as-is, you can remove `Action`, `Filter`, and `t.Any` from the ignore list.
* In hooks/consts.py, if you write `from typing import List, Tuple` instead of `import typing as t` (and update references accordingly), then the warnings for `t.List` and `t.Tuple` go away.
  * Alternatively, the docs build fine if you use `from __future__ import annotations` and then replace the types with the built-ins `list` and `tuple`.
* I could not figure out any way to resolve the warnings for `ActionTemplate`, `T`, or either `P`.

My non-expert guess is that Sphinx is getting confused by one or both of these things:
* things that are *types* but aren't *classes* (such as `Config`, `List`, `P`, etc.)
* things that use more than one canonical import path, for example:
  * `typing.List` being renamed to `t.List`
  * `tutor.hooks.consts.Actions` being re-exported as `tutor.hooks.Actions`

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ Every user-facing change should have an entry in this changelog. Please respect
 
 ## Unreleased
 
+- 💥[Feature] Strong typing of action and filter hooks: this allows us to detect incorrect calls to `actions.add` or `filters.add` early. Strong typing forces us to break the `do` and `apply` API by removing the `context` named argument. Developers should replace `do(context=...)` by `do_from_context(..., )` (and similar for `apply`).
```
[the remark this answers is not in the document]
regisb: Wait, what? Type checking is definitely supposed to work with 3.8. I'm able to run `make test` with my local 3.8.10 virtualenv.

tutor/hooks/consts.py
regisb: OK this is very strange. `from __future__ import annotations` is already present at the top of this file. I can't remember at all why I added it in the first place. You are absolutely right that this future module allows us to remove the quotation marks. I had not realised that.

So, great news! We can simplify the type definitions.

tutor/hooks/consts.py
```diff
@@ -171,21 +187,27 @@ def your_filter(items):
...
+    IMAGES_PULL: "Filter[t.List[t.Tuple[str, str]], [Config]]" = filters.get(
+        "images:pull"
+    )
 
     #: List of images to be pushed when we run ``tutor images push ...``.
     #: Parameters are the same as for :py:data:`IMAGES_PULL`.
-    IMAGES_PUSH = filters.get("images:push")
+    IMAGES_PUSH: "Filter[t.List[t.Tuple[str, str]], [Config]]" = filters.get(
+        "images:push"
+    )
```
regisb, line 210: I do not know why I didn't bother adding type definitions below this line. I guess I must have been tired. I'll add the definitions now.

docs/conf.py
regisb: > Alternatively, the docs build fine if you use from __future__ import annotations and then replace the types with the built-ins list and tuple.

OK this is really cool. I'm adding the built-ins now. We are only left with a couple nitpick-ignored classes now.

CHANGELOG.md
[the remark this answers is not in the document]
kdmccormick: Huh, for some reason I thought the generic param specs required 3.9. Nevermind!

tutor/hooks/consts.py
regisb: Let's keep this for a future PR.

tutor/env.py
```diff
@@ -77,20 +77,15 @@ def __init__(self, config: t.Optional[Config] = None):
...
                 fmt.echo_alert(f"Found conflicting template filters named '{name}'")
             self.environment.filters[name] = func
         self.environment.filters["walk_templates"] = self.walk_templates
 
         # Globals
-        plugin_globals: t.Iterator[
-            t.Tuple[str, JinjaFilter]
-        ] = hooks.Filters.ENV_TEMPLATE_VARIABLES.iterate()
-        for name, value in plugin_globals:
+        for name, value in hooks.Filters.ENV_TEMPLATE_VARIABLES.iterate():
```
kdmccormick, line 88: Nice to see how you were able to simplify the filter usages now that the declarations are typed.

tutor/hooks/actions.py
```diff
@@ -4,58 +4,83 @@
...
+from typing_extensions import ParamSpec
+
 from .contexts import Contextualized
 
-# Similarly to CallableFilter, it should be possible to refine the definition of
-# CallableAction in the future.
-CallableAction = t.Callable[..., None]
+P = ParamSpec("P")
+# Similarly to CallableFilter, it should be possible to create a CallableAction alias in
+# the future.
```
kdmccormick, line 13: nit: If "in the future" means "after Python 3.X is end-of-life", and if you know what X is, then you might just note that here for future developers.

tutor/hooks/actions.py
```diff
@@ -4,58 +4,83 @@
...
-            self.func(*args, **kwargs)
+        self.func(*args, **kwargs)
 
 
-class Action:
+class Action(t.Generic[P]):
     """
-    Each action is associated to a name and a list of callbacks, sorted by
-    priority.
+    Action hooks are callbacks that are triggered independently from one another.
```
kdmccormick, line 39: An action has callbacks, but the action itself isn't a callback.
```suggestion
    Action hooks are callbacks that have triggered independently from one another.
```
Otherwise, great improvement on the docstring.

tutor/hooks/actions.py
```diff
@@ -4,58 +4,83 @@
...
+    1. Create an action with method :py:meth:`get` (or function :py:func:`get`).
+    2. Add callbacks with method :py:meth:`add` (or function :py:func:`add`).
+    3. Call the action callbacks with method :py:meth:`do` (or function :py:func:`do`).
+
+    The `P` type parameter of the Action class corresponds to the expected signature of
+    the action callbacks. For instance, `Action[[str, int]]` means that the action
+    callbacks are expected to take two arguments: one string and one integer.
+
+    This strong typing makes it easier to quickly check whether hooks are implemented
+    correctly by plugin developers.
```
kdmccormick, line 55: It's obvious to me what you mean here, but newer Tutor devs might be tripped up if we're not consistent with our terminology.
```suggestion
    This strong typing makes it easier for plugin developers to quickly check whether
    they are adding and calling action callbacks correctly.
```

tutor/hooks/filters.py
```diff
@@ -4,82 +4,90 @@
...
-        return value
+    def apply(self, value: T, *args: P.args, **kwargs: P.kwargs) -> T:
+        return self.func(value, *args, **kwargs)
 
 
-class Filter:
+class Filter(t.Generic[T, P]):
     """
-    Each filter is associated to a name and a list of callbacks.
+    Filter hooks are callbacks that are triggered as a chain.
```
kdmccormick, line 32: Same comment
```suggestion
    Filter hooks have callbacks that are triggered as a chain.
```

tutor/hooks/filters.py
```diff
@@ -4,82 +4,90 @@
...
+    The `T` and `P` type parameters of the Filter class correspond to the expected
+    signature of the filter callbacks. `T` is the type of the first argument (and thus
+    the return value type as well) and `P` is the signature of the other arguments.
+
+    For instance, `Filter[str, [int]]` means that the filter callbacks are expected to
+    take two arguments: one string and one integer. Each callback must then return a
+    string.
+
+    This strong typing makes it easier to quickly check whether hooks are implemented
+    correctly by plugin developers.
```
kdmccormick, line 55: Same comment
```suggestion
    This strong typing makes it easier for plugin developers to quickly check whether
    they are adding and calling filter callbacks correctly.
```

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -122,7 +122,7 @@
 # Allow the platform to include itself in an iframe
 X_FRAME_OPTIONS = "SAMEORIGIN"
 
-{% set jwt_rsa_key = rsa_import_key(JWT_RSA_PRIVATE_KEY) %}
+{% set jwt_rsa_key | rsa_import_key %}{{ JWT_RSA_PRIVATE_KEY }}{% endset %}
```
kdmccormick, line 125: I don't know what's going on here but I'll trust you on it.

tutor/hooks/consts.py
```diff
@@ -113,19 +124,23 @@ def your_filter(items):
     #:     - ``path`` is a tuple that corresponds to a template relative path.
     #:       Example: ``("myplugin", "hooks", "myservice", "pre-init")`` (see:py:data:`IMAGES_BUILD`).
     #:       The command to execute will be read from that template, after it is rendered.
-    COMMANDS_INIT = filters.get("commands:init")
+    COMMANDS_INIT: Filter[list[tuple[str, tuple[str, ...]]], str] = filters.get(
```
kdmccormick, line 127: Subscripting the built-ins reads much better... @regisb are you interested in having this change be made across the codebase? If so I can create an issue for it. It'd be a `good-first-issue`.

kdmccormick (approved):
A few suggestions for you to take or leave. Looks great! 🚀

kdmccormick:
Ah, one other thing @regisb , I haven't been able to get Python 3.7 set up on my machine. I'm assuming you were still able to run unit tests under Python 3.7 with these changes.

regisb:
> I'm assuming you were still able to run unit tests under Python 3.7 with these changes.

I only just did and realised that positional-only parameters do not work with Python 3.7:

```
$ make test-lint 
pylint --errors-only --enable=unused-import,unused-argument --ignore=templates --ignore=docs/_ext ./tutor ./tests ./bin ./docs
************* Module tutor.hooks.actions
tutor/hooks/actions.py:113:9: E0001: Parsing failed: 'invalid syntax (<unknown>, line 113)' (syntax-error)
************* Module tutor.hooks.filters
tutor/hooks/filters.py:147:32: E0001: Parsing failed: 'invalid syntax (<unknown>, line 147)' (syntax-error)
```

Caused by the following pieces of code:

```
    def add_items(self: "Filter[t.List[E], P]", items: t.List[E]) -> None:
        @self.add()
        def callback(
            values: t.List[E], /, *_args: P.args, **_kwargs: P.kwargs # <------ the slash / is a syntax error in python 3.7
        ) -> t.List[E]:
            return values + items
```

(and similar in actions.py)

Positional-only arguments were only introduced in python 3.8 by PEP 570: https://peps.python.org/pep-0570/ As far as I know there is no `from __future__ import ...` statement to backport this feature to python 3.7.

The problem is that if we get rid of this slash then we end up with a mypy error:

```
$ make test-types 
mypy --exclude=templates --ignore-missing-imports --strict ./tutor ./tests ./bin ./docs
tutor/hooks/filters.py:145: error: Argument 1 has incompatible type "Callable[[Arg(List[E], 'values'), **P], List[E]]"; expected "Callable[[List[E], **P], List[E]]"  [arg-type]
tutor/hooks/filters.py:145: note: This is likely because "callback" has named arguments: "values". Consider marking them positional-only
Found 1 error in 1 file (checked 63 source files)
```

(Gotta love the irony of the mypy recommendation: "Consider marking them positional-only")

I'll address this issue by adding a `# type: ignore`. This is unfortunate but I could not find another solution.

tutor/hooks/consts.py
regisb: Yes, absolutely.

kdmccormick:
@regisb Bummer. I fiddled with that line a bit and couldn't find a solution either. Fortunately it seems like the impact is localized to the definition of `add_items`.

tutor/templates/apps/openedx/settings/partials/common_all.py
regisb: This is actually one of the few occurrences where typing highlighted a bug. `rsa_import_key` was considered as a Jinja2 variable, but variables are not expected to be callables. So we moved it to the list of Jinja2 filters instead, but that means that we have to use it slightly differently: https://jinja.palletsprojects.com/en/3.0.x/templates/#block-assignments

tutor/hooks/actions.py
regisb: It's actually independent of Python version, and only once the corresponding mypy issue (documented in filters.py) will be resolved.

tutor/templates/apps/openedx/settings/partials/common_all.py
kdmccormick: Got it. Cool!

PR #735, opened 2022-10-18
feat: use encode gzip for caddy

ghassanmas:
Compressing assests would lead to readuce transfer size.
  As testing with frontend-app-learning/Olive, the network traffic
  before was about ~4MB, after this it became ~1MB.

  This change was suggested by Google Lighthouse[1], there are of
  course more suggestion but this was one the easiest and one of most
  impactful.

  Also check orignal PR overhangio/tutor-mfe/pull/64 for more
  info.
  [1]: https://web.dev/uses-text-compression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Every user-facing change should have an entry in this changelog. Please respect 
 
 ## Unreleased
 
+- [Improvement] use caddy `encode gzip` directive to optimize transfer size (by @ghassanmas)
+
 ## v14.1.0 (2022-10-10)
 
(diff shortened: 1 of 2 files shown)


regisb (changes requested):
Changelog!

regisb (changes requested):
PLEASE TEST YOUR CHANGES! ALWAYS! this is breaking the Caddy config. The `encode gzip` directive should not be inside `reverse_proxy`.

ghassanmas:
Sorry about that, I was too confident!.
Now tha last change is tested on production and also using caddy cli, i.e. `caddy validate --config env/apps/caddy/Caddyfile`

regisb (approved):

PR #736, opened 2022-10-18
depr: get rid of the `local/dev bindmount` commands

regisb:
This command has always been clunky. It is now removed in favour of the `-m/--mount` option.

Close https://github.com/overhangio/2u-tutor-adoption/issues/88
Close https://github.com/overhangio/2u-tutor-adoption/issues/89

FYI @kdmccormick @Carlos-Muniz


kdmccormick (approved):
🆒

Issue #642, opened 2022-04-19
Do we still need to define `*-job` services in docker-compose?


regisb:
It seems to me that we could simply `docker-compose run service sh -e -c "command"`. Why do we have to rely on a convoluted job architecture?

- Jobs don't add anything in terms of environment variables, volume mounting or dependencies.
- If a service must be present in jobs, but not started as a long-running service (e.g: mysql when `RUN_MYSQL=false`) then we can add it to `docker-compose.jobs.yml`, but that should not be the case for most services.
- Services that are `run` bypass the `restart-policy`, which is good for jobs (we don't want to restart jobs).
- We stopped doing `k8s exec` ([here](https://github.com/overhangio/tutor/commit/bce6432d8531555e1a960ff51cbb91e4f64838bc)) because this caused us to bypass environment variables set by the Docker entrypoint. But since we no longer define env variables in the entrypoint, can we do `exec` again?

Such a change would considerably simplify the project architecture.

regisb:
I had a closer look at this issue today. Here's my braindump. Tl;dr: there really is nothing we can do about this issue, so let's close it.

## replacing jobs by exec: ❌

It appears that we cannot substitute jobs by `exec`. The reason is that `exec` requires the container to be up -- which might not be the case if it was not properly initialised. For instance, the lms container will crash if the mysql database was not created. In such a case, we will not be able to exec into the lms container because it will be stuck in a crash-fail loop.

## rename job resources: ❌

So, we need to preserve Kubernetes job resources. We could definitely rename them and get rid of the "-job" suffix. But we would not be able to do the same for docker-compose job services. That's because such jobs would then have the same name as app services. They would then override other service declarations, and that would be very confusing.

## get rid of job services in docker-compose only: ❌

We could just get rid of "*-job" services in docker-compose. Jobs would be executed as "run" commands with existing service declarations. But that would introduce a (potentially important) difference with Kubernetes. So that's probably not a good idea.

## rename jobs for clarification: ❌

It's a little weird that jobs are added to the [CLI_DO_INIT_TASKS](https://docs.tutor.overhang.io/reference/api/hooks/catalog.html#tutor.hooks.Filters.CLI_DO_INIT_TASKS) filter without the "-job" suffix. On the other hand, if it was optional to add the "-job" suffix, then users would be allowed to create job services without the suffix. This would be confusing (see above).

[...]

PR #738, opened 2022-10-21
feat: pluggable `local/dev/k8s do <job>` commands

regisb:
We introduce a new filter to implement custom commands in arbitrary containers.
It becomes easy to write convenient ad-hoc commands that users will
then be able to run either on Kubernetes or locally using a documented CLI.

Pluggable jobs are declared as Click commands and are responsible for
parsing their own arguments. See the new CLI_DO_COMMANDS filter.

Close https://github.com/overhangio/2u-tutor-adoption/issues/75


regisb:
@kdmccormick I'm curious to have your thoughts on this.

kdmccormick:
Nice, I'll review shortly @regisb .

tutor/commands/jobs.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # Init tasks
+    iter_init_tasks: t.Iterator[
+        t.Tuple[str, t.Iterable[str]]
+    ] = hooks.Filters.COMMANDS_INIT.iterate(context=filter_context)
+    for service, path in iter_init_tasks:
+        fmt.echo_info(f"Running init task: {'/'.join(path)}")
+        yield service, env.read_template_file(*path)
+
+    fmt.echo_info("All services initialised.")
```
kdmccormick, line 62: Hm, in the interest of reducing the total number of extension points plugin authors need to consider, I was kinda hoping that we could completely implement initialization within the `CLI_DO_COMMANDS` framework, allowing us to deprecate `COMMANDS_INIT` and `COMMANDS_PRE_INIT` and implement `do init` as an ordinary job with no calls to internal Tutor functions.

I understand why it is not currently this way: the `do` framework is designed to allow plugins to *add new* jobs, not to *extend existing* jobs. I don't really have any suggestions here yet... I'll think about it over the weekend.

tutor/commands/jobs.py
```diff
@@ -0,0 +1,236 @@
...
+
+    The context is important, because it allows us to select the init scripts based on
+    the --limit argument.
+    """
+    with hooks.Contexts.APP("mysql").enter():
+        hooks.Filters.COMMANDS_INIT.add_item(("mysql", ("hooks", "mysql", "init")))
+    with hooks.Contexts.APP("lms").enter():
+        hooks.Filters.COMMANDS_INIT.add_item(("lms", ("hooks", "lms", "init")))
+    with hooks.Contexts.APP("cms").enter():
+        hooks.Filters.COMMANDS_INIT.add_item(("cms", ("hooks", "cms", "init")))
```
kdmccormick, line 37: Since all the updates we've made to the Plugins API, the name and structure of the `hooks/` template directory has stopped making sense.

What about, instead, having a single, flat `init/` directory?

```suggestion
        hooks.Filters.COMMANDS_INIT.add_item(("mysql", ("init", "mysql")))
    with hooks.Contexts.APP("lms").enter():
        hooks.Filters.COMMANDS_INIT.add_item(("lms", ("init", "lms")))
    with hooks.Contexts.APP("cms").enter():
        hooks.Filters.COMMANDS_INIT.add_item(("cms", ("init", "cms")))
```
(I'd also support holding off on these sort of refactorings until the big tutorcore extraction.)

tutor/commands/k8s.py
```diff
@@ -590,12 +544,10 @@ def k8s_namespace(config: Config) -> str:
...
-k8s.add_command(createuser)
-k8s.add_command(importdemocourse)
-k8s.add_command(settheme)
 k8s.add_command(exec_command)
 k8s.add_command(logs)
 k8s.add_command(wait)
 k8s.add_command(upgrade)
 k8s.add_command(apply_command)
 k8s.add_command(status)
+jobs.add_job_commands(do)
```
kdmccormick, line 553: I found that custom commands cannot be added, because this line is run before plugin modules are imported. I was able to fix it by moving this line into a `Actions.PLUGINS_LOADED` callback.

tutor/commands/compose.py
```diff
@@ -295,77 +297,21 @@ def restart(context: BaseComposeContext, services: t.List[str]) -> None:
     context.job_runner(config).docker_compose(*command)
 
 
+@click.group("do")
+def do() -> None:
+    """
+    Run custom jobs in the right containers.
+    """
```
kdmccormick, line 304: I think supporting `--mount` would make `do` much more powerful. For example, if we incorporated quickdev, it'd be great to be able to write something like:
```
tutor dev do --mount=../xblock-drag-and-drop-v2 pip-install-mounts
```
And if we didn't incorporate quickdev, then we'll still want some sort of task to job to make [setting up a dev env easier](https://docs.tutor.overhang.io/dev.html#setting-up-a-development-environment-for-edx-platform), like:
```
tutor dev do --mount=../edx-platform edx-platform-setup
```

CHANGELOG-nightly.md
```diff
@@ -8,6 +8,9 @@ will be backported to the master branch at every major release.
 When backporting changes to master, we should keep only the entries that correspond to user-
 facing changes.
 -->
+- 💥[Feature] Add an extensible `local/dev/k8s do ...` command to trigger custom job commands. These commands are used to run a series of bash scripts in designated containers. Any plugin can add custom jobs thanks to the `CLI_DO_COMMANDS` filter. This causes the following breaking changes:
+    - The "createuser", "settheme", "importdemocourse" commands were all migrated to this new interface. For instance,  `tutor local createuser` was replaced by `tutor local do createuser`.
+    - The `init` command loses its `-m/--mount` option  (which was ignored by `quickstart`).
```
kdmccormick, line 13: I'm not sure I understand this. Launch/quickstart supports mounting, so why wouldn't init?

kdmccormick (changes requested):
I tried this out in a plugin of mine and hit a few issues, but overall I really like the interface. Being able to add proper Click commands that execute commands in containers feels like a huge step in capability for the plugin API.

tutor/commands/jobs.py
kdmccormick: OK, I thought about this more.

I was happy to realize that everything in the new `initialise` function above could basically have also been written in a plugin. There calls to `read_template_file` and `echo_info`, but those are just simple wrappers around standard Python library functions. This is very cool, and it signals to me that `CLI_DO_COMMANDS` is a well-designed interface 👍🏻 

On the other hand, the naming & signature of this corner API feels inconsistent now:
* `CLI_DO_COMMANDS` returns Click commands that yield (service, command template) pairs
* `COMMANDS_INIT` returns (service, command template path) pairs.

The term "command" here is being overloaded between a few different concepts:
1. "command" as in "Click command"
2. "command" as in "a shell command to be run in a container"
3. "command" as in "a (service, shell command) tuple". In some places, we call this a *task*, which I like.

How would you feel about deprecating `COMMANDS_[PRE_]INIT` in favor of new filters:
```python
    #: List of shell commands to be executed during initialization. These commands typically
    #: include database migrations, setting feature flags, etc.
    #:
    #: :parameter list[tuple[str, str]] tasks: list of ``(service, command)`` tasks.
    #:
    #:     - ``service`` is the name of the container in which the task will be executed.
    #:     - ``command`` is a shell command template which will be rendered and executed in ``service``.
    CLI_DO_INIT_TASKS = filters.get("cli:commands:do:init")

    #: ...
    CLI_DO_PRE_INIT_TASKS = filters.get("cli:commands:do:init:pre")
```
(Names up for debate, of course. I wanted to choose names that made it clear that this filter is for adding *tasks* to the existing `do init` command.)

In addition to having different names, these filters have different signatures: they are expected to return `(service, command)` tuples instead of `(service, path)` tuples. This is congruent with the signature of `CLI_DO_COMMANDS`. The mapping from the old filters to the new ones would be pretty straightforward. For example:
```python
    with hooks.Contexts.APP("lms").enter():
        hooks.Filters.COMMANDS_INIT.add_item(("lms", ("hooks", "lms", "init")))
```
into:
```python
    with hooks.Contexts.APP("lms").enter():
        hooks.Filters.CLI_DO_INIT_TASKS.add_item(
            ("lms", env.read_template_file("hooks", "lms", "init"))
        )
```

I realize that this would be a breaking change, and would best be done in a different PR.

tutor/commands/jobs.py
regisb: I think it's fine if we move the init scripts. It should not be a breaking change. I'll move the scripts to `tutor/templates/jobs/init/<service>.sh`.

tutor/commands/k8s.py
regisb: Right. I confess that I did not even take the time to try out the API...

tutor/commands/compose.py
regisb: Right. It took me a very long time to figure out how to properly call `do` subcommands from quickstart, which is why I decided to get rid of the `--mount` option. But it's actually pretty simple to add it back, so I'll do that now.

tutor/commands/jobs.py
regisb: Yes, your proposition makes a lot of sense. A couple precisions:

- Let's add the `CLI_DO_INIT_TASKS` filter.
- Let's keep the `COMMANDS_INIT` and `COMMANDS_PRE_INIT` filters for now. I don't want to force plugin maintainers to upgrade their plugins right away. We can just print a deprecation warning for the moment.
- We don't even need the `CLI_DO_PRE_INIT_TASKS` filter if we implement filter priorities. This is very easy, based on the existing action priority code.

I'll make these changes right now.

tutor/commands/jobs.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+hooks.Filters.CLI_DO_COMMANDS.add_items(
+    [
+        createuser,
+        importdemocourse,
+        initialise,
+        settheme,
+    ]
+)
```
kdmccormick, line 196: Ugh, I just banged my head against my keyboard for half an hour before realizing why this wasn't working:
```python
def myjob():
    return [("lms", "echo hello world")]
hooks.Filters.CLI_DO_COMMANDS.add(myjob)
```
(can you see it?)
Other than the strongly typed hooks PR you have open, I'm not sure what else we would do to save a plugin developer from this kind of mistake.

kdmccormick (approved):
Looks good and works on my machine! The only suggestion I feel strongly about is the "jobs" vs "tasks" one.

tutor/jobs.py
```diff
@@ -1,16 +1,13 @@
...
+from tutor.types import Config
 
 
 class BaseJobRunner:
     """
-    A job runner is responsible for getting a certain task to complete.
+    A job runner is responsible for running bash commands in the right context.
+
+    Commands may be loaded from string or template files. The `run_job` method must be
+    implemented by child classes.
```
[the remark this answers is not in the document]
regisb: I Agree.

tutor/commands/compose.py
```diff
@@ -295,77 +297,16 @@ def restart(context: BaseComposeContext, services: t.List[str]) -> None:
     context.job_runner(config).docker_compose(*command)
 
 
-@click.command(help="Initialise all applications")
-@click.option("-l", "--limit", help="Limit initialisation to this service or plugin")
+@click.group("do")
```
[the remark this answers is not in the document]
regisb: `subcommand_metavar`  is great, but I'm not sure I want to go down that rabbit hole...

OK I just did :sweat_smile: Let's hope it's not making the code too complex.

tutor/commands/jobs.py
regisb: `add_item`? To your credit, yes I must admit it's a little confusing. We'll see whether typed hooks help resolve this situation.

kdmccormick: I think it will.

tutor/commands/compose.py
```diff
@@ -7,13 +7,15 @@
 
 from tutor import config as tutor_config
 from tutor import env as tutor_env
-from tutor import fmt, hooks, jobs, serialize, utils
+from tutor import fmt, hooks, serialize, utils
+from tutor.commands import jobs
 from tutor.commands.context import BaseJobContext
```
kdmccormick, line 12: Can we rename BaseJobContext -> BaseTaskContext ?

tutor/commands/k8s.py
```diff
@@ -46,7 +49,7 @@ def core_api(self):  # type: ignore
         return self._core_api
 
 
-class K8sJobRunner(jobs.BaseJobRunner):
+class K8sTaskRunner(BaseTaskRunner):
```
kdmccormick, line 52: We may want to address the mismatched meaning of "jobs" between Tutor and K8s:
```suggestion
class K8sTaskRunner(BaseTaskRunner):
    """
    Run tasks (bash commands) in Kubernetes-managed services.
    
    Terminology note:
        A Tutor "task" corresponds to a Kubernetes "job".
        In contrast, a "Tutor job" (invoked on the CLI using `tutor k8s do`)
        will trigger one or more Tutor tasks (Kubernetes jobs).
    """
```

tutor/commands/k8s.py
```diff
@@ -339,17 +342,34 @@ def delete(context: K8sContext, yes: bool) -> None:
     )
 
 
+@jobs.do_group
+@click.pass_obj
+def do(context: K8sContext) -> None:
+    """
+    Run custom jobs in the right containers.
```
kdmccormick, line 349: The rest of the command helptexts are written in a singular voice (e.g. `run  --  Run a command in a new container`) so I suggest doing the same here.
```suggestion
    Run a custom job in the right containers.
```

tutor/commands/compose.py
```diff
@@ -295,77 +297,23 @@ def restart(context: BaseComposeContext, services: t.List[str]) -> None:
...
+def do(
+    context: BaseComposeContext, mounts: t.Tuple[t.List[MountParam.MountType]]
 ) -> None:
-    config = tutor_config.load(context.root)
-    runner = context.job_runner(config)
-    domains = domains or jobs.get_all_openedx_domains(config)
-    jobs.set_theme(theme_name, domains, runner)
-
+    """
+    Run custom jobs in the right containers.
```
kdmccormick, line 307: Same suggestion.
```suggestion
    Run custom a job in the right containers.
```

tutor/commands/compose.py
```diff
@@ -79,7 +81,7 @@ def update_docker_compose_tmp(
             docker_compose_jobs_tmp_path,
```
kdmccormick, line 81: Didn't you mention on an issue or PR somewhere that we could phase out the `*.jobs.yml` compose files at some point? Was it contingent on some refactoring?

If not, I think `*.task.yml` and `COMPOSE_LOCAL_TASKS_TMP` would be a more appropriate names for the compose file and filter. I don't want to block this PR any longer, so I'm happy to make a backlog ticket for that if you agree.

kdmccormick (approved):
A few more suggestions, but again, incorporate whatever you like and merge as you see fit.

tutor/commands/k8s.py
regisb: Good point, I improved the docs.

tutor/commands/compose.py
regisb: Not so much an issue, more of an open question that I raised here: https://github.com/overhangio/tutor/issues/642
I don't want to phase out or rename this template or the corresponding patches at this point, as it would break plugins -- I'd rather not break plugins just for a name change. The rename would affect the following files:

- tutor/templates/k8s/jobs.yml
- tutor/templates/dev/docker-compose.jobs.yml
- tutor/templates/local/docker-compose.jobs.yml

And patches:

- local-docker-compose-jobs-services
- dev-docker-compose-jobs-services
- k8s-jobs

I don't think that the name change warrants a new backlog ticket; "jobs" might not be the perfect name but it's close enough.

kdmccormick: Sounds good.

PR #743, opened 2022-11-18
feat: common CORS settings for LMS and CMS

arbrandes:
The ENABLE_CORS_HEADERS feature flag is already true for the LMS. Instead of duplicating it for Studio via yaml settings, make this a common Django setting to both LMS and CMS and all their environments.


tutor/templates/apps/openedx/config/cms.env.yml
```diff
@@ -9,6 +9,7 @@ FEATURES:
   {{ patch("cms-env-features")|indent(2) }}
   CERTIFICATES_HTML_VIEW: true
   PREVIEW_LMS_BASE: "{{ PREVIEW_LMS_HOST }}"
+  ENABLE_CORS_HEADERS: true
```
regisb, line 12: Could we take this opportunity to remove `ENABLE_CORS_FEATURES` from both lms/cms.env.yml and move it to templates/apps/openedx/settings/partials/common_all.py?

    FEATURES["ENABLE_CORS_HEADERS"] = True

I'd like to have as few things as possible in the yml file, for multiple reasons, but in particular because they are not used consistently between dev and prod environment.

arbrandes: Makes sense to me.  Updated.

The only thing, as you'll note, is that we have to duplicate [this section](https://github.com/openedx/edx-platform/blob/master/cms/envs/production.py#L586) because it's executed before the feature flag is actually set to true.

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -160,6 +160,17 @@
 # Enable/Disable some features globally
 FEATURES["ENABLE_DISCUSSION_SERVICE"] = False
 FEATURES["PREVENT_CONCURRENT_LOGINS"] = False
+FEATURES["ENABLE_CORS_HEADERS"] = True
+
+# CORS
+CORS_ALLOW_CREDENTIALS = True
+CORS_ORIGIN_WHITELIST = ENV_TOKENS.get('CORS_ORIGIN_WHITELIST', ())
```
regisb, line 167: Let's simplify this:

```
CORS_ORIGIN_WHITELIST = []
CORS_ORIGIN_ALLOW_ALL = False
CORS_ALLOW_INSECURE = {% if ENABLE_HTTPS %}False{% else %}True{% endif %}
CORS_ALLOW_HEADERS = corsheaders_default_headers + ('use-jwt-cookie',)
```

regisb (commented):
Can you please rebase your single commit on top of master? It looks like you pulled a lot of changes from the nightly branch.

tutor/templates/apps/openedx/settings/partials/common_all.py
arbrandes: Done!

regisb (changes requested):
Almost there! Can you please add an entry to the changelog?

regisb (approved):

PR #749, opened 2022-11-28
fix: pin openedx dev Docker image tag with tutor version

regisb:
When running:

    tutor dev run -m /path/to/edx-platform lms
    pip install -r requirements/edx/development.txt

I realised that I was re-installing packages that should already have been present in the image. The reason for that was that I was running an outdated version of the dev version of the openedx Docker image. This happens because `tutor dev run` does not trigger an image re-build.

We solve this issue by pinning the openedx dev Docker image tag to the current tutor version.


regisb:
I'm curious to have your thoughts here @kdmccormick. Do you think this will break anything?

kdmccormick (commented):
Looks good to me, in theory at least. In a few hours I'll be back on wifi and will test it more throughly.

kdmccormick (approved):
Wonderful improvement, looks great.

How I tested:
```bash
cd ~/openedx/tutor

# check our your branch, which is based on master
git checkout regisb/fix-docker-openedx-dev-version
tutor config save

# confirmed: I can build the dev image for v14.2.1 and run lms
tutor dev dc build lms
tutor dev run lms echo Test

# switch to nightly, plus your patch
git checkout nightly
git cherry-pick regisb/fix-docker-openedx-dev-version
tutor config save

# confirmed: running lms in nightly triggers a rebuild,
# which it didn't before this PR.
tutor dev run lms echo Test

# switch back to your branch, based on master
git checkout regisb/fix-docker-openedx-dev-version
tutor config save

# confirmed: running lms again *DOESN'T* trigger a rebuild,
# since the image I built on master earlier is still in my Docker
# cache.
tutor dev run lms echo Test
```

PR #752, opened 2022-12-06
fix: update problem templates according newer python versions

mariajgrimaldi:
### Description
This PR adds a patch that fixes a syntax error in the Circuit Schematic Builder problem template reported [here](https://github.com/openedx/build-test-release-wg/issues/209) during the olive testing plan.

Please, let me know if this is the right place to add this fix.


mariajgrimaldi:
@regisb: thanks! both work for me :s not sure why. I changed it already :)

regisb:
I think you forgot to `git add` the new changelog entry ;)

regisb (approved):

regisb:
Thanks @mariajgrimaldi!

PR #760, opened 2022-12-13
depr: get rid of `quickstart` command

regisb:
We meant to get rid of `quickstart` in Olive, but somehow we forgot to delete the command.


regisb:
I'm going to merge this now because I'd like to release v15.1.0 before the day is over.

PR #762, opened 2022-12-13
feat: backfill persistent grades during upgrade

regisb:
In Olive, persistent grades are now mandatory. This was already the case in Nutmeg, but it didn't mean that existing, non-persistent grades were migrated to be persistent. This introduces a job that is run whenever a user upgrades from Nutmeg. Alternatively, it can be run manually with:

    tutor local upgrade --from=nutmeg

See relevant documentation: https://openedx.atlassian.net/wiki/spaces/AC/pages/755171487/Migrating+to+Persistent+Grading#The-Persistent-Grades-Backfill
See discussion: https://app.slack.com/client/T02SNA1T6/C049JQZFR5E/thread/C02SNA1U4-1670937183.881709


kdmccormick (approved):
Thanks for the patch.

PR #775, opened 2023-01-06
feat: refactor hooks API for simplification

regisb:
The hooks API had several issues which are summarized in this comment: https://github.com/openedx/wg-developer-experience/issues/125#issuecomment-1313553526

1. "consts" was a bad name
2. "hooks.filters" and "hooks.Filters" could easily be confused
3. docs made it difficult to understand that plugin developers should use the catalog

To address these issues, we:

1. move "consts.py" to "catalog.py"
2. Remove "hooks.actions", "hooks.filters", "hooks.contexts" from the API.
3. re-organize the docs and give better usage examples in the catalog.

This change is a partial fix for https://github.com/openedx/wg-developer-experience/issues/125


kdmccormick:
At a glance, I like the look of this. I can take a closer look next week.

tutor/hooks/filters.py
```diff
@@ -79,6 +79,13 @@ def add(
     ) -> t.Callable[
         [t.Callable[Concatenate[T, P], T]], t.Callable[Concatenate[T, P], T]
     ]:
+        """
+        Add a callback to the filter.
+
+        Callbacks are added by increasing priority. Highest priority score are called
+        last.
+        """
```
kdmccormick, line 87: Rather than write a new docstring for `Filter.add`, why not strike the docstring from module-level `add()` and adapt its contents here?

```suggestion
        """
        Decorator to add a callback function to be applied to this filter.
        :param int priority: optional order in which the filter callbacks are called. Higher
            values mean that they will be performed later. The default value is
            ``priorities.DEFAULT`` (10). Filters that should be called last should have a
            priority of 100.
        The return value of each filter function callback will be passed as the first argument to the next one.
        Usage::
            from tutor import hooks
            @hooks.Filters.MY_FILTER.add()
            def my_func(value, some_other_arg):
                # Do something with `value`
                ...
                return value
            # After filters have been created, the result of calling all filter callbacks is obtained by running:
            hooks.Filters.MY_FILTER.apply(initial_value, some_other_argument_value)
        """
```

docs/tutorials/plugin.rst
```diff
@@ -45,7 +45,7 @@ Modifying existing files with patches
 
 We'll start by modifying some of our Open edX settings files. It's a frequent requirement to modify the ``FEATURES`` setting from the LMS or the CMS in edx-platform. In the legacy native installation, this was done by modifying the ``lms.env.yml`` and ``cms.env.yml`` files. Here we'll modify the Python setting files that define the edx-platform configuration. To achieve that we'll make use of two concepts from the Tutor API: :ref:`patches` and :ref:`filters`.
```
kdmccormick, line 46: Somewhere in this tutorial, we ought to link to the new Hooks Catalog landing page you wrote. I also think it would be good to link to the Hooks Catalog from the [Plugins Introduction](https://docs.tutor.overhang.io/plugins/intro.html).

I think it may not immediately click to new plugin devs that the V1 Plugin API is just a CLI around the Hooks API. Adding cross-links to the docs might make that connection more obvious.

kdmccormick (commented):
I have some suggestions, but even without applying any them, I think this is a good improvement. Thanks for all your work here!

docs/reference/api/hooks/catalog.rst
```diff
@@ -0,0 +1,27 @@
+.. _hooks_catalog:
+
+=============
+Hooks catalog
+=============
```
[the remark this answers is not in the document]
regisb: I agree. To that end I'll move the docs from the `tutor.hooks.*.do/add/clear/...` functions to their respective methods. I'll also implement your other suggestions.

tutor/hooks/__init__.py
```diff
@@ -3,14 +3,16 @@
 
 import typing as t
 
+from . import actions as _actions
+from . import filters as _filters
```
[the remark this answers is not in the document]
regisb: Ah, yes. You mean that this does not resolve the confusion around `hooks.Actions` vs `hooks.actions`, right? In your [other comment](https://github.com/openedx/wg-developer-experience/issues/125#issuecomment-1369973269) you mentioned that we should flatten the actions.py/filters.py modules into `__init__.py`. But this would again cause confusion between `hooks.Actions` and `hooks.Action`, right?

My original sin here was that I wanted to expose the hooks catalog in `tutor.hooks`. I think that we should have gone with:

    import tutor.hooks.catalog as hooks

If you agree, then we can make this change in the plugin cookiecutter after this PR is merged. Then in Palm we can get rid of the hooks/__init__.py module altogether.

tutor/hooks/filters.py
regisb: Yes, I agree. I moved all docstrings from the utility functions to their methods.

docs/tutorials/plugin.rst
regisb: I agree. I added a few words to the docs page. Let me know if you think this is sufficient.

regisb:
Please check out my changes in the latest commit @kdmccormick: https://github.com/overhangio/tutor/pull/775/commits/9522785e21ad196adfbbbc7b8046c34206a2e69c

tutor/hooks/__init__.py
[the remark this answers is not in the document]
kdmccormick: Hm... stepping back, we're trying to balance three things here: clarity of the API, brevity of the import paths, and the backwards-compatibility of this this change. Let's look at the options we have for the most important public API elements.

(1) Currently, we have:
```
tutor.hooks.Actions
tutor.hooks.Filters
tutor.hooks.priorities.Priorities
tutor.hooks.actions.Action
tutor.hooks.filters.Filter
tutor.hooks.contexts.Context
```

(2) My comment suggested:
```
tutor.hooks.Actions
tutor.hooks.Filters
tutor.hooks.Priorities
tutor.hooks.Action
tutor.hooks.Filter
tutor.hooks.Context
```

(3) Your comment suggests:
```
tutor.hooks.catalog.Actions
tutor.hooks.catalog.Filters
tutor.hooks.priorities.Priorities
tutor.hooks.actions.Action
tutor.hooks.filters.Filter
tutor.hooks.contexts.Context
```

I think (2) wins on brevity, but (3) is clearer in that it separates the "catalog" classes (Actions/Filters) out from the "type" classes (Action/Filter/Context). In either case, we changing a handful of import paths, and could maintain backwards-compatibility until Palm.

I'll throw one last suggestion in: (4)
```
tutor.hooks.Actions
tutor.hooks.Filters
tutor.hooks.Priorities
tutor.hooks.types.Actions
tutor.hooks.types.Filter
tutor.hooks.types.Context
```

I like this one because it separates the catalog classes from the type classes, but the imports for the catalog classes (including Priorities) are shorter. As for backwards-compatibility, we are changing 3 import paths, but it would be easy to maintain compatibility until Palm: we'd have wrapper modules (actions.py, filters.py, contexts.py, priorities.py), each of which could raise a warning a la _Importing from ... is deprecated, please import ... instead_ and then re-export their old items from their canonical locations.

Let me know what you think. I'd like to make this decision carefully--this is our one shot to restructure the V1 API!  😄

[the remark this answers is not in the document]
regisb: Your suggestion is better, but I don't like "types". I suggest that we start creating a new "core" module:

```
tutor.hooks.Actions
tutor.hooks.Filters
tutor.hooks.Priorities
tutor.core.hooks.Action
tutor.core.hooks.Filter
tutor.core.hooks.Context
```

After all, hook classes are very much at the "core" of Tutor now. Just looking at the imports we can see that they don't pull anything from outside the hooks framework, and are thus completely unrelated to Open edX.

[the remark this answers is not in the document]
kdmccormick: Are you thinking that tutor.core would be the eventual destination for other "tutorcore" things too, like plugins/ and config.py?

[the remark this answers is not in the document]
regisb: Yes, exactly.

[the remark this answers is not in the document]
kdmccormick: Sounds good to me then 👍

Priorities is funny--I could see it go in either place. It feels like a part of the "catalog", but also, there's nothing openedx-specific about the priority constants.

regisb:
Sphinx has given me hell with the new typing annotations... It's been crazy hard to get the docs to build across different Python version, to the point that I considered switching to a different documentation backend entirely.
At this point, the docs no longer build on Python 3.10. Here are the errors that I'm getting:

    /home/data/regis/projets/overhang/repos/overhang/tutor/tutor/hooks/catalog.py:docstring of tutor.hooks.Actions.COMPOSE_PROJECT_STARTED:1: WARNING: py:class reference target not found: tutor.core.hooks.actions.Action[(<class 'str'>, typing.Dict[str, typing.Union[str, float, NoneType, bool, typing.List[str], typing.List[typing.Any], typing.Dict[str, typing.Any], typing.Dict[typing.Any, typing.Any]]], <class 'str'>)]
    /home/data/regis/projets/overhang/repos/overhang/tutor/tutor/hooks/catalog.py:docstring of tutor.hooks.Actions.DO_JOB:1: WARNING: py:class reference target not found: tutor.core.hooks.actions.Action[(<class 'str'>, typing.Any)]
    /home/data/regis/projets/overhang/repos/overhang/tutor/tutor/hooks/catalog.py:docstring of tutor.hooks.Actions.PLUGIN_UNLOADED:1: WARNING: py:class reference target not found: tutor.core.hooks.actions.Action[(<class 'str'>, <class 'str'>, typing.Dict[str, typing.Union[str, float, NoneType, bool, typing.List[str], typing.List[typing.Any], typing.Dict[str, typing.Any], typing.Dict[typing.Any, typing.Any]]])]
        /home/data/regis/projets/overhang/repos/overhang/tutor/tutor/hooks/catalog.py:docstring of tutor.hooks.Actions.PROJECT_ROOT_READY:1: WARNING: py:class reference target not found: tutor.core.hooks.actions.Action[(<class 'str'>,)]
    /home/data/regis/projets/overhang/repos/overhang/tutor/tutor/hooks/catalog.py:docstring of tutor.hooks.Filters.COMPOSE_MOUNTS:1: WARNING: py:class reference target not found: tutor.core.hooks.filters.Filter[list[tuple[str, str]], (<class 'str'>,)]

For the record, these errors come from the recent [type annotations PR](https://github.com/overhangio/tutor/pull/781), not from this API simplification PR. 

If anyone wants to give a stab at building the docs on Python 3.10, they are more than welcome.

regisb:
@kdmccormick I know that this PR has grown to a humongous scale and I hate to ask this from you, but could you please have another look at this PR? If not it's OK, I think I can merge as it is.

regisb:
> can we choose 1 specific Python version to support for Tutor development (docs, linting, typechecking)?

I guess it would make sense, yes. The problem is that I'd rather default to the newest stable Python version (3.11) but the docs don't build on 3.11 either. Next step will probably be to ask a question on stackoverflow...

kdmccormick:
Sorry that I did not get to re-reviewing this. It's at the top of my priority list of Monday.

changelog.d/20230106_190620_regis_hooks_api.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] Simplify the hooks API. Plugin developers who were previously using `hooks.actions`, `hooks.filters` or `hooks.contexts` should now import these modules explicitely. (by @regisb)
```
kdmccormick, line 1: I'm not sure what the changelog syntax allows, but what do think about enumerating the breaking changes here?

```suggestion
- 💥[Feature] Simplify the hooks API. The modules `tutor.hooks.actions`, `tutor.hooks.filters`, and `tutor.hooks.contexts` are no longer part of the API.
  * Instead of `tutor.hooks.actions.get("some:action")` , use `tutor.hooks.Action.SOME_ACTION` or `tutor.core.hooks.Actions.get("some:action")`. The former is highly preferred for all predefined Tutor Actions, as its usage can be checked by pylint and mypy.
  * Instead of `tutor.hooks.filters.get("some:filter")`, use `tutor.hooks.Filter.SOME_FILTER` or `tutor.core.hooks.Filter.get("some:filter")`. The former is highly preferred for all predefined Tutor , as its usage can be checked by pylint and mypy.
  * Instead of `tutor.hooks.actions.add("some:action")`, use `tutor.hooks.Action.SOME_ACTION.add()`. The same applies to the `do` method.
  * Instead of `tutor.hooks.actions.add("some:action")`, use `tutor.hooks.Action.SOME_ACTION.add()`. The same applies to the `add_item`, `add_items`, `apply`, and `iterate` methods.
  * ... and a couple more bullet points about changes to context-related functions
  (by @regisb)    
```

tutor/hooks/__init__.py
```diff
@@ -3,14 +3,7 @@
 
 import typing as t
 
-# These imports are the hooks API
-from . import actions, contexts, filters, priorities
-from .consts import *
+# The imports that follow are the hooks API
+from tutor.core.hooks import clear_all, contexts, priorities
```
kdmccormick, line 7: It seems to me like `contexts` belongs to the `tutor.core.hooks` API, not here in the`tutor.hooks` catalog. Would you agree?

docs/reference/api/hooks/contexts.rst
```diff
@@ -6,6 +6,6 @@ Contexts
 
 Contexts are a feature of the hook-based extension system in Tutor, which allows us to keep track of which components of the code created which callbacks. Contexts are very much an internal concept that most plugin developers should not have to worry about.
 
-.. autofunction:: tutor.hooks.contexts::enter
+.. autofunction:: tutor.core.hooks.contexts::enter
 
-.. autoclass:: tutor.hooks.contexts.Context
+.. autoclass:: tutor.core.hooks.Context
```
kdmccormick, line 11: This page appears under the catalog docs, so I think we'd want to show the Contexts catalog class, not the Context core type class.

```suggestion
.. autoclass:: tutor.hooks.Contexts
```

kdmccormick (commented):
The API and the docs are looking better and better. Sorry that Sphinx has been such a PITA. For what it's worth, the build is passing on for me on Python 3.9, too.

docs/reference/api/hooks/contexts.rst
regisb: Hmmm not quite. This page will actually appear in the "Hook types" section. The `Contexts` class is indeed documented in catalog.rst.

tutor/hooks/__init__.py
regisb: Hmmm not quite. This page will actually appear in the "Hook types" section. The `Contexts` class is indeed documented in catalog.rst.

regisb: Yes, you are correct.

regisb:
Thanks a lot for the many reviews Kyle! I applied your recommendations and will merge this now.

PR #776, opened 2023-01-07
minor typo fixed

CodeWithEmad:
context -> contexts


sehsanm:
I think previous wording is correct.

CodeWithEmad:
Thanks, @sehsanm. so I'll close this.

Issue #777, opened 2023-01-07
upgrade_from_maple fails


CodeWithEmad:
**Bug description**

Hi @regisb
actually, this is odd that no one came across this issue
when you are not using MySQL container/pod on your local/k8s environment (I'm using AWS RDS so `RUN_MYSQL: false` inside `config.yml`) there will be no deployment for MySQL but in the `upgrade_from_maple` function, we're trying to apply something that doesn't exists.

https://github.com/overhangio/tutor/blob/380786219e8e7af9c429eda8b5e47117d7af6b5e/tutor/commands/upgrade/k8s.py#L123-L128

and running the `upgrade` command raise: `no objects passed to apply`

**How to reproduce**

- having a maple version of tutor 13.3.1
- update tutor with `pip install "tutor[full]==v14.2.3"`
- run `tutor k8s upgrade --from=maple`

**Environment**

OS: Ubuntu
tutor: 14.2.3

**Additional context**

Woman, Life, Freedom

CodeWithEmad:
I like to work on a fix for this issue, if you're agree

PR #778, opened 2023-01-07
fix: start and wait for mysql if RUN_MYSQL was true

CodeWithEmad:
this will fix https://github.com/overhangio/tutor/issues/777
if you want to upgrade from maple to nutmeg and you were hosting your MySQL somewhere like RDS, and RUN_MYSQL is set to false.

running `tutor k8s upgrade --from=maple` will fail since there's no MySQL deployment.
this will check `RUN_MYSQL` value and make sure it's not `false`.

p.s: this is the 777th PR of Tutor which fix the 777th issue. 😄

### DIFF
--- a/tutor/commands/upgrade/k8s.py
+++ b/tutor/commands/upgrade/k8s.py
@@ -120,13 +120,14 @@ def upgrade_from_maple(context: Context, config: Config) -> None:
     # The environment needs to be updated because the backpopulate/backfill commands are from Nutmeg
     tutor_env.save(context.root, config)
 
-    # Start mysql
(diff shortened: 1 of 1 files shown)


regisb (changes requested):
Nice catch. Can you please add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing-to-tutor

CodeWithEmad:
> Nice catch. Can you please add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing-to-tutor

sure. added

regisb (approved):

PR #783, opened 2023-01-16
feat: plugin indexes

regisb:
We implement this TEP: https://discuss.openedx.org/t/tutor-enhancement-proposal-tep-plugin-indices/8182

With plugin indexes, tutor users can install and upgrade plugins directly from indexes:

    tutor plugins install ecommerce
    tutor plugins index add contrib
    tutor plugins install codejail
    tutor plugins upgrade all

This change has been long in the coming \o/

### DIFF
--- a/docs/plugins/intro.rst
+++ b/docs/plugins/intro.rst
@@ -4,7 +4,7 @@
 Introduction
 ============
 
-Tutor comes with a plugin system that allows anyone to customise the deployment of an Open edX platform very easily. The vision behind this plugin system is that users should not have to fork the Tutor repository to customise their deployments. For instance, if you have created a new application that integrates with Open edX, you should not have to describe how to manually patch the platform settings, ``urls.py`` or ``*.env.yml`` files. Instead, you can create a "tutor-myapp" plugin for Tutor. Then, users will start using your application in three simple steps::
+Tutor comes with a plugin system that allows anyone to customise the deployment of an Open edX platform very easily. The vision behind this plugin system is that users should not have to fork the Tutor repository to customise their deployments. For instance, if you have created a new application that integrates with Open edX, you should not have to describe how to manually patch the platform settings, ``urls.py`` or ``*.env.yml`` files. Instead, you can create a "tutor-myapp" plugin for Tutor. This plugin will be in charge of making changes to the platform settings. Then, users will be able to use your application in three simple steps::
 
     # 1) Install the plugin
     pip install tutor-myapp
@@ -40,4 +40,29 @@ The full plugins CLI is described in the :ref:`reference documentation <cli_plug
 Existing plugins
 ================
 
-Officially-supported plugins are listed on the `Overhang.IO <https://overhang.io/tutor/plugins>`__ website.
+Many plugins are available from plugin indexes. These indexes are lists of plugins, similar to the `pypi <https://pypi.org>`__ or `npm <npmjs.com/>`__ indexes. By default, Tutor comes with the "main" plugin index. You can check available plugins from this index by running::
+
+    tutor plugins update
+    tutor plugins search
+
+More plugins can be downloaded from the "contrib" and "wizard" indexes::
+
+    tutor plugins index add contrib
+    tutor plugins index add wizard
+    tutor plugins search
+
+The "main", "contrib" and "wizard" indexes include a curated list of plugins that are well maintained and introduce useful features to Open edX. These indexes are maintained by `Overhang.IO <https://overhang.io>`__. For more information about these indexes, refer to the official `overhangio/tpi <https://github.com/overhangio/tpi>`__ repository.
+
+Thanks to these indexes, it is very easy to download and upgrade plugins. For instance, to install the `notes plugin <https://github.com/overhangio/tutor-notes/>`__::
+
+    tutor plugins install notes
+
+Upgrade all your plugins with::
+
+    tutor plugins upgrade all
+
+To list indexes that you are downloading plugins from, run::
+
+    tutor plugins index list
+
+For more information about these indexes, check the `official Tutor plugin indexes (TPI) <https://github.com/overhangio/tpi/>`__ repository.
--- a/tutor/commands/plugins.py
+++ b/tutor/commands/plugins.py
@@ -1,21 +1,24 @@
 from __future__ import annotations
 
 import os
-import urllib.request
+import tempfile
+import typing as t
 
 import click
 import click.shell_completion
 
 from tutor import config as tutor_config
-from tutor import exceptions, fmt, hooks, plugins
+from tutor import exceptions, fmt, hooks, plugins, utils
+from tutor.plugins import indexes
 from tutor.plugins.base import PLUGINS_ROOT, PLUGINS_ROOT_ENV_VAR_NAME
+from tutor.types import Config
 
 from .context import Context
 
 
 class PluginName(click.ParamType):
     """
-    Convenient param type that supports plugin name autocompletion.
+    Convenient param type that supports autocompletion of installed plugin names.
     """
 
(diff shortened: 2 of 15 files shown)


regisb:
@overhangio/tutor-maintainers This PR is ready for a review. I think that the new plugin indexes that we introduce here are very important for plugin maintainers, so I would really like to have 3rd party opinions. For instance, do folks from eduNEXT or Opencraft see the point of plugin indexes? This is what they would look like, in the current implementation;

https://github.com/overhangio/tutor-plugins-index/blob/master/main/olive/plugins.yml
https://github.com/overhangio/tutor-plugins-index/blob/master/contrib/olive/plugins.yml

Would you consider building your own index? Would you like to add your plugins to the contrib index?

docs/plugins/intro.rst
```diff
@@ -40,4 +40,27 @@ The full plugins CLI is described in the :ref:`reference documentation <cli_plug
 Existing plugins
 ================
 
-Officially-supported plugins are listed on the `Overhang.IO <https://overhang.io/tutor/plugins>`__ website.
+Many plugins are available from plugin indexes. These indexes are list of plugins, similar to the `pypi <https://pypi.org>`__ or `npm <npmjs.com/>`__ indexes. By default, Tutor comes with the "main" plugin index. You can check available plugins from this index by running::
```
arbrandes, line 43:
```suggestion
Many plugins are available from plugin indexes. These indexes are lists of plugins, similar to the `pypi <https://pypi.org>`__ or `npm <npmjs.com/>`__ indexes. By default, Tutor comes with the "main" plugin index. You can check available plugins from this index by running::
```

docs/plugins/intro.rst
```diff
@@ -40,4 +40,27 @@ The full plugins CLI is described in the :ref:`reference documentation <cli_plug
...
+
+    tutor plugins update
+    tutor plugins search
+
+More plugins can be downloaded from the "contrib" and "wizard" indexes::
```
arbrandes, line 48: When reading this the first time (before reading the code), it wasn't clear to me that "contrib" and "wizard" are keywords that Tutor will match against indexes defined in [tutor-plugins-index](https://github.com/overhangio/tutor-plugins-index).  Maybe we could add something along the lines of:

```suggestion
More plugins can be downloaded from the "contrib" and "wizard" indexes (both of which are defined in the `official index repository <https://github.com/overhangio/tutor-plugins-index>`__, and thus automatically recognized as such by Tutor)::
```

tutor/plugins/indexes.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    Remove an index to the list if present.
+
+    Return True if if the list of indexes was modified.
+    """
+    indexes = get_all(config)
+    url = hooks.Filters.PLUGIN_INDEX_URL.apply(url)
+    if url not in indexes:
+        return False
+    indexes.remove(url)
```
arbrandes, line 119: What happens if somebody removes one of the official indices?

arbrandes (commented):
Did not actually test this yet, but made a couple of comments.  Looks great!

tutor/plugins/indexes.py
regisb: It's just like removing an apt or pypi index: nothing much happens. Installed plugins are not uninstalled or deactivated. But plugins will no longer be automatically upgraded with `tutor plugins upgrade all` and cannot be installed or uninstalled.

docs/plugins/intro.rst
regisb: Good catch, I'll change the wording.

tutor/plugins/indexes.py
arbrandes: Makes sense, thanks!

felipemontoya:
I like the idea of the plugin indexes quite a lot. I can't comment on the specifics of how it's implemented due to my lack of familiarity with the tutor-core code.

As a third party with a team I would probably build my own index so that I can share plugins easily with them. I also like the idea of the contrib index so that things that are used by the community broadly can found even more easily. I suppose that contrib index would be also installed by default but listing something there will require some form of vouching or vetting.

regisb:
Thanks for commenting @felipemontoya! Do you agree with the addition of codejail to the contrib index? https://github.com/overhangio/tpi/blob/master/contrib/olive/plugins.yml Are there any other plugins you would like to add?

Alec4r:
@regisb wow thanks, I have been waiting for this, but I've got a question for you, what happen with private plugins? will we able to add our own index?.

I mean, I would like add something like:

tutor plugins index add edunext https://GitHub.com/eduNEXT/tutor-indices.txt

Something like a git remote.

regisb:
> what happen with private plugins? will we able to add our own index?

Yes :) First of all, the user should have access to the plugins.yml file. It can be accessible behind a VPN, a basic auth http://user:password@edunext.com/index/ url, or because they downloaded the index it to their disk (`tutor plugins index add` supports local paths).

Second, "src" parameter for each all plugins that are listed in the index are actually pip-compatible specs that will be read from a requirements.txt file. This means that you can add there any reference that follows this spec: https://pip.pypa.io/en/stable/reference/requirements-file-format/ In particular, you can write:

    # Python package from a private index
    src: |
        --index-url=https://pip.mymirror.org
        my-plugin>=10.0 

All of this is explained with better words in the docs: https://github.com/overhangio/tutor/pull/783/files#diff-d711f720e3c5a7d95c0d3a3bfce00b6aa51a4377c94306328c95893693586de1

felipemontoya:
I definitely agree that codejail should be in contrib. Ideally lots of users for that one.

I think that codejail is the only plugin maintained by edunext that should land there at the moment. I hope we will be making more in the future that are broad like this, but we have to make them first ;)

regisb:
@kdmccormick I'd like to merge this -- unless you want to review it first? Let me know.

kdmccormick:
@regisb Yes, I would like to do a quick review, I will do it within the next two hours. Thanks for the reminder.

tutor/commands/plugins.py
```diff
@@ -105,48 +171,279 @@ def disable(context: Context, plugin_names: list[str]) -> None:
...
+
+        tutor plugins update
+
+    Plugin index cache is stored locally in the following subdirectory of the Tutor project environment:
+
+        plugins/index/cache.yml
+    """
+
+
+@click.command(name="list", help="List plugin indices")
```
kdmccormick, line 362:
```suggestion
@click.command(name="list", help="List plugin indexes")
```
Both words are correct, but you're using "indexes" everywhere else so I recommend sticking to that.

tutor/commands/plugins.py
```diff
@@ -105,48 +171,279 @@ def disable(context: Context, plugin_names: list[str]) -> None:
...
+Source: {plugin.src}
+Status: {plugin_status(name)}
+Author: {plugin.author}
+Maintainer: {plugin.maintainer}
+Homepage: {plugin.url}
+Index: {plugin.index}
+Description: {plugin.description}"""
+            )
+            return
+    raise exceptions.TutorError(f"Plugin could not be found: {name}")
```
kdmccormick, line 328:
```suggestion
    raise exceptions.TutorError(f"Plugin could not be found in indexes: {name}")
```

If I run this on a locally-installed plugin that's not in an index, then I will see this error. I think it'd be clearer to be told that the plugin is "not found in an index" than "not found".

kdmccormick (approved):
Fantastic!

tutor/commands/plugins.py
regisb: Good catch! I changed the terminology after I wrote the TEP.

regisb:
Thanks for the quick review Kyle! I'll merge this now.

PR #788, opened 2023-01-21
add REDIS_KEY_PREFIX

lpm0073:
adds a suffix to the redis KEY_PREFIX, which in turns enables a single remote Redis cluster to be used across multiple environments. This is used in k8s deployments that rely on a remote Redis service, in cases where the cluster hosts multiple open edx environments.


regisb:
I would really not recommend storing data from different Open edX instances in a single Redis database. The risk of key confusion is just too big. Also, we would have to define different prefixes for 3rd-party IDAs as well, such as ecommerce and discovery.

But I understand the need to run multiple Open edX instances that are connected to a single Redis cluster. Instead of defining key prefixes, can I suggest that you use the [`*_REDIS_DB`](https://docs.tutor.overhang.io/configuration.html#open-edx-customisation) settings? In particular: OPENEDX_CELERY_REDIS_DB, OPENEDX_CACHE_REDIS_DB and DISCOVERY_CACHE_REDIS_DB. (and I just realized that there is no such db setting for ecommerce)

regisb:
Now closing. Please re-open if you think this change is still needed.

PR #798, opened 2023-02-08
feat: add cli tool for listing available patches

MaferMazu:


regisb:
Looking good so far. Just let me know when you need a review.

MaferMazu:
This is ready for review @regisb @kdmccormick

tutor/env.py
```diff
@@ -215,6 +215,69 @@ def __render(self, template: jinja2.Template) -> str:
...
+        for template_name in self.iter_templates_in(*prefix):
+            self.render_template(template_name)
+
+    def print_patches_locations(self) -> None:
+        """
+        Print patches locations.
+        """
+        plugins_table: list[tuple[str, ...]] = [("PATCH", "LOCATIONS")]
+        self.render_all()
+        for patch, locations in self.patches_locations.items():
```
regisb, line 269: Can we print patches in alphabetical order?

tutor/env.py
```diff
@@ -215,6 +215,69 @@ def __render(self, template: jinja2.Template) -> str:
...
+        if not self.patches_locations.get(name):
+            self.patches_locations.update({name: [self.current_template]})
+        else:
+            if self.current_template not in self.patches_locations[name]:
+                self.patches_locations[name].append(self.current_template)
+
+        # Store the template's name, and replace it with the name of this patch.
+        # This handles the case where patches themselves include patches.
+        original_template = self.current_template
+        self.current_template = f"within patch: {name}"
```
regisb, line 248: I could confirm that this works. I created the following plugin:

```
from tutor import hooks

    (
        "openedx-common-settings",
        """
pouac
{{ patch('aaa-patch') }}
""",
    )
)
```

Then, `tutor config patches list` included the line:

```
aaa-patch                                       within patch: openedx-common-settings
```

Now I'm wondering if we should try to improve the formatting. "within patch" is explicit but it does not look too great in the command output. What do you think of:  `aaa-patch                    [openedx-common-settings]`?

It's a matter of personal taste of course, so if you disagree then I'm perfectly fine with your choice.

regisb (changes requested):
I really like your implementation, which is very lean and does not change the rest of the codebase. Also, the command output looks great and quite useful. Finally, I appreciate the fact that you took the time to write extensive unit tests.

Can you just address the comments and add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing-to-tutor

kdmccormick:
Nice work. The output looks great! I have no suggestions beyond what @regisb has already said.

tutor/env.py
MaferMazu: Yes, we can :D

tutor/env.py
MaferMazu: I prefer the explicit form, but I don't have any problems with brackets; maybe we need to add somewhere the meaning of the brackets, and the issue is solved :D
What do you think about explicit vs. good-looking brackets, @kdmccormick?

regisb: Explicit form it is!

tutor/env.py
MaferMazu: Ready

MaferMazu:
This is ready @regisb

PR #808, opened 2023-03-04
feat: configure uwsgi through an ini file

MoisesGSalas:
## Description
The current invocation of uwsgi doesn't allow to modify any configuration value besides the amount of workers.

This changes introduces a new patch `uwsgi-config` and a `uwsgi.ini` file to be mounted on the container. This allows us to add additional configuration values or override the existing ones.

I've tested it by manually creating a cookie with a size big enough to exceed the maximum buffer size of 8192. This caused a crash when accessing the base lms page. Afterwards I added a small plugin:

```yaml
name: uwsgi
version: 0.1.0
patches:

  uwsgi-config: |
    buffer-size = 16384
```

And was able to access the page again. I duplicated the cookie a few times and was able to crash it again and the LMS logs reported the increased buffer size:

```
invalid request block size: 20564 (max 16384)...skip
```

## Additional notes

The `if-env` block is particularly ugly but [seems to be](https://github.com/unbit/uwsgi/issues/1604#issuecomment-321320233) the only way use fallback values for env variables. This is to keep changes at a minimum and replicate the previous behaviour.

Interestingly enough, the `UWSGI_WORKERS` variable is not present in the kubernetes deployments, only on the local/docker-compose.yml file. Adding the variable to de K8S deployments and relying on the `OPENEDX_LMS_UWSGI_WORKERS` and `OPENEDX_CMS_UWSGI_WORKERS` values should allow us to remove the `if-env` block while keeping compatibility with people using the variables (note that you can redefine the value of the workers using the patch).

### DIFF
--- a/changelog.d/20230310_021412_moises.gonzalez_configurable_uwsgi.md
+++ b/changelog.d/20230310_021412_moises.gonzalez_configurable_uwsgi.md
@@ -1,0 +1,12 @@
+<!--
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+
+<!-- - 💥[Feature] Foobarize the blorginator. This breaks plugins by renaming the `FOO_DO` filter to `BAR_DO`. (by @regisb) -->
+<!-- - [Improvement] This is a non-breaking change. Life is good. (by @billgates) -->
+- [Improvement] Make it possible to extend or override the configuration of the uWSGI server. (by @MoisesGSalas)
--- a/docs/reference/patches.rst
+++ b/docs/reference/patches.rst
@@ -356,3 +356,14 @@ Python-formatted LMS settings in development. Values defined here override the v
 File: ``apps/openedx/settings/lms/production.py``
 
 Python-formatted LMS settings in production. Values defined here override the values from :patch:`openedx-lms-common-settings`.
+
+``uwsgi-config``
+================
+
+File: ``apps/openedx/settings/uwsgi.ini``
+
+A .INI formatted file used to extend or override the uWSGI configuration.
+
+Check the uWSGI documentation for more details about the `.INI format <https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#ini-files>`__ and the `list of available options <https://uwsgi-docs.readthedocs.io/en/latest/Options.html>`__.
+
+.. patch:: uwsgi-config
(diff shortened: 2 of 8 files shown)


MoisesGSalas:
@regisb they all seem like sensible requests to me. I will start working on them.

MoisesGSalas:
Hi @regisb, I made the changes you requested and opened #811.

tutor/templates/build/openedx/Dockerfile
```diff
@@ -160,8 +160,10 @@ COPY --chown=app:app revisions.yml /openedx/config/
 ENV LMS_CFG /openedx/config/lms.env.yml
 ENV CMS_CFG /openedx/config/cms.env.yml
 ENV REVISION_CFG /openedx/config/revisions.yml
+ENV UWSGI_WORKERS=2
 COPY --chown=app:app settings/lms/*.py ./lms/envs/tutor/
 COPY --chown=app:app settings/cms/*.py ./cms/envs/tutor/
+COPY --chown=app:app settings/uwsgi.ini .
```
regisb, line 166: These ENV and COPY statements should be in the "final" layer. Otherwise any change in the uwsgi.ini file will trigger useless rebuilds of the "production" layer.

changelog.d/20230310_021412_moises.gonzalez_configurable_uwsgi.md
```diff
@@ -0,0 +1,12 @@
...
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+
+<!-- - 💥[Feature] Foobarize the blorginator. This breaks plugins by renaming the `FOO_DO` filter to `BAR_DO`. (by @regisb) -->
+<!-- - [Improvement] This is a non-breaking change. Life is good. (by @billgates) -->
+- [Improvement] Make it possible to extend or override the configuration of the uWSGI sever. (by @MoisesGSalas)
```
regisb, line 12: nit: sever -> server

docs/reference/patches.rst
```diff
@@ -356,3 +356,12 @@ Python-formatted LMS settings in development. Values defined here override the v
 File: ``apps/openedx/settings/lms/production.py``
 
 Python-formatted LMS settings in production. Values defined here override the values from :patch:`openedx-lms-common-settings`.
+
+``uwsgi-config``
+================
+
+File: ``apps/openedx/settings/uwsgi.ini``
+
+A .INI formatted file used to extend or override the uWSGI configuration.
```
regisb, line 365: Please add links to the uwsgi docs:
https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#ini-files
https://uwsgi-docs.readthedocs.io/en/latest/Options.html

regisb (changes requested):

MoisesGSalas:
Hi @regisb, I've addressed your comments. Tell me if you catch something else.

regisb (approved):
This is both a great simplification and a welcome new feature. Thanks a lot!

PR #810, opened 2023-03-10
docs: troubleshoot high memory usage

### LINKED ISSUE #671: Extremely high memory utilization running locally on certain linux distributions
**Bug description**
On Redhat variants containers are taking an unreasonably large amount of memory running 

`tutor local start`

The issue is especially acute with the mysql container, but is also an issue with the lms and cms.

image

Memory is consumed until the OOM kills the container, which is then immediately restarted.

**How to reproduce**

Everything is vanilla.

Simply using

`tutor local start`

**Environment**
OS: 5.17.7-200.fc35.x86_64
Tutor: tutor, version 13.2.2

**Additional context**
This can/has been resolved by adding the following to a few service definitions in the local version of docker-compose.yml

```
mysql:
    blah: foo
    ...
    ulimits:
      nproc: 65535
      nofile:
        soft: 26677
        hard: 46677

```
If you are amenable, I can submit a PR to add this to the template with config variables.  Sane limits seem good generally.  I would consider adding them to mysql, mongo, elasticsearch, lms, and cms.

After picture looks like so:

image

regisb:
On some OS, high ulimit causes some container to use a lot of memory. We do not attempt to resolve this issue in Tutor because this is a mysql/uwsgi issue. Instead, we explain how to resolve it in the troubleshooting docs.

Close #671.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -84,13 +84,13 @@ The most common reason this happens is that you are running two different instan
     tutor dev stop
     tutor local stop
     tutor k8s stop
-    
+
 And then run your command(s) again, ensuring you're consistently using the correct Tutor variant (``tutor dev``, ``tutor local``, or ``tutor k8s``).
 
 If that doesn't work, then check if you have any other Docker containers running that may using port 3306::
 
     docker ps -a
-   
+
 For example, if you have ever used `Tutor Nightly <https://docs.tutor.overhang.io/tutorials/nightly.html>`_, check whether you still have ``tutor_nightly_`` containers running. Conversely, if you're trying to run Tutor Nightly now, check whether you have non-Nightly ``tutor_`` containers running. If so, switch to that other version of Tutor, run ``tutor (dev|local|k8s) stop``, and then switch back to your preferred version of Tutor.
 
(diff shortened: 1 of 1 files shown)


regisb:
Cc @keithgg @e0d who commented on the [corresponding issue](#671).

docs/troubleshooting.rst
```diff
@@ -112,6 +112,34 @@ You can identify which containers are consuming most resources by running::
 
     docker stats
 
+In idle mode, the "mysql" container should use ~200MB memory; ~200-300MB for the the "lms" and "cms" containers.
+
+On some operating systems, such as RedHat, ArchLinux or Fedora, a very high limit of the number of open files (``nofile``) per container may cause the " mysql", "lms" and "cms" containers to use a lot of memory: up to 8-16GB. To check whether you might impacted, run::
```
keithgg, line 117:
```suggestion
On some operating systems, such as RedHat, Arch Linux or Fedora, a very high limit of the number of open files (``nofile``) per container may cause the "mysql", "lms" and "cms" containers to use a lot of memory: up to 8-16GB. To check whether you might impacted, run::
```

keithgg (approved):
Thanks for this @regisb. :+1: from my side.

- [x] I read through the changes

PR #811, opened 2023-03-10
fix: add the missing `UWSGI_WORKERS` env variable to the k8s deployments

MoisesGSalas:
## Description

The amount of workers processes that uWSGI spawns can be configured through the environment variable `UWSGI_WORKERS` which in turn can be adjusted with the `OPENEDX_LMS_UWSGI_WORKERS` and `OPENEDX_CMS_UWSGI_WORKERS` variables.

This env variable was missing from the k8s deployments.


regisb (approved):
Awesome, thanks.

PR #813, opened 2023-03-13
feat: full edx-platform setup with `tutor dev launch -m ...`

kdmccormick:
## Preface

This is the part of [effort to make it easier to work with mounted repositories](https://github.com/openedx/wg-developer-experience/issues/146).

@regisb , previously we had agreed that it'd be good to [generate the .egg-info folder automatically using an init task](https://github.com/openedx/wg-developer-experience/issues/152#issuecomment-1379465966). I realized the other day, though, that we may as well do the same thing for installing node_modules and regenerating static assets too! With this change, **we can setup a dev env with a bind-mounted edx-platform in one command**.

In the future, I am still interesting in removing the need to reinstall node_modules and regenerate static assets; for now, though, I think that this is an improvement, and its user interface is compatible with the long-term vision.

I did my best to rework the Open edX development docs around the simplified setup. Hopefully they feel simpler now.... very open to feedback here, though.

Finally, I expect that `bindmount-canary` might be controversial, but I'm not sure of a better solution--open to other suggestions 😄 

## Description

Before this commit, setting up an edx-platform development environment took multiple steps:

      tutor dev launch
      tutor dev run --mount=/path/to/edx-platform lms bash
      >> pip install -e .
      >> npm clean-install
      >> openedx-assets build --env=dev

This commit moves the steps under ``run`` into an init task, which is automatically run by ``launch``. Thus, setup is now one command:

      tutor dev launch --mount=edx-platform

These extra init steps are only applicable when bind-mounting edx-platform (because bind-mounting the repository overrides some important artifacts that exist on the image, which must be re-generated). Thus, the new init tasks exists early if it detects that it is *not* operating on a bind-mounted repository.

Finally, we try to simplify the Open edX development docs so that it is clearer how bind-mounting fits into the development process.

Part of:
* https://github.com/openedx/wg-developer-experience/issues/146
Closes:
* https://github.com/openedx/wg-developer-experience/issues/152

This works around (but does not close) these related issues:
* https://github.com/openedx/wg-developer-experience/issues/150
* https://github.com/openedx/wg-developer-experience/issues/151

## Changed docs
image

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -5,15 +5,21 @@ Open edX development
 
 In addition to running Open edX in production, Tutor can be used for local development of Open edX. This means that it is possible to hack on Open edX without setting up a Virtual Machine. Essentially, this replaces the devstack provided by edX.
 
+.. _edx_platform_dev_env:
 
 First-time setup
 ----------------
 
 First, ensure you have already :ref:`installed Tutor <install>` (for development against the named releases of Open edX) or :ref:`Tutor Nightly <nightly>` (for development against Open edX's master branches).
 
-Then, launch the developer platform setup process::
+Then, run one of the following in order to launch the developer platform setup process::
 
+    # To use the edx-platform repository that is built into the image:
     tutor dev launch
+
+    # To bind-mount and run your local clone of edx-platform,
+    # where './edx-platform' should be replaced with your local edx-platform's path:
+    tutor dev launch --mount=./edx-platform
 
 This will perform several tasks for you. It will:
 
@@ -33,12 +39,30 @@ This will perform several tasks for you. It will:
 
 * run service initialization scripts, such as service user creation and Waffle configuration.
 
+Additionally, if you chose to bind-mount your local clone of edx-platform, it will:
+
+* re-run setup.py,
+
+* clean-reinstall Node modules, and
+
+* regenerate static assets.
+
 Once setup is complete, the platform will be running in the background:
 
 * LMS will be accessible at `http://local.overhang.io:8000 <http://local.overhang.io:8000>`_.
 * CMS will be accessible at `http://studio.local.overhang.io:8001 <http://studio.local.overhang.io:8001>`_.
 * Plugged-in services should be accessible at their documented URLs.
 
+Now, you can use the ``tutor dev ...`` command-line interface to manage your development environment. Some common commands are described below.
+
+.. note::
+
+  Wherever you see ``[--mount=./edx-platform]``, either:
+
+  * omit it, if you are using the edx-platform repository built into the image, or
+  * substitute it with ``--mount=<path/to/your/edx-platform>``.
+
+  You can :ref:`read more about bind-mounts below <bind_mounts>`.
 
 Stopping the platform
 ---------------------
@@ -47,38 +71,52 @@ To bring down your platform's containers, simply run::
 
   tutor dev stop
 
-
 Starting the platform back up
 -----------------------------
 
-Once you have used ``launch`` once, you can start the platform in the future with the lighter-weight ``start`` command, which brings up containers but does not perform any initialization tasks::
+Once you have used ``launch`` once, you can start the platform in the future with the lighter-weight ``start -d`` command, which brings up containers *detached* (that is: in the background), but does not perform any initialization tasks::
 
-  tutor dev start     # Run platform in the same terminal ("attached")
-  tutor dev start -d  # Or, run platform the in the background ("detached")
+  tutor dev start -d [--mount=./edx-platform]
 
-Nonetheless, ``launch`` is idempotent, so it is always safe to run it again in the future without risk to your data. In fact, you may find it useful to use this command as a one-stop-shop for pulling images, running migrations, initializing new plugins you have enabled, and/or executing any new initialization steps that may have been introduced since you set up Tutor::
+If you prefer to run containers *attached* (that is: in the foreground, your current terminal), you can omit the ``-d`` flag::
 
-  tutor dev launch --pullimages
+  tutor dev start [--mount=./edx-platform]
 
+When running containers attached, you can stop the platform with ``Ctrl+c``, or switch to detached mode using ``Ctrl+z``.
+
+Finally, you can always start your platform with ``launch``, even if you have already run it in the past. It will take longer, but it will ensure that your config is applied, your database is provisioned, your plugins are fully initialized, and (if mounted) your local edx-platform is set up. If you include the ``--pullimages`` flag it will also ensure that your container images are up-to-date as well::
+
+  tutor dev launch [--mount=./edx-platform] --pullimages
+
+Debugging with breakpoints
+--------------------------
+
+To debug a local edx-platform repository, you can add a `python breakpoint <https://docs.python.org/3/library/functions.html#breakpoint>`__ with ``breakpoint()`` anywhere in your code. Then, attach to the applicable service's container by running ``start`` (without ``-d``) followed by the service's name:
+
+  # Debugging LMS:
+  tutor dev start [--mount=./edx-platform] lms
+
+  # Or, debugging CMS:
+  tutor dev start [--mount=./edx-platform] cms
 
 Running arbitrary commands
 --------------------------
 
 To run any command inside one of the containers, run ``tutor dev run [OPTIONS] SERVICE [COMMAND] [ARGS]...``. For instance, to open a bash shell in the LMS or CMS containers::
 
-    tutor dev run lms bash
-    tutor dev run cms bash
+    tutor dev run [--mount=./edx-platform] lms bash
+    tutor dev run [--mount=./edx-platform] cms bash
 
 To open a python shell in the LMS or CMS, run::
 
-    tutor dev run lms ./manage.py lms shell
-    tutor dev run cms ./manage.py cms shell
+    tutor dev run [--mount=./edx-platform] lms ./manage.py lms shell
+    tutor dev run [--mount=./edx-platform] cms ./manage.py cms shell
 
 You can then import edx-platform and django modules and execute python code.
 
-To collect assets, you can use the ``openedx-assets`` command that ships with Tutor::
+To rebuild assets, you can use the ``openedx-assets`` command that ships with Tutor::
 
-    tutor dev run lms openedx-assets build --env=dev
+    tutor dev run [--mount=./edx-platform] lms openedx-assets build --env=dev
 
 
 .. _specialized for developer usage: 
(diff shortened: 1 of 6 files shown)


kdmccormick:
Something I just realized but don't have time to confirm right now: If `-job` services don't bindmount themes, then that'll cause an issue with static asset compilation for custom theme users, right?

docs/dev.rst
```diff
@@ -47,38 +71,52 @@ To bring down your platform's containers, simply run::
...
+
+  tutor dev start -d [--mount=./edx-platform]
+
+If you prefer to run containers *attached* (that is: in the foreground, your current terminal), you can omit the ``-d`` flag::
+
+  tutor dev start [--mount=./edx-platform]
 
-  tutor dev start     # Run platform in the same terminal ("attached")
-  tutor dev start -d  # Or, run platform the in the background ("detached")
+When running containers attached, you can stop the platform with ``Ctrl+c``, or switch to detached mode using ``Ctrl+z``.
```
regisb, line 85: Somehow I always forget about this!

docs/dev.rst
```diff
@@ -47,38 +71,52 @@ To bring down your platform's containers, simply run::
...
+
+Finally, you can always start your platform with ``launch``, even if you have already run it in the past. It will take longer, but it will ensure that your config is applied, your database is provisioned, your plugins are fully initialized, and (if mounted) your local edx-platform is set up. If you include the ``--pullimages`` flag it will also ensure that your container images are up-to-date as well::
```
regisb, line 87: I understand why you removed the `launch is idempotent` paragraph, but I'd like to keep the part where we reassure the users that it's safe to run `launch` multiple times. For instance:

> Running `launch` is always safe, even if it was already run it in the past. It will take longer...

docs/dev.rst
```diff
@@ -33,12 +39,30 @@ This will perform several tasks for you. It will:
 
 * run service initialization scripts, such as service user creation and Waffle configuration.
 
+Additionally, if you chose to bind-mount your local clone of edx-platform, it will:
+
+* re-run setup.py,
```
regisb, line 44: nit: let's remove empty lines

tutor/commands/jobs.py
```diff
@@ -40,6 +41,15 @@ def _add_core_init_tasks() -> None:
             ("mysql", env.read_core_template_file("jobs", "init", "mysql.sh"))
         )
     with hooks.Contexts.APP("lms").enter():
+        hooks.Filters.CLI_DO_INIT_TASKS.add_item(
```
regisb, line 44: now let's hope that no one will want to bind-mount edx-platform in the cms only :crossed_fingers: it's really a edge case, and I don't want to add another init task to handle it. If I'm wrong we'll come back to this comment.

tutor/templates/jobs/init/mounted-edx-platform.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Performing additional setup for bind-mounted edx-platform."
+set -x # Echo out executed lines
+
+# Regenerate Open_edX.egg-info
+pip install -e .
+
+# Regenerate node_modules
+npm clean-install
+
```
ghassanmas, line 21: There is an issue with `npm clean-install` i.e./aliased `npm ci`. 

The issue is the new version of npm >8.5.x would check depencey conflict and it would fail/errored if it exits which is the case until now in edx-platform.

I don't if it would work here, because I am not sure what the deafult version of `npm` in tutor. 

Anyway we have two option here: 
- stick to npm install as the de facto of tutor 
- pin it to spesfic version, similar to how to its done in edx-platform https://github.com/openedx/edx-platform/blob/75715e51feb49277db09027db5db398bb7046536/.github/workflows/js-tests.yml#L33

Lastly if this is expected to run **only** in dev mode, then I think in any case the default can `npm install` is okay, but for our project where devs probably might not want to touch any JS then `npm ci` might be better. 

Reference `npm ci` issue:
- a recent attemtp for me to resolve openedx/edx-platform/pull/31889 
- openedx/wg-frontend/issues/26
- openedx/wg-frontend/issues/105

docs/dev.rst
```diff
@@ -33,12 +39,30 @@ This will perform several tasks for you. It will:
...
+
+* regenerate static assets.
+
 Once setup is complete, the platform will be running in the background:
 
 * LMS will be accessible at `http://local.overhang.io:8000 <http://local.overhang.io:8000>`_.
 * CMS will be accessible at `http://studio.local.overhang.io:8001 <http://studio.local.overhang.io:8001>`_.
 * Plugged-in services should be accessible at their documented URLs.
 
+Now, you can use the ``tutor dev ...`` command-line interface to manage your development environment. Some common commands are described below.
```
regisb, line 56: I'd like to limit the use of "you" or "you can" as much as possible in the docs. I understand it's sometimes quite difficult, but I think that a more direct phrasing is more appropriate. It's not really a matter of "you" vs "we" vs "one" pronouns, but more of conciseness.

For instance:

> Now, run the ``tutor dev ...``` command line interface to manage the development environment.

> Wherever the ``[--mount=./edx-platform]`` option is present, either:
>
>   * omit it when running off the edx-platform repository built into the image, or
> * substitute it...

> Once ``launch`` was run once...

> Read more about bind-mounts :ref:`below <bind_mounts>`.

> To debug a local edx-platform repository, add a python breakpoint...

I'm not very strict about this rule. It's really just a general recommendation. If you disagree it's totally fine.

tutor/templates/build/openedx/Dockerfile
```diff
@@ -207,6 +207,13 @@ RUN openedx-assets themes \
 # Create a data directory, which might be used (or not)
 RUN mkdir /openedx/data
 
+# If this "canary" file is missing from a container, then that indicates that a
+# local edx-platform was bind-mounted into that container, thus overwriting the
+# canary. This information is useful during edx-platform initialisation.
+RUN echo \
+  "This copy of edx-platform was built into a Docker image." \
+  > bindmount-canary
```
regisb, line 215: :laughing: I wish I had a better solution, but I don't.

All I got is defining a custom template variable (via [Filters.ENV_TEMPLATE_VARIABLES](https://docs.tutor.overhang.io/reference/api/hooks/catalog.html#tutor.hooks.Filters.ENV_TEMPLATE_VARIABLES)) whenever edx-platform is bind-mounted. But it's more complex, not very intuitive, and I'm not even sure it would work, because of the order in which filters are applied...

tutor/templates/jobs/init/mounted-edx-platform.sh
```diff
@@ -0,0 +1,26 @@
...
+# Regenerate static assets.
+openedx-assets build --env=dev
```
regisb, line 23: These latter two steps will be performed every time we run `launch` with a bind-mount. I don't have a better alternative, so let's keep it that way, but we should keep it in mind.

regisb (changes requested):
This is a great step forward, I love it!

tutor/templates/jobs/init/mounted-edx-platform.sh
regisb: Ghassan, according to my tests `npm ci` works in olive.2. I recently opened a PR to replace `install` by `clean-install`: https://github.com/overhangio/tutor/pull/812/files
Why would it work in this PR, but not elsewhere?

regisb:
> If -job services don't bindmount themes, then that'll cause an issue with static asset compilation for custom theme users, right?

Yes, this is correct. So we should bind-mount the themes folder as well. 

But we're not going to be able to solve all issues related to running a custom edx-platform repo. For instance, if the edx-platform python requirements are modified, then in theory we should also run `pip install` -- but we are not going to, because it would take forever.

tutor/templates/jobs/init/mounted-edx-platform.sh
ghassanmas: Yes you are right @regisb, it works because by deafault
- tutor install node 16.14.0 https://github.com/overhangio/tutor/blob/732e0c43be5b801b7c62892bdc9b44bc74b7263f/tutor/templates/build/openedx/Dockerfile#L115 
- node 16.14.0 comes preshipped with npm 8.3.1 [^1]
- And the issue will come up if you upgrade npm version to > 8.5 i.e manually running `npm i -g npm@latest` 
- Also it will come up if for node v18 upgrade, hence version 18 comes preshipped with npm 8.x hence [^2]

Lastly usually I see this pattern: 
- install a node version 
- install latest of npm `npm -g npm@latest` 
If tutor/upstream does that it would fail hence the pinning of npm verison in my above commet. 

In summary it's an upstream issue, it doesn't affect tutor so far, it might in the future if tutor to upgrade to node 18 while its not fixed upstream yet, or it might affect us if we want to follow the above pattern _which I thinnk its the recommend way secuirty wise_. But in any case this I guess that out of scope of this PR.  


[^1]: changelog of node 16.4.0 https://github.com/nodejs/node/blob/v16.14.0/doc/changelogs/CHANGELOG_V16.md#commits 
[^2]: changelog of node 18.0.0 https://github.com/nodejs/node/blob/v18.0.0/doc/changelogs/CHANGELOG_V18.md#commits

tutor/templates/jobs/init/mounted-edx-platform.sh
kdmccormick: Something we could do, eventually, would be to use a hash to ensure that these steps are not run extraneously.
* node_modules would be easy: upon running `npm install`, create a node_modules/.package_lock_hash file, containing a hash of package-lock.json. If package-lock.json matches node_modules/.package_lock_hash, then we can skip node_modules installation.
* assets would be more complicated, but still doable. The scripts/build-assets.sh has the means to a compile a list of all asset input files. Saving the combined hash of those somewhere would allow us to skip asset compilation when assets are unchanged.

Thing is, this is just reimplementing what Docker already does. That's why I'd rather move [node_modules](https://github.com/openedx/wg-developer-experience/issues/150) and [assets](https://github.com/openedx/wg-developer-experience/issues/151) outside of the bind-mount if at all possible.

docs/dev.rst
kdmccormick: 😎

tutor/commands/jobs.py
kdmccormick: I think we should be OK here because `launch` cannot be limited to just cms. They would only really run into this edge case if they tried to skip launch and instead just run:

```
tutor dev do --mount=edx-platform init --limit=cms
```

🤞🏻

docs/dev.rst
kdmccormick: Good point. I tried that out, and the document does feel more concise now.

kdmccormick:
Alright @regisb , this is ready for another look.

regisb (approved):
This looks great to me! Do you want to add anything else before we merge?

PR #815, opened 2023-03-14
docs: clarify that extra pip requirements are a list, not a string

kdmccormick:
Credit to @bmtcril for pointing this out.


regisb (approved):
indeed!

PR #817, opened 2023-03-16
fix: v0 plugin patch ordering

regisb:
v0 plugin patches were directly added to ENV_PATCH, while v1 patches were added to ENV_PATCHES. ENV_PATCHES items are forwarded to ENV_PATCH later at runtime, which means that yaml plugin patches were always loaded before v1 patches.

This was causing weird ordering issues, as discussed here: https://discuss.openedx.org/t/using-mfe-config-to-set-logo-trademark-url-for-the-footer-component-in-olive/9578

### DIFF
--- a/changelog.d/20230316_113639_regis_fix_yaml_patch_ordering.md
+++ b/changelog.d/20230316_113639_regis_fix_yaml_patch_ordering.md
@@ -1,0 +1,1 @@
+- [Bugfix] Make sure that v0 plugin patches are applied in the same order as plugins are listed. (by @regisb)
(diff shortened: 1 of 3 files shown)


kdmccormick (approved):
Looks good.

In the future, would it make sense to get rid of ENV_PATCH? It seems like it's only a middle-man between ENV_PATCHES and iter_patches.

regisb:
Thanks for the review!

> In the future, would it make sense to get rid of ENV_PATCH? It seems like it's only a middle-man between ENV_PATCHES and iter_patches.

:thinking: the problem is that the `ENV_PATCHES` list of tuples is quite inefficient for querying by patch name. So we have to build some sort of global `dict` to keep track of all patches. If we have a global `dict`, then we might as well just make a filter out of it.

It's not very clean, but I don't have a better alternative...

PR #819, opened 2023-03-20
fix: ora2 file upload directory

### LINKED ISSUE #707: ORA2 folder is incorrectly set
According to tutor/templates/apps/openedx/settings/partials/common_all.py
**ORA2_FILEUPLOAD_BACKEND = “filesystem”
ORA2_FILEUPLOAD_ROOT = “/openedx/data/ora2"
ORA2_FILEUPLOAD_CACHE_NAME = “ora2-storage”**

Unfortunately when using Tutor (v14.0.3), the directory **data/lms/ora2/'SET-ME-PLEASE (ex. bucket-name)'** is created and used instead.

This is the default value of AWS_STORAGE_BUCKET_NAME in edx-platform.

I've tried setting up AWS_STORAGE_BUCKET_NAME to a different value through a plugin but it keeps recreating the directory **data/lms/ora2/'SET-ME-PLEASE (ex. bucket-name)'**

Are the ORA2_* variables even used? Is ORA2 properly supported in Tutor?

Is there a way to change the value of AWS_STORAGE_BUCKET_NAME in Tutor?

regisb:
ORA2 file uploads were stored in a folder named "SET-ME-PLEASE (ex.
bucket-name)" (sigh). With this change, the folder should be
automatically renamed to "openedxuploads".

This issue has been occuring since June 2019... (sigh²)

Close #707

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -120,7 +120,19 @@ GRADES_DOWNLOAD = {
 
 ORA2_FILEUPLOAD_BACKEND = "filesystem"
 ORA2_FILEUPLOAD_ROOT = "/openedx/data/ora2"
+FILE_UPLOAD_STORAGE_BUCKET_NAME = "openedxuploads"
 ORA2_FILEUPLOAD_CACHE_NAME = "ora2-storage"
+# For a long while, we were storing ora2 uploads this directory.
+ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, FILE_UPLOAD_STORAGE_BUCKET_NAME)
+deprecated_ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, "SET-ME-PLEASE (ex. bucket-name)")
+if deprecated_ora2_folder != ora2_folder and os.path.exists(deprecated_ora2_folder):
+    if os.path.exists(ora2_folder):
+        # We are really not supposed to fall in this scenario, so print a warning
+        sys.stderr.write(f"⚠️ Deprecated ORA2 folder '{deprecated_ora2_folder}' exists. Its content should be moved to '{ora2_folder}'.\n")
+    else:
+        # This folder is a relic, let's rename it
+        print(f"Moving deprecated ORA2 folder from '{deprecated_ora2_folder}' to '{ora2_folder}'...")
+        os.rename(deprecated_ora2_folder, ora2_folder)
 
 # Change syslog-based loggers which don't work inside docker containers
 LOGGING["handlers"]["local"] = {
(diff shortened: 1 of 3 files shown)


regisb:
@sambapete let me know what you think!

sambapete:
It looks good. Thank @regisb 

I am really grateful that you thought about moving the previous directory to the newest directory. I was wondering if I would have to move things around as we are currently using data/lms/ora2/'SET-ME-PLEASE (ex. bucket-name)' in order to keep things simple while you fixed the issue.

Unfortunately, I will only be able to test it when we begin testing Palm and I can install Palm on my side and migrate my current Olive test server to a Palm test server. Unless of course I try to migrate to Olive again a copy of our current Nutmeg production server.

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -120,7 +120,19 @@
 
 ORA2_FILEUPLOAD_BACKEND = "filesystem"
 ORA2_FILEUPLOAD_ROOT = "/openedx/data/ora2"
+FILE_UPLOAD_STORAGE_BUCKET_NAME = "openedxuploads"
 ORA2_FILEUPLOAD_CACHE_NAME = "ora2-storage"
+# For a long while, we were storing ora2 uploads this directory.
+ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, FILE_UPLOAD_STORAGE_BUCKET_NAME)
+deprecated_ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, "SET-ME-PLEASE (ex. bucket-name)")
```
kdmccormick, line 127:
```suggestion
ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, FILE_UPLOAD_STORAGE_BUCKET_NAME)
# For a long while, we were storing ora2 uploads this directory.
deprecated_ora2_folder = os.path.join(ORA2_FILEUPLOAD_ROOT, "SET-ME-PLEASE (ex. bucket-name)")
```
I would put the comment right above the line that it's referring to.

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -120,7 +120,19 @@
...
+if deprecated_ora2_folder != ora2_folder and os.path.exists(deprecated_ora2_folder):
```
kdmccormick, line 128:
```suggestion
if os.path.exists(deprecated_ora2_folder):
```
This:
> deprecated_ora2_folder != ora2_folder

 will always be true, as the former will always equal `$ORA2_FILEUPLOAD_ROOT/SET-ME-PLEASE (ex. bucket-name)` and the latter `$ORA2_FILEUPLOAD_ROOT/openedxuploads`.

kdmccormick (changes requested):
A couple requests, but I think this should work fine (I haven't tested it myself).

Regarding the migration approach, are you sure that doing a destructive system operation in the settings file is the way to go? I don't know of any specific reason that it would be bad; it's just something I've never seen before.

An alternative to consider: change the setting default in Tutor nightly (or even better, edx-platform master!) and add the folder rename to `upgrade_from_olive`.

regisb:
Let's close this. The change will be included in Palm, along with a nice upgrade script.

fghaas:
Apologies for the incorrect cross-reference to this issue that I made in #896; please disregard.

PR #827, opened 2023-04-12
feat: upgrade to Palm

regisb:
TODO:

- [x] Upgrade the Scorm XBlock
- [x] Upgrade MySQL to 8.0
- [x] Update the ARM docs to stop recommending switching to MySQL 8.0 (since we will already be using it)
- [x] Actually, update tutor to support arm64 natively, as discussed [here](https://github.com/openedx/wg-developer-experience/issues/81#issuecomment-1523849767).
- [x] Integrate the change to ORA2 configuration that was initially started [here](https://github.com/overhangio/tutor/pull/819)
- [x] Can we get rid of filter/action string names? They are mostly useless and confusing.
- [x] Remove support for Python 3.7. (End of life is June 2023)
- [x] Update the MFE plugin to make use of the new IMAGE_BUILD_MOUNTS filter.
- [x] Get rid of the `--skip-build` option in `local/dev start` and don't ever add a build context to the docker-compose files. (and don't build on start)


tutor/templates/build/openedx/Dockerfile
```diff
@@ -15,12 +15,26 @@ RUN apt update && \
...
+FROM minimal as dockerize
+# https://github.com/powerman/dockerize/releases
+ARG DOCKERIZE_VERSION=v0.19.0
+RUN dockerize_url="https://github.com/powerman/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-$(uname -m | sed 's@aarch@arm@')" \
+    && echo "Downloading dockerize from $dockerize_url" \
+    && curl --fail --location --output /usr/local/bin/dockerize $dockerize_url \
+    && chmod a+x /usr/local/bin/dockerize
+
+>>>>>>> c8ea6838 (feat: upgrade to Palm)
 ###### Checkout edx-platform code
```
ghassanmas, line 38: Following the slack thread in regard of the ARM issue, this bit what caused the problem for me, when I just changed it to use what exists in master i.e. this [commit](https://github.com/overhangio/tutor/commit/fa9e86e188f0a70b84f8b688e83a32a65b7bea10), it worked

ghassanmas: So I didn't open an issue since I guess the issue wouldn't exists if palm to rebase with master

regisb: :+1: I also resolved the conflict in this PR.

tutor/templates/build/openedx/Dockerfile
```diff
@@ -124,14 +142,17 @@ RUN useradd --home-dir /openedx --create-home --shell /bin/bash --uid ${APP_USER
 USER ${APP_USER_ID}
 
 # https://hub.docker.com/r/powerman/dockerize/tags
-COPY --from=docker.io/powerman/dockerize:0.19.0 /usr/local/bin/dockerize /usr/local/bin/dockerize
-COPY --chown=app:app --from=code /openedx/edx-platform /openedx/edx-platform
+COPY {% if is_buildkit_enabled() %}--link {% endif %}--from=docker.io/powerman/dockerize:0.19.0 /usr/local/bin/dockerize /usr/local/bin/dockerize
+COPY --chown=app:app --from=edx-platform / /openedx/edx-platform
```
ghassanmas, line 146: `COPY --link` caused docker to crash on my machine  when building[^1], and it was resolved by adding following the line in Dockerfile
`#syntax=docker/dockerfile:1.4` ref[^2]. Any yes BuildKit is enabled by default for me. 

[^1]: Darwin M2 docker  engine: 20.10.24
[^2]: https://stackoverflow.com/a/75199570/5532723

regisb: Thanks a lot Ghassan, this is super useful.

PR #829, opened 2023-04-13
feat: leverage `RUN --mount` for faster image building

regisb:
We make use of the Docker build cache to install python and nodejs requirements faster in the case of repeated builds.

We also implement a build-time bind-mount to make it easy to build an image given a local edx-platform repository.

TODO:
- Add a `tutor images build --mount` option to automatically mount an edx-platform repo at build-time.
- Document how to build an openedx docker image with a local repo.
- Deprecate other ways of customising the openedx Docker image (private requirements, extra xblocks, etc.)

For reference, see:

https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/ https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context


regisb:
Closing this, will merge in the palm branch.

PR #830, opened 2023-04-25
Remove hardcoded docker.io registry

angonz:
fix: Replace hardcoded references to "docker.io" by the DOCKER_REGISTRY variable.

Docker is limiting the rate of image pulls and issuing this error: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit In large deployments with high activity this issue may prevent pods to start. Removing the hardcoded references to docker.io would allow users to have copies of the docker images in our own registry and overcome the limitation imposed by docker.


regisb:
Hi @angonz! Thanks for the PR! I'm on the fence about this change. First, let me say that I understand your problem: hitting a 429 can be a pretty serious issue, especially if you need to fix a platform as soon as possible. 

On the other hand, the change that you are proposing may break other people's platforms. Consider for instance the case when a Tutor user runs a Docker registry that is _not_ a mirror of docker.io, but simply a place to store custom-built openedx images. That custom registry will not store the mysql image, for instance. But because they build and push their custom openedx Docker images to their own registry, they have overridden the `DOCKER_REGISTRY` variable. Merging your change would break their platform because they would not have the mysql image in their registry.

Alternatively, we could add a setting and replace `docker.io` by `{{ DOCKER_IO }}` everywhere. But this would come with its own set of challenges. That change would have to be ported to Tutor plugins that pull other images from docker.io. We would also need to rename base images, such as docker.io/ubuntu:20.04, for all services. This would result in many changes across all plugins.

I'm wondering if we can consider other options? For instance, logging in to Docker would double that rate limit you're facing. It can be as simple as running `docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"` in CI (this is what we do in Tutor CI).

If you are already running `docker login` on your nodes, then would you consider throwing money at the problem and buying a Docker subscription? https://www.docker.com/pricing/

Are there other ideas that I may be leaving out?

erickhgm:
@angonz could you please put the docker registry variable [here](https://github.com/angonz/tutor/blob/d3335a92105198e75522cb57bc577b5b9e6793ea/tutor/templates/build/openedx/Dockerfile#L2) as well?
I'm getting too many requests because of this line as well: `FROM docker.io/ubuntu:20.04 as minimal`

angonz:
@regisb you are completely right. 
I have another proposal, based on the fact that we actually have two kind of images: 

1. those standard, not likely to be customized (mysql, caddy, mongo, etc.), and 
2. those coming from the platform implementation (mfe, openedx, forum, plugins, etc.), likely to be changed frequently.

Most of us use the [docker.io](docker.io) registry for case 1), and a private registry for case 2). 
The proposal is to have two variables: the original DOCKER_REGISTRY for case 2) for backwards compatibility, and say "PUBLIC_DOCKER_REGISTRY" for case 1). We can set `docker.io` as default for "PUBLIC_DOCKER_REGISTRY" so we don't impact existing installations. And we would be able to override one or both of them to our private mirrors.

angonz:
> @angonz could you please put the docker registry variable [here](https://github.com/angonz/tutor/blob/d3335a92105198e75522cb57bc577b5b9e6793ea/tutor/templates/build/openedx/Dockerfile#L2) as well? I'm getting too many requests because of this line as well: `FROM docker.io/ubuntu:20.04 as minimal`

In this case, I've seen many references to docker.io:

- [Makefile](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/Makefile#L93-L94)
- [Tutor docker](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/Dockerfile#L13)
- Platform Dockerfile [as you pointed](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/tutor/templates/build/openedx/Dockerfile#L2) and also [here](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/tutor/templates/build/openedx/Dockerfile#L127)
- The [permissions Dockerfile](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/tutor/templates/build/permissions/Dockerfile#L1)

In my opinion, here is not that necessary, as you don't usually build images from production environment. The big problem is when the production servers attempt to download many images in a short timeframe. 
Actually, I wouldn't use a private registry for the base images. However if this is a relevant use case, we could have another variable for images used in the build process.

regisb:
> Most of us use the [docker.io](https://github.com/overhangio/tutor/pull/docker.io) registry for case 1), and a private registry for case 2). The proposal is to have two variables: the original DOCKER_REGISTRY for case 2) for backwards compatibility, and say "PUBLIC_DOCKER_REGISTRY" for case 1). We can set docker.io as default for "PUBLIC_DOCKER_REGISTRY" so we don't impact existing installations. And we would be able to override one or both of them to our private mirrors.

This is basically the same as my own proposal @angonz, right?

> Alternatively, we could add a setting and replace docker.io by {{ DOCKER_IO }} everywhere. But this would come with its own set of challenges. That change would have to be ported to Tutor plugins that pull other images from docker.io. We would also need to rename base images, such as docker.io/ubuntu:20.04, for all services. This would result in many changes across all plugins.

Please review the drawbacks I listed. What do you think about them?

Also, what do you think about my other two suggestions?

> I'm wondering if we can consider other options? For instance, logging in to Docker would double that rate limit you're facing. It can be as simple as running docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD" in CI (this is what we do in Tutor CI).

> If you are already running docker login on your nodes, then would you consider throwing money at the problem and buying a Docker subscription? https://www.docker.com/pricing/

erickhgm:
Ok, but the two first locations only a tutor contributor will use, and in this case, that person doesn't need to download the image every time, the docker local cache is used. The problem happens in the locations that are used in a CI pipeline.
- [Makefile](https://github.com/overhangio/tutor/blob/2e276cbb09c47e177bca6c0bcd44aee53bb5a0b6/Makefile#L93-L94)
- [Tutor docker]

angonz:
Sorry I messed up the PR. I'll try to fix it...

angonz:
I understand your points.
The drawback of the latest approach is that it requires to touch the docker daemons, which may depend on the cluster deployment. 

I agree it's a good idea to make authenticated downloads (a pro account costs only $5/mo and allows 5000 downloads a day). It would also help get more control on the account usage.

According to [this article](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) it can be achieved creating a k8s secret and using `imagePullSecrets` in the pods. 

For AWS EKS there's [an article explaining](https://aws.amazon.com/es/blogs/containers/authenticating-with-docker-hub-for-aws-container-services/), which involves creating a service account and referencing it in the pod's manifest. I don't know why it creates the service account with `imagePullSecrets` instead of directly applying it to the pod as in the previous article.

What do you think?

regisb:
Right. The two articles you point out summarize a "proper" solution. Do you think this is an acceptable solution for you?

>  I don't know why it creates the service account with imagePullSecrets instead of directly applying it to the pod as in the previous article.

I would think this is to avoid repeating the same imagePullSecrets in many different deployments. That way, if the secret changes, you only to modify a single ServiceAccount, and not all deployments.

angonz:
Yes, I agree that -if the effect is the same-, having a service account would centralize the management of the secret.
This is how I think we can do:
- Create variables DOCKERHUB_USERNAME and DOCKERHUB_PASSWORD
- Add a service account to all deployments (how can we name it?)
- If DOCKERHUB_USERNAME and DOCKERHUB_PASSWORD are set, create the secret and assign to the service account. If not, just leave the service account as default, which should not hurt the deployments.

If it looks good I can try this way.

regisb:
> If it looks good I can try this way.

Hmmm I'm not sure what you mean by that. Do you push those changes to Tutor core? I would advise against that, as I am generally opposed to adding new, very specific config settings. Can you instead create a plugin?

angonz:
Ok... sorry I didn't know it was possible to change part of a manifest using a plugin... would you give me a tip here?

regisb:
You should implement the "k8s-override" patch: https://docs.tutor.overhang.io/reference/patches.html#k8s-override

PR #831, opened 2023-04-27
Simplify "*-permissions" containers and other docker-compose shenanigans

regisb:
- [Improvement] Greatly simplify ownership of bind-mounted volumes with docker-compose. Instead of running one service per application, we run just a single "permissions" service. This change should be backward-compatible. 
- 💥[Improvement] Deprecate the `RUN_LMS` and `RUN_CMS` tutor settings, which should be mostly unused.


regisb:
Closing this, will merge in the palm branch.

Issue #823, opened 2023-04-03
`authsource` configuration for mongo credentials 


johanv26:
**Is your feature request related to a problem? Please describe.**

Hi, I was configuring my mongo credentials for my openedx instance but I found an error related to the `authSource` key. 

In the first approach, I configurated with the `MONGODB_AUTH_SOURCE` in my config.yml but I  got an authentication error in the LMS.
![image](https://user-images.githubusercontent.com/51926076/229566353-ec9c49d3-2074-4f85-9611-5320d98603fd.png)

This is because while researching I find that the key is in camelcase.
https://github.com/overhangio/tutor/blob/master/tutor/templates/apps/openedx/settings/partials/common_all.py#L16
Something I think is ok because pymongo use it in camelcase.
![2023-04-03_10-52](https://user-images.githubusercontent.com/51926076/229563381-68a9430e-a3ec-42cd-bc8b-910f01be7f1e.png)

However, the error was presented because edx-platform use another variable when credentials are provided.
https://github.com/openedx/edx-platform/blob/master/xmodule/mongo_utils.py#L72

Looking more ,  the `auth_source `is configured with another key name:  `authsource` from the kwargs.
https://github.com/openedx/edx-platform/blob/master/xmodule/mongo_utils.py#L34
So for my second approach, and provisional solution I added `authsource` kwarg  in a patch:

``` yaml
 openedx-common-settings: |

  DOC_STORE_CONFIG['authsource'] = "{{ MONGODB_AUTHSOURCE }}"
  CONTENTSTORE["DOC_STORE_CONFIG"] = DOC_STORE_CONFIG
  update_module_store_settings(MODULESTORE,doc_store_settings=DOC_STORE_CONFIG)
```
And then the connection was successful and my mongo atlas instance started working with my tutor openedx lms and cms.

**Describe the solution you'd like**
- One possibility is to provide another field in the `mongodb_parameters` dict https://github.com/overhangio/tutor/blob/master/tutor/templates/apps/openedx/settings/partials/common_all.py#L8  for `authsource` without camelcase.
 - another  possibility is adding custom extra fields(kwargs) for the `mongo_db_parameters` dict.

**Describe alternatives you've considered**
[...]

PR #841, opened 2023-05-24
fix: use correct case for authsource mongo

johanv26:
# Description
Update `authSource`  keyword argument to  `authsource`

Authsource was configurated in camelcase.
https://github.com/overhangio/tutor/blob/master/tutor/templates/apps/openedx/settings/partials/common_all.py#L16 Something I think is ok because pymongo uses it in camelcase. 

However, the error was presented because edx-platform use another variable when credentials are provided **(LOWERCASE)**. https://github.com/openedx/edx-platform/blob/master/xmodule/mongo_utils.py#L72

Looking more ,  the `auth_source `is configured with another key name:  `authsource` **(LOWERCASE)** from the kwargs. https://github.com/openedx/edx-platform/blob/master/xmodule/mongo_utils.py#L34

This won't change the **kwargs behavior of the camelCase in pymongo `MongoClient` because pymongo has a caseInsentive method. https://github.com/mongodb/mongo-python-driver/blob/3.12.3/pymongo/mongo_client.py#L651
 it's probably OK to pass `authsource` instead of authSource as a keyword argument.

## issues related
https://github.com/overhangio/tutor/issues/823

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -13,7 +13,7 @@ mongodb_parameters = {
     "password": {% if MONGODB_PASSWORD %}"{{ MONGODB_PASSWORD }}"{% else %}None{% endif %},
     # Connection/Authentication
     "ssl": {{ MONGODB_USE_SSL }},
-    "authSource": "{{ MONGODB_AUTH_SOURCE }}",
+    "authsource": "{{ MONGODB_AUTH_SOURCE }}",
     "replicaSet": {% if MONGODB_REPLICA_SET %}"{{ MONGODB_REPLICA_SET }}"{% else %}None{% endif %},
     {% if MONGODB_AUTH_MECHANISM %}"authMechanism": "{{ MONGODB_AUTH_MECHANISM }}",{% endif %}
 }
--- a/changelog.d/20230526_151849_johan.castiblanco_update_authsource_mongo.md
+++ b/changelog.d/20230526_151849_johan.castiblanco_update_authsource_mongo.md
@@ -1,0 +1,1 @@
+- [Bugfix] Change `authSource` to `authsource`(LOWERCASE) in mongo db parameters. This allow to authenticate with credentials in openedx code.(by @johanv26)


tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -13,7 +13,7 @@
     "password": {% if MONGODB_PASSWORD %}"{{ MONGODB_PASSWORD }}"{% else %}None{% endif %},
     # Connection/Authentication
     "ssl": {{ MONGODB_USE_SSL }},
-    "authSource": "{{ MONGODB_AUTH_SOURCE }}",
+    "authsource": "{{ MONGODB_AUTH_SOURCE }}",
     "replicaSet": {% if MONGODB_REPLICA_SET %}"{{ MONGODB_REPLICA_SET }}"{% else %}None{% endif %},
```
johanv26, line 17:
```suggestion
    "replicaset": {% if MONGODB_REPLICA_SET %}"{{ MONGODB_REPLICA_SET }}"{% else %}None{% endif %},
```
Looking more if pymongo use insensitive dict https://github.com/mongodb/mongo-python-driver/blob/3.12.3/pymongo/mongo_client.py#L651  we can test changing  also `replicaSet` to `replicaset` in order to have mongo_db_parameters in standard way kwargs using lowercase.

johanv26:
@regisb @keithgg I have created the PR to continue. Regards,

keithgg (approved):
Happy to merge this @johanv26. I checked why we never ran into this issue and it's because we add the username and password to `MONGODB_HOST` and don't specify them separately.

- [x] I read through the code
- [x] I tested this locally to make sure the migration `split_modulestore_django.0002_data_migration` passes for mongodb+srv urls if the username and password is encoded in the `MONGODB_HOST`.

tutor/templates/apps/openedx/settings/partials/common_all.py
regisb: It ain't broken (yet) so let's not try to fix this.

regisb (commented):
Looking good! Could you please add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing

johanv26:
> Looking good! Could you please add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing

Changelog added, regards.

regisb (approved):
thanks again for the PR!

PR #844, opened 2023-05-27
feat: Add support for loading in-cluster config when running inside a pod

### LINKED ISSUE #843: tutor k8s quickstart/launch fails when using serviceAccount on Kubernetes
Tutor is using the Kubernetes package to load kube config.
https://github.com/overhangio/tutor/blob/87ae658773e454a65bc2afd8a7b318a2d9d8ef59/tutor/commands/k8s.py#L29
this is not an issue but itself, but imagine you're using serviceAccount and role binding in EKS. there's no need to create .kube/config file and the pod has access to `kubectl` commands. 
running `tutor k8s quickstart` will fail, since there's no config inside the `.kube/config`.

``` python
  File "/usr/local/bin/tutor", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/cli.py", line 24, in main
    cli()  # pylint: disable=no-value-for-parameter
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 201, in quickstart
    context.invoke(init, limit=None)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 339, in init
    jobs.initialise(runner, limit_to=limit)
  File "/usr/local/lib/python3.9/site-packages/tutor/jobs.py", line 76, in initialise
    runner.run_job_from_template(service, *path)
  File "/usr/local/lib/python3.9/site-packages/tutor/jobs.py", line 22, in run_job_from_template
    self.run_job(service, command)
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 84, in run_job
    active_jobs = self.active_job_names()
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 68, in active_job_names
    api = K8sClients.instance().batch_api
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 33, in instance
    cls._instance = cls()
  File "/usr/local/lib/python3.9/site-packages/tutor/commands/k8s.py", line 25, in __init__
    config.load_kube_config()
  File "/usr/local/lib/python3.9/site-packages/kubernetes/config/kube_config.py", line 813, in load_kube_config
    loader = _get_kube_config_loader(
  File "/usr/local/lib/python3.9/site-packages/kubernetes/config/kube_config.py", line 772, in _get_kube_config_loader
    raise ConfigException(
kubernetes.config.config_exception.ConfigException: Invalid kube-config file. No configuration found.
```

Os: Ubuntu
Tutor version: 14.2.3

CodeWithEmad:
in some cases, tutor might run inside a pod, which that pod has access to a cluster via role binding and a service account. this way, there's no ./kube/config file, but kubectl commands run with no issue.

Close #843

### DIFF
--- a/changelog.d/20230527_174319_codewithemad_kubeconfig.md
+++ b/changelog.d/20230527_174319_codewithemad_kubeconfig.md
@@ -1,0 +1,1 @@
+ - [Feature] Add support for loading in-cluster config when running inside a pod. In certain scenarios, Tutor may operate within a pod that has access to a cluster through role binding and a service account. In these instances, the ./kube/config file may not be present, but kubectl commands can still execute without any problems. (by @CodeWithEmad)
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -1,3 +1,4 @@
+import os
 from datetime import datetime
 from time import sleep
 from typing import Any, Iterable, List, Optional, Type
@@ -26,7 +27,20 @@ class K8sClients:
         # pylint: disable=import-outside-toplevel
         from kubernetes import client, config
 
-        config.load_kube_config()
+        if os.path.exists(config.kube_config.KUBE_CONFIG_DEFAULT_LOCATION):
+            # found the kubeconfig file, let's load it!
+            config.load_kube_config()
+        elif (
+            config.incluster_config.SERVICE_HOST_ENV_NAME in os.environ
+            and config.incluster_config.SERVICE_PORT_ENV_NAME in os.environ
+        ):
+            # We are running inside a cluster, let's load the in-cluster configuration.
+            config.load_incluster_config()
+        else:
+            raise exceptions.TutorError(
+                f"there is no Kubernetes configuration file located in {config.kube_config.KUBE_CONFIG_DEFAULT_LOCATION}, and the service environment variables {config.incluster_config.SERVICE_HOST_ENV_NAME} and {config.incluster_config.SERVICE_PORT_ENV_NAME} have not been defined. To connect to a cluster, please configure your host appropriately."
+            )
+
         self._batch_api = None
         self._core_api = None
         self._client = client


CodeWithEmad:
does it need to add documentation on the service account and role binding inside https://docs.tutor.overhang.io/k8s.html ?

CodeWithEmad:
Hi @regisb 
I tested this approach on my fork, installed inside a pod within EKS and it works, but I'm a bit skeptical about the approach. maybe using a switch would be better?

CodeWithEmad:
Hi @regisb 
I tested this approach on my fork, installed inside a pod within EKS and it works, but I'm a bit skeptical about the approach. maybe using a switch would be better?

CodeWithEmad:
Hi @regisb 
I tested this approach on my fork, installed inside a pod within EKS and it works, but I'm a bit skeptical about the approach. maybe using a switch would be better?

tutor/commands/k8s.py
```diff
@@ -26,7 +26,13 @@ def __init__(self) -> None:
         # pylint: disable=import-outside-toplevel
         from kubernetes import client, config
 
-        config.load_kube_config()
+        try:
+            config.load_incluster_config()
```
regisb, line 30: The problem I have with this implementation is that the default path will fail for most people -- who are not running Tutor inside a pod. 

Is there any other clue that could allow Tutor to be smart and guess that it's running inside a pod? For instance: an environment variable, a certain file, etc.?

regisb: According to the kubernetes package source code, we should check for the following environment variables: `kubernetes.config.incluster_config.SERVICE_HOST_ENV_NAME`, `kubernetes.config.incluster_config.SERVICE_PORT_ENV_NAME`.
We should also check for the existence of the presence of the following file: `kubernetes.config.kube_config.KUBE_CONFIG_DEFAULT_LOCATION`.

regisb: Exactly.

CodeWithEmad: I made a slight modification.
Initially, we will verify if the KUBE_CONFIG_DEFAULT_LOCATION exists. Next, we will check if the tutor is operating within a pod and load the incluster_config. If not, an error will be raised.

tutor/commands/k8s.py
```diff
@@ -49,6 +50,43 @@ def core_api(self):  # type: ignore
...
+        )
+
+    def kube_config_exists(self) -> bool:
+        # pylint: disable=import-outside-toplevel
+        from kubernetes import config
+
+        kube_config_path = config.kube_config.KUBE_CONFIG_DEFAULT_LOCATION
+        return os.path.exists(kube_config_path)
+
+    def load_config(self) -> None:
```
regisb, line 71: Your implementation is top-notch. I just have two minor comments:

1. Would you like to simplify the code by removing the `kube_config_exists` and `is_running_inside_pod` functions, and inlining the corresponding call? The functions are not used anywhere else, and they are quite simple, so I figure we can just remove them. I don't have a strong opinion, so follow your judgment here.
2. Exception management: I think that if either `load_kube_config` or `load_incluster_config` calls fail, then we should just let Tutor crash, so there is no need to catch the ConfigException. On the other hand, if we detect early that there is no kube config file or cluster env vars, then we should raise a TutorException that will be caught by the Tutor main process. In other words:

```
if self.kube_config_exists():
    # Loading kubeconfig file.
    config.load_kube_config()
elif self.is_running_inside_pod():
    # Loading in-cluster configuration.
    config.load_incluster_config()
else:
    raise TutorError(f"No Kubernetes configuration file in {config.kube_config.KUBE_CONFIG_DEFAULT_LOCATION} and service environment variables {SERVICE_HOST_ENV_NAME}, {SERVICE_PORT_ENV_NAME} are not defined. Configure your host to connect to blablabla (add intelligent suggestions here)")
```

changelog.d/20230527_174319_codewithemad_kubeconfig.md
```diff
@@ -0,0 +1 @@
+[Bugfix] Prioritize loading of incluster_config over kube_config. In certain scenarios, Tutor may operate within a pod that has access to a cluster through role binding and a service account. In these instances, the ./kube/config file may not be present, but kubectl commands can still execute without any problems. (by @CodeWithEmad)
```
regisb, line 1: nit: add a hyphen "-" at the beginning of the line because the changelog entries are collected into an item list.

regisb (changes requested):
Your changes look great! I have just a couple minor comments.

changelog.d/20230527_174319_codewithemad_kubeconfig.md
CodeWithEmad: > The functions are not used anywhere else, and they are quite simple, so I figure we can just remove them.

I agree with you. I made those changes to make things clearer. It's no problem to remove the functions.

> if we detect early that there is no kube config file or cluster env vars, then we should raise a TutorException that will be caught by the Tutor main process

Totally Agree.

CodeWithEmad: Updated,
is it necessary to update documents on k8s?

regisb (approved):
Looks good to me!

changelog.d/20230527_174319_codewithemad_kubeconfig.md
regisb: No, we are good to go :) Thanks!

CodeWithEmad:
@regisb 
oops. I accidentally changed my global config for name and email :(
can I change the author's name and email?

PR #858, opened 2023-06-15
fix: intermittent connection errors to mongodb because of eager uwsgi

viadanna:
This prevents `pymongo.errors.ServerSelectionTimeoutError: No servers found yet`

https://discuss.openedx.org/t/mongodb-intermittent-connection-timeout-errors-in-pymongo-topology-py/7278/4

### DIFF
--- a/tutor/templates/build/openedx/settings/uwsgi.ini
+++ b/tutor/templates/build/openedx/settings/uwsgi.ini
@@ -8,3 +8,4 @@ processes = $(UWSGI_WORKERS)
 thunder-lock = true
 single-interpreter = true
 enable-threads = true
+lazy-apps = true


regisb:
Hi @viadanna! Thanks for the PR. Can you please explain what's the impact of that setting, and how setting it resolves the pymongo connection issue? In the [uwsgi docs](https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html?highlight=lazy#preforking-vs-lazy-apps-vs-lazy) I see:

> lazy-apps mode instead loads your application one time per worker. It will require about O(n) time to load it (where n is the number of workers), will very probably consume more memory, but will run in a more consistent and clean environment.

I must admit that this warning scares me a little. Did you observe any increase in memory usage? My worry is that this change is not really a fix for the mongodb issue, but more like a way to just ignore it. I'd rather have a sustainable solution.

Also, do you still observe this issue with pymongo 3.13.0? (the version now installed in Palm)

viadanna:
Hi @regisb 

We decided to enable Lazy loading the apps in our deployments to fix the `ServerSelectionTimeoutError` after reading the following pymongo documentation: https://pymongo.readthedocs.io/en/3.12.1/faq.html#is-pymongo-fork-safe 

It states: _Instances of MongoClient copied from the parent process have a high probability of deadlock in the child process due to the inherent incompatibilities between fork(), threads, and locks described [below](https://pymongo.readthedocs.io/en/3.12.1/faq.html#pymongo-fork-safe-details). PyMongo will attempt to issue a warning if there is a chance of this deadlock occurring._

This was consistent with the Warnings we saw in our instance logs, along with `uWSGI` documentation that says _By default uWSGI loads the whole application in the first process and after the app is loaded it does fork() itself multiple times._

After applying this patch we didn't see pymongo warnings or `ServerSeletionTimeoutErrors` in our logs again. We have this change deployed for more than a week and found no issues related to memory usage with the same underlying resources.

We haven't updated the client instance to Palm yet but we'll be spinning sandboxes in the coming weeks. But the PyMongo warning still exists in v4.4.0, the latest stable, so I expect the change to be required.

Let me know if you have further questions.

regisb:
I understand, thanks for the thorough explanation. While it does look like this is a serious issue, I do observe a rather large increase in memory requirements. In idle mode, the LMS has a memory usage of 300MB. With `lazy-apps=true` this usage jumps to 456MB. (+52%)

So I acknowledge that we have an issue, but I would like to explore other options that would not have such a large impact on the app footprint.

There is an alternative solution I found here: https://stackoverflow.com/questions/41905472/pymongo-create-mongoclient-with-connect-false-or-create-client-after-forking

If I understand correctly, adding `connect=False` to the MongoClient parameters will not cause the client to connect immediately. Instead, the client will only connect on the first request. If we are lucky, no such request is performed before the uwsgi fork.

Have you tried this solution before? If not, would you mind giving it a try?

viadanna:
I haven't tried that @regisb might be a good solution. Unfortunately, we have quite a few priority works for the client, so I wouldn't be able to test this in the short term.

How do you recommend handling this PR? Should I close it until we can get the alternative tested?

regisb:
Let's keep this open until you have time to try out the fix I suggested above.

viadanna:
Hi @regisb I've been experimenting with the suggested change but I'm seeing a few issues in Studio. I'll investigate those and get back here.

regisb:
@viadanna have you resolved the issues you mentioned?

I suspect that this issue will become moot in Quince (Dec. 2023): uwsgi is now [in maintenance mode](https://uwsgi-docs.readthedocs.io/en/latest/) and I don't feel confortable relying on a project that is no longer maintained, so we will probably want to replace uwsgi by something else. (TBD)

viadanna:
@regisb I see. I've created https://github.com/overhangio/tutor/pull/902 for the MONGODB_CONNECT setting anyway as it might be necessary with whatever replaces uwsgi.

PR #862, opened 2023-06-20
fix: add a temporary patch for the fix of cache programs command error

Faraz32123:
This PR is created as temporary fix for cache programs command that is used in tutor-discovery plugin to cache programs.
The issue link is [here](https://github.com/overhangio/tutor-discovery/issues/40).


Faraz32123:
@regisb , you can review it.

regisb (changes requested):
Looking good! Can you please add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing

Faraz32123:
@regisb this PR may not be needed now, as the [change](https://github.com/openedx/edx-platform/pull/32506) is merged in open-release/palm.master branch of edx-platform.

changelog.d/20230620_163516_faraz.maqsood_faraz_add_temporary_patch_for_cache_program_command_issue_for_discovery.md
```diff
@@ -0,0 +1,2 @@
+- 💥[Improvement] Add a patch for this commit (https://github.com/openedx/edx-platform/pull/32506/commits/b3e9f0cbd5e44f25a42a7cbf4929b71ef122b697)
```
regisb, line 1: In the changelog, try to add information that is actually useful to the end user. In this case, I suggest:

```
- [Bugfix] Make it possible to cache programs for a single site, instead of crashing when running `manage.py lms cache_programs` for all sites. (by @Faraz32123)
```

Note that this is not a breaking change, so we should not add "💥".

regisb:
We still need this PR, because tutor does not run Open edX from the open-release/palm.master branch, but from the open-release/palm.1 *tag*.

regisb (approved):
thanks for your contribution!

PR #868, opened 2023-07-10
docs: tutor uses `docker compose` now, not `docker-compose`

kdmccormick:
Per v16 changelog: https://github.com/overhangio/tutor/blob/master/CHANGELOG.md#v1600-2023-06-14

@regisb , a couple related questions:

* Is v20.10.15+ still the right Docker version to require?
* Do we count on BuildKit now? If so, it's not [default in Docker Engine until v23](https://docs.docker.com/build/buildkit/).
* I didn't update the [podman tutorial](https://docs.tutor.overhang.io/tutorials/podman.html). Do you if the podman shim still works now that we're using builtin `docker compose`?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -9,7 +9,7 @@ Requirements
 ------------
 
 * Supported OS: Tutor runs on any 64-bit, UNIX-based OS. It was also reported to work on Windows (with `WSL 2 <https://docs.microsoft.com/en-us/windows/wsl/install>`__).
-* Architecture: support for ARM64 is a work-in-progress. See `this issue <https://github.com/overhangio/tutor/issues/510>`__.
+* Architecture: Both AMD64 and ARM64 are supported.
 * Required software:
 
     - `Docker <https://docs.docker.com/engine/installation/>`__: v20.10.15+
@@ -17,7 +17,6 @@ Requirements
 
 .. warning::
     Do not attempt to simply run ``apt-get install docker docker-compose`` on older Ubuntu platforms, such as 16.04 (Xenial), as you will get older versions of these utilities.
-
 * Ports 80 and 443 should be open. If other web services run on these ports, check the tutorial on :ref:`how to setup a web proxy <web_proxy>`.
 * Hardware:
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -51,7 +51,7 @@ Tutor simplifies the deployment of Open edX by:
   :width: 500px
   :align: center
 
-Because Docker containers are becoming an industry-wide standard, that means that with Tutor it becomes possible to run Open edX anywhere: for now, Tutor supports deploying on a local server, with `docker-compose <https://docs.docker.com/compose/overview/>`_, and in a large cluster, with `Kubernetes <http://kubernetes.io/>`_. But in the future, Tutor may support other deployment platforms.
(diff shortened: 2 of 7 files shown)


kdmccormick:
Updating to a newer Docker version helped resolve an issue several folks were having with Tutor Palm: https://discuss.openedx.org/t/issue-in-tutor-palm-release-with-tuotr-dev-launch-while-installing/10629/12

Abdess updated to the latest Engine version (v24.0.3), which is also what I have. Shall I update this PR to make that the minimum Docker version, or should we go with v23.0.0, which is the first version that made Buildx the default?

regisb:
This is very strange. A couple of notes:

- Compose does not ship by default with Docker (at least on some OS). This is why we must still specify a minimum version for Compose.
- Tutor will default to `docker compose` if it's available, but it should still be compatible with `docker-compose`. If not, then it's a bug.
- The minimum required Docker version was updated in Palm, and I don't think we need to bump it to v24.
- It's difficult to troubleshoot the issue that was reported in the forum without more information, including a full stacktrace of `docker build`. For instance, it might be that there is a bug in how Tutor generates the Dockerfile based on the presence or absence of `buildx`.

docs/install.rst
```diff
@@ -10,10 +10,7 @@ Requirements
 
 * Supported OS: Tutor runs on any 64-bit, UNIX-based OS. It was also reported to work on Windows (with `WSL 2 <https://docs.microsoft.com/en-us/windows/wsl/install>`__).
 * Architecture: support for ARM64 is a work-in-progress. See `this issue <https://github.com/overhangio/tutor/issues/510>`__.
-* Required software:
-
-    - `Docker <https://docs.docker.com/engine/installation/>`__: v20.10.15+
-    - `Docker Compose <https://docs.docker.com/compose/install/>`__: v2.0.0+
+* Required software: `Docker <https://docs.docker.com/engine/installation/>`__ v20.10.15+.
```
regisb, line 13: We still need to specify the minimum Compose version, because Compose does not always ship with Docker v20.

regisb (changes requested):
Looking good outside of one minor point.

kdmccormick:
@regisb sorry, there is one other approach we would take, as a compromise:

Restore legacy builder support for Palm, but assume buildkit for Nightly, and communicate this as a breaking change for Quince.

(Selfishly, I want to be able to assume BuildKit in Nightly, because it'll make the static asset caching work I'm wrapping up much simpler :)

regisb:
Kyle, I still think we need to merge this PR, because the default is now BuildKit's `docker compose`, and not `docker-compose`.

That being said, you are right that in the current state Docker without BuildKit is broken in Tutor. We do want to support the legacy builder in Palm because its deprecation was only 5 months ago, which is a bit too short for my taste. Some servers might not have been upgraded since then.

Here's what I suggest:

1. Merge this PR, after you re-introduce the minimal compose requirement. Optionally, you can add a "Docker recommended version". You may (also optionally) remove the line about arm64 work-in-progress support from the installation docs.
2. I'll open a PR to fix the build in non-buildkit mode.
3. You can assume that BuildKit will be required in Quince, and thus ignore non-buildkit mode in nightly.

docs/install.rst
kdmccormick: Good catch, fixed.

kdmccormick:
@regisb 
1. Done 👍🏻 Merge this at your leisure.
2. Thanks!
3. I have [branch that removes the buildkit conditionals](https://github.com/overhangio/tutor/compare/nightly...kdmccormick:tutor:kdmccormick/buildkit?expand=1). Next week I'll test it out and then open a PR against nightly.

regisb:
I merged this PR manually to fix the documentation build issue. Thanks for the PR!

PR #872, opened 2023-07-17
build: copy in copy-node-modules.sh for npm post-install hook

kdmccormick:
scripts/copy-node-modules.sh replaces `openedx-assets npm`.
We run it automatically as a post-install NPM hook.
That means that the script must be available when running `npm clean-install`.

This should merge right after https://github.com/openedx/edx-platform/pull/32767 merges.

Part of: https://github.com/openedx/edx-platform/issues/31604


regisb (approved):
LGTM! Don't hesitate to ping me for urgent matters, as you did on Slack. I expect to be off for the next two weeks or so.

PR #885, opened 2023-08-02
Add INFO_EMAIL variable

misilot:
Add a default value for the INFO_EMAIL which is used by the Authn frontend app.

https://github.com/openedx/frontend-app-authn/blob/a723058bc114a8c70b463e2aa15fd11dea1e1705/README.rst?plain=1#L91-L93


regisb:
What's the impact of this change? Can you please screenshots before/after?

misilot:
I'm trying that as well. For changes like that does the `mfe` image need to be rebuilt for it to to be recoginzied? As, I am having trouble building mfe on Olive.

misilot:
They need to be applied to the MFE it looks like.

regisb:
No, you don't need to rebuild the MFE image after a config change since Palm. It was the case in Olive, but it's no longer necessary now that we have runtime config.

PR #886, opened 2023-08-07
Fix: Docs

CodeWithEmad:
in the Hooks catalog, the word "Filters" was left out in the documentation.
here:
https://github.com/overhangio/tutor/blob/8de5edfab735d077fab7b2694d668354caeeead7/tutor/hooks/catalog.py#L150
and here:
https://github.com/overhangio/tutor/blob/8de5edfab735d077fab7b2694d668354caeeead7/tutor/hooks/catalog.py#L154

I also had a change log entry, related to docs, so I've added this to it.


changelog.d/20230801_201804_codewithemad_remove_wizard.md
```diff
@@ -1 +1 @@
- - [Improvement] Wizard references removed. (by @CodeWithEmad)
```
regisb, line 1: Please don't modify or add new changelog entries for documentation-only changes.

regisb (changes requested):
Looking good! Just get rid of the changelog changes.

changelog.d/20230801_201804_codewithemad_remove_wizard.md
CodeWithEmad: sure thing.

regisb (approved):
thanks for the PR :) it's a tiny change but it's useful!

CodeWithEmad:
Happy to be a part of this 🤘

PR #896, opened 2023-08-24
fix: Apply mysqld character set fix to Kubernetes deployment

fghaas:
This is a follow-up fix to #890, where the corresponding change was added to the mysqld invocation in the "tutor local" (that is, docker-compose) deployment method, but omitted from its "tutor k8s" equivalent.


fghaas:
This is a somewhat speculative change that I am submitting based on the outcome of the discussion in #819 — this is why this PR is currently marked as a Draft.

regisb (approved):
This is 100% right, so I moved the PR away from draft mode. There's even a changelog entry! Let's merge this right away.

regisb:
Thanks a lot @fghaas!

PR #898, opened 2023-09-06
fix: parse strings prefixed with "#" in `config save --set ...`

### LINKED ISSUE #866: Pound signs in string settings are sometimes interpreted as a comment prefix
**Bug description**

When a string setting value contains a pound sign (`#`), the yaml parser in Tutor will sometimes interpret it as opening a comment rather than just a literal `#` within a string.

**How to reproduce**
```bash
tutor plugins install indigo
tutor plugins enable indigo
tutor config save --set "INDIGO_PRIMARY_COLOR=#225522"
tutor config printvalue INDIGO_PRIMARY_COLOR
   # expected: #225522
   # actual: ''
```

**Environment**
I don't think it's env dependent, but:

tutor-nightly 16.0.2, macOS Ventura 13.4.1, MacBook M1 Pro, zsh 5.9

**Additional context**
Based on discussion from: https://github.com/overhangio/tutor-indigo/pull/45#pullrequestreview-1500035659

regisb:
Pound keys were interpreted as comments. This is annoying when we want to parse html color codes, such as in:

    $ tutor config save --set "INDIGO_PRIMARY_COLOR=#225522"
    $ tutor config printvalue INDIGO_PRIMARY_COLOR
    null

Close #866.

### DIFF
--- a/changelog.d/20230906_105044_regis_parse_pound_keys_in_settings.md
+++ b/changelog.d/20230906_105044_regis_parse_pound_keys_in_settings.md
@@ -1,0 +1,1 @@
+- [Bugfix] Correctly parse strings prefixed with pound "#" key in `tutor config save --set KEY=#value` commands. (by @regisb)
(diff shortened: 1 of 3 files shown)


regisb:
Taking the liberty to merge without review, as this is such a small change.

PR #901, opened 2023-09-25
CONFIG_LOADED Action

### LINKED ISSUE #557: Preview mode breaks in Maple if PREVIEW_LMS_HOST is not a subdomain of LMS_HOST
**Bug description**

As of Maple, if `PREVIEW_LMS_HOST` is set to something other than a subdomain of `LMS_HOST`, preview mode breaks since it unexpectedly prompts the user for authentication.

**How to reproduce**
* Deploy a Tutor environment on Maple, but instead of following the standard naming naming convention of `preview.lms.example.com`, set `PREVIEW_LMS_HOST` to something like `lms-preview.example.com`.
* Open a course in Studio, such as the demo course.
* Open a unit in that course and click Preview.
* Observe that you see an authentication prompt.

**Environment**
Tutor 13.0.2.

**Additional context**
In Maple, as explained in the [release notes](https://edx.readthedocs.io/projects/open-edx-release-notes/en/latest/maple.html#altered-features), there's no Preview for the Learning MFE yet, and previews render in the legacy LMS. 

However, we can also [disable the Learning MFE](https://discuss.openedx.org/t/disabling-the-learning-mfe-in-maple-slightly-misleading-documentation/6320), at which point we get the legacy learning experience throughout. This, I suppose, might be something that quite a few sites will want to opt to use, until they can offer a smooth course authoring experience where preview mode and the published course do look the same again.

Now, if I open a course unit in Studio and then hit the Preview button, I am prompted to authenticate. I don't think that that should be happening. (I should add that authentication also does not work in that scenario, so preview mode effectively isn't functional at all.)

I am guessing (please correct me if I'm wrong here) that [unlike Studio which now uses OAuth2 authentication](https://github.com/edx/edx-platform/blob/master/docs/guides/studio_oauth.rst), preview mode is still relying on a shared session cookie. And in 7c157eccd5f0dba69b023e3dd64ed12ce6cbeaa4, Tutor followed the recommendation from the docs and changed the value of `SESSION_COOKIE_DOMAIN` from the domain shared between `LMS_HOST` and `CMS_HOST`, to _just_ `LMS_HOST`.

That happens to work fine for preview mode as well, as long as `PREVIEW_LMS_HOST` is indeed a subdomain of `LMS_HOST`. When it isn't, however, things break.

I am guessing that instead of bringing back

```json
"SESSION_COOKIE_DOMAIN": ".{{ LMS_HOST|common_domain(CMS_HOST) }}"
```

(which `templates/apps/openedx/config/lms.env.json` and `cms.env.json` were using prior to 7c157eccd5f0dba69b023e3dd64ed12ce6cbeaa4), we could instead use

```json
"SESSION_COOKIE_DOMAIN": ".{{ LMS_HOST|common_domain(PREVIEW_LMS_HOST) }}"
```

... but this will of course mean that the session cookie will be exposed to "a potentially large number of subdomains", which is exactly what the Studio OAuth2 change is aiming to avoid. So I'm not sure if that's a good way forward.

@regisb, what are your thoughts on this? How wrong am I? :)

CodeWithEmad:
This will introduce a new Action to tutor: CONFIG_LOADED
close #557

### DIFF
--- a/changelog.d/20230926_172753_codewithemad_config_loaded_action.md
+++ b/changelog.d/20230926_172753_codewithemad_config_loaded_action.md
@@ -1,0 +1,1 @@
+- 💥[Feature] New action introduced: CONFIG_LOADED. This actions is called whenever the config is loaded. (by @CodeWithEmad)
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -147,16 +147,18 @@ def save(
     env_only: bool,
 ) -> None:
     config = tutor_config.load_minimal(context.root)
-    config_full = tutor_config.load_full(context.root)
     if interactive:
         interactive_config.ask_questions(config)
     if set_vars:
         for key, value in set_vars:
             config[key] = env.render_unknown(config, value)
     if append_vars:
+        config_defaults = tutor_config.load_defaults()
         for key, value in append_vars:
             if key not in config:
-                config[key] = config_full.get(key, [])
+                config[key] = config[key] = config.get(
+                    key, config_defaults.get(key, [])
+                )
             values = config[key]
             if not isinstance(values, list):
                 raise exceptions.TutorError(
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -114,7 +114,7 @@ class ConfigListKeyValParamType(ConfigKeyValParamType):
     type=ConfigListKeyValParamType(),
     multiple=True,
     metavar="KEY=VAL",
-    help="Append an item to a configuration value of type list. The value will only be added it it is not already present. (can be used multiple times)",
+    help="Append an item to a configuration value of type list. The value will only be added if it is not already present. (can be used multiple times)",
 )
 @click.option(
     "-A",
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from copy import deepcopy
 
 import os
 
@@ -26,6 +27,15 @@ def load(root: str) -> Config:
     return load_full(root)
 
 
+def load_defaults() -> Config:
+    """
+    Load default configuration.
+    """
+    config: Config = {}
+    update_with_defaults(config)
+    return config
+
+
 def load_minimal(root: str) -> Config:
     """
     Load a minimal configuration composed of the user and the base config.
@@ -51,6 +61,7 @@ def load_full(root: str) -> Config:
     update_with_base(config)
     update_with_defaults(config)
     render_full(config)
+    hooks.Actions.CONFIG_LOADED.do(deepcopy(config))
     return config
 
 
@@ -319,3 +330,19 @@ def _update_enabled_plugins_on_unload(_plugin: str, _root: str, config: Config) 
     Note that this action must be performed after the plugin has been unloaded, hence the low priority.
     """
     save_enabled_plugins(config)
+
+
+@hooks.Actions.CONFIG_LOADED.add()
+def _check_preview_lms_host(config: Config) -> None:
+    """
+    This will check if the PREVIEW_LMS_HOST is a subdomain of LMS_HOST.
+    if not, prints a warning to notify the user.
+    """
+
+    lms_host = get_typed(config, "LMS_HOST", str, "")
+    preview_lms_host = get_typed(config, "PREVIEW_LMS_HOST", str, "")
+    if not preview_lms_host.endswith("." + lms_host):
+        fmt.echo_alert(
+            f'Warning: PREVIEW_LMS_HOST="{preview_lms_host}" is not a subdomain of LMS_HOST="{lms_host}". '
+            "This configuration is not typically recommended and may lead to unexpected behavior."
+        )
--- a/tutor/hooks/catalog.py
+++ b/tutor/hooks/catalog.py
@@ -57,6 +57,13 @@ class Actions:
     #: :parameter str name: docker-compose project name.
     COMPOSE_PROJECT_STARTED: Action[[str, Config, str]] = Action()
 
+    #: This action is called at the end of the tutor.config.load_full function.
+    #: Modifying this object will not trigger changes in the configuration.
+    #: For all purposes, it should be considered read-only.
+    #:
+    #: :parameter dict config: project configuration.
+    CONFIG_LOADED: Action[[Config]] = Action()
+
     #: Called whenever the core project is ready to run. This action is called as soon
     #: as possible. This is the right time to discover plugins, for instance. In
     #: particular, we auto-discover the following plugins:


CodeWithEmad:
@regisb Should I also consider increasing the version in this pull request?

tutor/commands/config.py
```diff
@@ -185,6 +189,7 @@ def save(
     # Reload configuration, without version checking
     config = tutor_config.load_full(context.root)
     env.save(context.root, config)
+    tutor_hooks.Actions.CONFIG_LOADED.do(deepcopy(config))
```
regisb, line 192: Can you move this action inside `load_full`? The goal is to enable anyone to check the configuration whenver the configuration is loaded, not just when we run `tutor config save`.

tutor/config.py
```diff
@@ -26,6 +26,15 @@ def load(root: str) -> Config:
     return load_full(root)
 
 
+def load_defaults(root: str) -> Config:
+    """
+    Load default configuration.
+    """
+    config = get_user(root)
```
regisb, line 33: We should not be loading user config here. We only need the defaults. Thus the `root` argument is unnecessary.

tutor/config.py
```diff
@@ -319,3 +328,21 @@ def _update_enabled_plugins_on_unload(_plugin: str, _root: str, config: Config)
...
+
+
+@hooks.Actions.CONFIG_LOADED.add()
+def _check_preview_lms_host(config: Config) -> None:
+    """
+    This will check if the PREVIEW_LMS_HOST is a subdomain of LMS_HOST.
+    if not, prints a warning to notify the user.
+    """
+
+    lms_host = str(config.get("LMS_HOST"))
```
regisb, line 340: `lms_host = config.get("LMS_HOST", "")`

tutor/config.py
```diff
@@ -319,3 +328,21 @@ def _update_enabled_plugins_on_unload(_plugin: str, _root: str, config: Config)
...
+    preview_lms_host = config.get("PREVIEW_LMS_HOST")
```
regisb, line 341: `preview_lms_host = config.get("PREVIEW_LMS_HOST", "")`

tutor/config.py
```diff
@@ -319,3 +328,21 @@ def _update_enabled_plugins_on_unload(_plugin: str, _root: str, config: Config)
...
+    if not preview_lms_host.endswith("." + lms_host):
+        fmt.echo_alert(
+            f"Warning: PREVIEW_LMS_HOST does not appear to be a subdomain of LMS_HOST.\n"
+            "This configuration is not typically recommended and may lead to unexpected behavior.\n"
+            f"LMS_HOST: {lms_host}\n"
+            f"PREVIEW_LMS_HOST: {preview_lms_host}\n"
```
regisb, line 347: Let's rewrite this such that the warning fits on a single line. I propose the following:

```
f"Warning: PREVIEW_LMS_HOST=\"{preview_lms_host}\" is not a subdomain of LMS_HOST=\"{lms_host}\"."
"This configuration is not typically recommended and may lead to unexpected behavior."
```

tutor/hooks/catalog.py
```diff
@@ -57,6 +57,11 @@ def run_this_on_start(root, config, name):
     #: :parameter str name: docker-compose project name.
     COMPOSE_PROJECT_STARTED: Action[[str, Config, str]] = Action()
 
+    #: Called whenever the config is loaded. This action is called at the end of the save command.
+    #:
+    #: :parameter dict config: project configuration.
```
regisb, line 62: Please specify that modifying this object will not trigger changes in the configuration. For all purposes, it should be considered read-only.

regisb (changes requested):
Please add a changelog entry with `make changelog-entry`. You should not bump the version number. https://docs.tutor.overhang.io/tutor.html#contributing

tutor/config.py
regisb: Actually, you should fix type tests by writing:

```
from tutor.types import get_typed
lms_host = get_typed(config, "LMS_HOST", str, "")
```

Same for `PREVIEW_LMS_HOST` below.

tutor/commands/config.py
CodeWithEmad: Sure.

CodeWithEmad:
>  You should not bump the version number.

sure, but we need to have a new version, if we want to use the CONFIG_LOADED action in the mfe plugin and increase update the tutor requirement there from `tutor>=16.0.0` to something like `tutor>=16.1.2`

CodeWithEmad:
Hi @regisb. Just following up on this to see if you're able to review this.
Thanks.

tutor/config.py
```diff
@@ -319,3 +330,19 @@ def _update_enabled_plugins_on_unload(_plugin: str, _root: str, config: Config)
...
+    """
+    This will check if the PREVIEW_LMS_HOST is a subdomain of LMS_HOST.
+    if not, prints a warning to notify the user.
+    """
+
+    lms_host = get_typed(config, "LMS_HOST", str, "")
+    preview_lms_host = get_typed(config, "PREVIEW_LMS_HOST", str, "")
+    if not preview_lms_host.endswith("." + lms_host):
+        fmt.echo_alert(
+            f'Warning: PREVIEW_LMS_HOST="{preview_lms_host}" is not a subdomain of LMS_HOST="{lms_host}".'
```
regisb, line 346: Nit: append a space at the end: `... is not a subdomain of LMS_HOST="{lms_host}". '`

CodeWithEmad: Done. And what about the version update?

regisb (approved):
thanks for the quick turnaround!

tutor/config.py
regisb: We'll publish a new version ourselves.

changelog.d/20230926_172753_codewithemad_config_loaded_action.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] New action introduced: CONFIG_LOADED. This actions is called whenever the config is loaded. (by @CodeWithEmad)
```
fghaas, line 1: Got a notification for this because I was watching #557, so I took a look. Thanks for adding this! 

One minor note though: are we sure that this should be marked as a breaking change (💥)? It doesn't look to me like this change has the potential to break anything, since at worst it generates a warning — or am I misreading something?

regisb: You're absolutely right. I'll remove the 💥 flag when I create the new version.

CodeWithEmad: Oops. My bad.

PR #905, opened 2023-09-28
Fix command to Transfer the configuration, environment, and platform …

thasipablo:
I added `rsync` to Transfer the configuration, environment, and platform data from server 1 to server 2 command so that we can we able to transfer data. I found that `-avr` options fit well to it.


regisb (approved):
Nice catch, thanks a lot!

thasipablo:
You’re welcome @regisb

PR #906, opened 2023-10-02
refactor: use customizable registry for images

gabor-boros:
## Description

This PR allows the possibility to override what should be the docker registry used for pulling images. This feature was available only for the edx image, though it is needed on a bigger scale, for all the images that can be pulled.

The reasoning behind this is allowing users to setup instances and registries with restricted network access, setup pull-thorough caches, docker proxies, etc. Due to buildx is incapable of reusing local images, it is not an option to pre-pull the needed images.

As there were already a `DOCKER_REGISTRY` setting, I decided to reuse it, though it introduces a small change, the `DOCKER_REGISTRY` is not ending with a slash anymore in the config. This requires those who were using a custom registry to update their config, though in the code it reads better `${DOCKER_REGISTRY}/python:3.8-slim` rather than `${DOCKER_REGISTRY}python:3.8-slim` (notice the slash in between the registry and image name).

The documentation is updated/extended accordingly.


regisb:
Hi @gabor-boros,
I really wish you had started a conversation about this before opening 8 different pull requests. There was an extended conversation about that very same question a few months ago: https://github.com/overhangio/tutor/pull/830

In particular, you should be aware that your changes would be a backward-incompatible change for all users who ever modified `DOCKER_REGISTRY`, for the following two reasons:

1. The trailing slash must now be removed.
2. If someone pushed/pulled their openedx image to a custom registry, the setting will then also apply to all other images.

The conversation I linked to above outlines multiple alternative and simpler solutions. Please consider them. In particular:

> Due to buildx is incapable of reusing local images, it is not an option to pre-pull the needed images.

I may be wrong, but I believe this is incorrect. It would help if you could clarify what you mean by that.

gabor-boros:
@regisb 

> I really wish you had started a conversation about this before opening 8 different pull requests.

Yeah, that could have saved some time.

> you should be aware that your changes would be a backward-incompatible

Yes, I'm aware of that, that's why I was mentioning it.

> The conversation I linked to above outlines multiple alternative and simpler solutions.

I was considering and trying them before. I was preferring using something else than docker subscription, and the daemon modification was not really a choice in my scenario.

> I may be wrong, but I believe this is incorrect.

I was surprised too that it is not working as expected. I was trying it on GitLab CI. The images were pulled from the proxy, then retagged them to have the expected repository, like:

`docker tag <registry>/<image>:<tag> docker.io/<image>/<tag>`

These were the images pulled before the build:

Screenshot 2023-10-02 at 10 04 58

And here is the build failure showing that `COPY --from` triggered a rate limit:

Screenshot 2023-10-02 at 10 04 49

However, when I tried it locally now,  with a simple scenario, it was reusing the local images, indeed.

Screenshot 2023-10-02 at 10 16 42

So it seems that the images are not reused only in CI.

regisb:
> So it seems that the images are not reused only in CI.

I'm not sure about that. Can you please verify by turning off internet access on your local computer?

> I was preferring using something else than docker subscription

Why? Unless I'm mistaken, you will still have to `docker login` to your self-hosted registry, right?

> and the daemon modification was not really a choice in my scenario.

Why not?

gabor-boros:
> > So it seems that the images are not reused only in CI.

> I'm not sure about that. Can you please verify by turning off internet access on your local computer?

I meant the other way around. It seems that the CI did not used the local image, but my computer did.

> > I was preferring using something else than docker subscription

> Why? Unless I'm mistaken, you will still have to docker login to your self-hosted registry, right?

I may missed something here, but having a docker subscription and running `docker-login` are two different things. Yes, we still have to do `docker login` to the self-hosted registry but that's a different story. With the self-hosted registry, I can setup helm charts as well, for example, not just docker containers, so why would I want to have a docker subscription?

> > and the daemon modification was not really a choice in my scenario.

> Why not?

If the registry-mirror contains slashes, the mirror won't be used as expected. To fix that, the one needs to set up an additional reverse proxy, ensuring the rewrite of target path to append the necessary path. For example, if your registry mirror is `https://example.com/dockerhub-proxy`, then the registry won't be used, unless you setup a reverse proxy to rewrite from `https://example.com` to `https://example.com/dockerhub-proxy` so you can set `https://example.com` in your daemon config. See more here: [ref](https://github.com/moby/moby/issues/36598#issuecomment-1435119078)

---

In general, I still do think that the ability to set a custom registry would be a great feature for tutor. The change can be done to be backward compatible with a little bit of care. I'm pretty sure this feature would help many people in the community.

What do you think?

gabor-boros:
> Yes, that's what I understood. I think that you should verify that buildx is indeed not making any call to docker.io on your computer, because I suspect it is doing it without informing you in the logs. I suggest you do so by cutting off your internet access.

Turned off the network interface and it is still using the cache. Weird that Gitlab is not acting the same (i.e. using the cache).

> My point was that you will have to login to your self-hosted registry in any case -- unless it's completely public or hidden behind a VPN. So you will have to go to the trouble of logging in from your CI. Why not start now with docker.io and resolve your 429 issues?

As I mentioned, in our case, having a registry would allow us hosting Helm charts as well. Docker hub does not allow that (and many more things). Logging in to any registry is not a trouble at all, it is `docker login -u <USER> -p <PASSWORD> <registry>`.

> I wasn't aware of that. But it seems to me that this would be a rather small devops issue that affects few users. Wouldn't you agree?

With all the respect, I don't agree. This is an issue for hundreds of people globally. Although we are just a subset of them, the issue affects many people.

I wouldn't agree with the "small devops issue" part either, as it requires setting up and maintaining a reverse proxy just because of this issue with Docker.

> Why? What's the point of self-hosting images of elasticsearch, mysql, etc.?

I feel we have a misalignment here. The ultimate goal is not self-hosting but having a pull-through cache. Self-hosting is a side-effect.

> The only use case that I'm seeing is performance.

Possible use cases can be

* using a custom registry for the sake of unification (think about Nexus or Harbor)
* using custom security scanners on docker images
* using a pull-thorugh cache
* having registries close to the environment it will be used (a.k.a. performance)
* setting up instances in a restricted or highly controlled environment (like military facilities -- although we did not encounter this scenario, I had to deal with situations like this in the past in the aviation industry)

I can list more use cases if I think more about it.

> But you can get that by running a self-hosted mirror (with {"registry-mirrors": ["http://localhost:5000"]})

As we discussed above, that is working only with the conditions above that I would like to avoid as the proper, recommended solution is to be able to configure the registry.

---

To understand your point better, could you please help me what's the reason of the push-back if the change can be done in a backward compatible manner and would help more developers in the community (as you wrote, there were other conversations about this before this one).

regisb:
> To understand your point better, could you please help me what's the reason of the push-back if the change can be done in a backward compatible manner and would help more developers in the community (as you wrote, there were other conversations about this before this one).

Part of my reaction is pushback, and part is trying to understand your use case. 

The pushback is for the following reasons:

1. You opened 8 different, non-backward compatible PRs without talking about it first, or researching previous conversations. But that's OK, I got over it. Let's move on to the next reason.
2. Even if we do make these PRs backward-compatible, the changes will introduce yet another complexity for plugin developers and platform administrators. If possible at all, I would like to avoid that.

The reason why I'm trying to understand your use case, is to suggest alternative solutions that do not make life more difficult for everyone else.

Introducing a new configuration `DOCKER_IO` setting as I suggested [here](https://github.com/overhangio/tutor/pull/830#issuecomment-1533321203) is one option, but it's important to understand it's not free. With this new setting, each and every plugin developer will have to remember that they should not write "docker.io" but "{{ DOCKER_IO }}" in their templates. This is _not_ a minor change.

At first, it seems not too difficult to introduce new settings to cover each and every use case. But complexity of the project grows as a quadratic (roughly) function of the number of settings. If you've worked with edx-platform settings then you know what I mean... This is why I generally want to avoid introducing new configuration settings.

Instead, the recommended approach with Tutor with to make it easy for users to run Open edX in the default scenario, and then enable them to override the default using plugins.

In a nutshell, this is why I'm trying to understand your use case. 

Now, what I do understand is that you do not want to connect to docker.io at all: neither during image building or pulling. This is what you meant with: 

> The ultimate goal is not self-hosting but having a pull-through cache.

I'm not challenging that. What I think I would do in your position is not relevant. You want to do that, fine. Let's do it. Can we try to write a plugin to do that? That would be ideal for everyone, right? That plugin would have to replace "docker.io" by your own registry host in:

1. every configuration setting
2. every template

I think 1. might be possible already. To implement 2. we would have to create a new filter, but that would be easy enough.

Would this be an acceptable solution? (let's call this solution **candidate A**)

Now, let's consider an alternative solution: point the Docker service to a registry mirror. I understand that Nexus and Harbor are not compatible off the shelf with the `{"registry-mirrors": ["https://registryhost.com/subpath/"]}` solution. If we did manage to point the Docker daemon to Nexus/Harbor, either by fixing the upstream bug or by making a change to the Nexus/Harbor configuration, would that be an acceptable solution? (let's call this solution **candidate B**)

bradenmacdonald:
> So it seems that the images are not reused only in CI.

@gabor-boros Most likely the reason is [this](https://github.com/docker/buildx/issues/159#issuecomment-539238262): 

> Buildx can only access the local docker images with docker driver. Run `buildx inspect` to see what driver is being used.

I would guess you are using the default `docker` driver locally but `docker-container` on CI. That would explain your observations.

------------------------

@regisb @gabor-boros The ultimate problem we are currently trying to solve is a 429 rate limit issue. We are trying to finish up a project that is going to dramatically increase our usage of Tutor-based image builds and k8s deployments, yet we are already hitting this rate issue just in the development and testing phase. When we deploy to production, it's possible we would sooner or later hit even the 5,000 pulls per day limits of the paid Docker accounts, and increasing beyond that limit [starts at](https://docs.docker.com/docker-hub/service-accounts/#enhanced-service-account-add-on-pricing) $10,000 per year. So it seems like if there's a way to quickly deploy our own container registry and use that as a cache, it would be a much more affordable option for us to fix the 429 issues we're facing. Of course, if our own container registry would require too much developer time and maintenance, then it would make more sense to just pay Docker for the increased limits. So that's an open question. It does seem however, that setting up our own registry is a simple, effective, and affordable solution to the problem, if we can find a good approach for getting Tutor and plugins to use it.

Trying to recap the options here:

A. Extend Tutor so that plugins have even more power - to filter every configuration setting and template, as described above. Requires some development to implement, but could provide a powerful new way to achieve this sort of thing in the future, without introducing additional settings. Of course, if the plugin is used by a few different organizations to modify the same aspect of Tutor (e.g. five different Open edX Providers use a `change-docker-io` plugin) that shifts the maintenance burden onto the plugin maintainer rather than Regis, but IMHO it still acts as a de facto configuration setting that increases the combinatorial complexity of the Tutor ecosystem. That said, it does keep the default, out of the box experience+config simple, and it does let power users do more to customize Tutor, so I think I overall like the idea.

B. Leave the code as-is, but configure our environment to use a mirror automatically - set [`"registry-mirror"` in the `dind` config on the CI runners](https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#enable-registry-mirror-for-dockerdind-service), and use something like [`k8s-image-swapper`](https://estahn.github.io/k8s-image-swapper/v1.5/index.html) on k8s. This approach [is not directly compatible with any registry that requires a slash in its URL](https://blog.smile.eu/en/devops/how-configure-docker-hub-proxy-harbor) which currently rules out Harbor. However, I believe that Nexus [specifically offers a default registry option](https://help.sonatype.com/repomanager3/nexus-repository-administration/formats/docker-registry/ssl-and-repository-connector-configuration) that doesn't require a slash, and [the normal docker registry/proxy](https://docs.docker.com/registry/recipes/mirror/) can easily be run without a slash. We will need to investigate this further. I'm also not sure about the overhead of configuring and using `k8s-image-swapper` nor how well it works for this use case.

C. Introduce a new config option, `DOCKER_IO`. This is straightforward and backwards compatible, but requires making a lot of changes anywhere that's used in the codebase, introduces a new setting in the base installation that complicates maintenance and increases complexity, and would only work with those plugins that also make the same change. As the maintainer, @regisb would prefer to avoid this option if possible.

D. Pay Docker for increased limits. Solves our problem but perhaps at too high of cost, and doesn't solve the same problem for anyone else. As we know, others have hit this same issue with Tutor.

regisb:
I'd like to introduce option E: resolve the [upstream issue](https://github.com/moby/moby/issues/36598) by proposing a fix. According to my preliminary tests, this could be as simple as removing the mirror path check. I was able to run a docker daemon that supports a  `{"registry-mirrors": ["http://localhost:5001/hub"]}` configuration in daemon.json. Sure, there might be something that I'm missing, as I'm not familiar with the moby internals or the consequences on other projects, but that could be a great starting point.

If you agree, you could go with option B until we're through with E.

The more I think about it, the more I believe that option C is less than optimal. You might resolve the 429 errors with Open edX images, but then you'll face the same issue with other images in non-Open edX projects.

Option A is possible, but it's a Pandora box, as we've seen with the comprehensive theming approach in edx-platform. If we allow plugin developers to modify all parts of all templates, there is a risk that it will become the default solution to override just about anything. And this solution makes it very difficult to upgrade plugins.

Option D is terrible, for obvious reasons.

bradenmacdonald:
@regisb Nice, thanks for opening the PR! We'll look closer at option B and see if it's doable or not.

brian-smith-tcril:
Would splitting `DOCKER_REGISTRY` up work as a way to avoid backwards compatibility issues?

Something like: 
```yml
DOCKER_REGISTRY: "docker.io/"
DOCKER_REGISTRY_OPENEDX: "{{ DOCKER_REGISTRY }}"
DOCKER_REGISTRY_OTHER: "docker.io/"
```

Then replace all current instances of `DOCKER_REGISTRY` with `DOCKER_REGISTRY_OPENEDX` and hardcoded references to `docker.io` for non Open edX images with `DOCKER_REGISTRY_OTHER`.

I'd think this would provide the added flexibility in configuration without breaking existing configurations.

regisb:
@brian-smith-tcril yes, that change would be backward compatible. But I'm not in favor of this approach. See my comment above:

> Introducing a new configuration DOCKER_IO setting as I suggested https://github.com/overhangio/tutor/pull/830#issuecomment-1533321203 is one option, but it's important to understand it's not free. With this new setting, each and every plugin developer will have to remember that they should not write "docker.io" but "{{ DOCKER_IO }}" in their templates. This is not a minor change.
>
> At first, it seems not too difficult to introduce new settings to cover each and every use case. But complexity of the project grows as a quadratic (roughly) function of the number of settings. If you've worked with edx-platform settings then you know what I mean... This is why I generally want to avoid introducing new configuration settings.
>
> Instead, the recommended approach with Tutor with to make it easy for users to run Open edX in the default scenario, and then enable them to override the default using plugins.

gabor-boros:
> Nice, thanks for opening the PR! We'll look closer at option B and see if it's doable or not.

We will go with Option B until moby merges the PR.

regisb:
Hi @gabor-boros, I wonder if you noticed [Cory's comment](https://github.com/moby/moby/pull/46594#issuecomment-1751238705):

> the [Docker image reference](https://pkg.go.dev/github.com/distribution/reference) syntax used in Dockerfiles and on the docker command line only has affordances for registries hosted at the root of a domain.

If I understand correctly, setting the registry mirror in daemon.json will not affect how images are pulled during the image build. Is this also your understanding? Is it an issue for you?

gabor-boros:
I may be a bit confused in this implementation, but if I understand correctly, this means that if one wants to pull a namespaced image, the whole path should be provided. However, the registry mirror will be able to work as a pull through cache. Right?

If not, then I lost track of what the recently merged change solves, as you already tested image building with the mirror defined in daemon.json.

regisb:
I did not test building the image. I incorrectly assumed that the registry mirror would be used both by `docker pull` and `docker build`. But I'm not sure it's actually the case. Can you please give it a try?

gabor-boros:
I do not have the time for that nowadays, but hopefully in ~2 weeks I can.

gabor-boros:
@regisb It seems the fix will be released as part of https://github.com/moby/moby/milestone/119

PR #912, opened 2023-10-04
fix: override CMS_BASE setting in Studio for dev

ormsbee:
The LMS was overriding CMS_BASE properly, but Studio (CMS) configuration was not. That meant that Studio's CMS_BASE in dev mode was using the devstack default of localhost:18010 (because this is what's defined in edx-platform). This in turn broke parts of Studio that use this value, such as the XBlock v2 API (/api/xblock/v2).

This commit derives the value of the CMS_BASE Django setting from Tutor's CMS_HOST config value, in the same way that the LMS does it.

### DIFF
--- a/tutor/templates/apps/openedx/settings/cms/development.py
+++ b/tutor/templates/apps/openedx/settings/cms/development.py
@@ -5,6 +5,9 @@ from cms.envs.devstack import *
 LMS_BASE = "{{ LMS_HOST }}:8000"
 LMS_ROOT_URL = "http://" + LMS_BASE
 
+CMS_BASE = "{{ CMS_HOST }}:8001"
+CMS_ROOT_URL = "http://" + CMS_BASE
+
 # Authentication
 SOCIAL_AUTH_EDX_OAUTH2_KEY = "{{ CMS_OAUTH2_KEY_SSO_DEV }}"
 SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT = LMS_ROOT_URL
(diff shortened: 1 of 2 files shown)


tutor/templates/apps/openedx/settings/cms/development.py
```diff
@@ -5,6 +5,9 @@
 LMS_BASE = "{{ LMS_HOST }}:8000"
 LMS_ROOT_URL = "http://" + LMS_BASE
 
+CMS_BASE = "{{ CMS_HOST}}:8001"
+CMS_ROOT_URL = "http://{}".format(CMS_BASE)
```
kdmccormick, line 9: Nit: let's make this match the style of LMS_BASE above (functionally equivalent)
```suggestion
CMS_BASE = "{{ CMS_HOST }}:8001"
CMS_ROOT_URL = "http://" + CMS_BASE
```

kdmccormick (changes requested):
Thanks for the fix. Just one style nit.

To make sure that Palm dev users also get this fix, could you change the branch & PR bases to master? Once we merge to master, Régis's automation will forward-port it to nightly.

tutor/templates/apps/openedx/settings/cms/development.py
ormsbee: Oh, sure. I literally just copy-pasted these lines from the other config file.

ormsbee:
> To make sure that Palm dev users also get this fix, could you change the branch & PR bases to master? Once we merge to master, Régis's automation will forward-port it to nightly.

Ah, that makes sense. I'll do that.

ormsbee:
@kdmccormick: rebased

kdmccormick (approved):

regisb:
Thanks for the PR @ormsbee!

PR #913, opened 2023-10-05
feat!: assume BuildKit is available

kdmccormick:
## Description

BuildKit replaces and improves upon the legacy Docker builder, which was deprecated back in Feb 2023. Assuming that BuildKit is enabled allows us to seriously simplify the Dockerfile, and it also unlocks future build performance improvements based on BuildKit-specific features such as:

* COPY --link
* Multiple build contexts
* More aggressive parallelization during the build process

The Docker versions which Tutor recommends (v20+) all come with BuildKit, so this should be a seamless transition for Tutor users who have been installing the recommended version.

We leave the ``is_buildkit_enabled`` template variable in place in order to avoid breaking plugins, but we now just set it to ``True`` as a constant. It will be removed some time after Quince, so plugins should remove any uses of the variable from their Dockerfiles.

Relevant discussion:
https://github.com/overhangio/tutor/pull/868#issuecomment-1657961059

## Testing

I've been passively testing this for while. Specifically, I have been using https://github.com/kdmccormick/tutor/pull/34 (which is based off of this branch) to work on Open edX features for the past couple months, using the following:

 * both tutor-local and tutor-dev modes  
 * both on Ubuntu 22.04 (x86-64) and macOS Ventura (arm64)
 * Docker v24.x + Compose v2.2.x

I haven't been testing the images in production-like tutor-local or tutor-k8s deployments. I don't believe the legacy->BuiltKit switch would have any impact that is specific to those modes, but I welcome additional testing if there is any concern.


tutor/images.py
```diff
@@ -4,8 +4,7 @@
 def build(path: str, tag: str, *args: str) -> None:
     fmt.echo_info(f"Building image {tag}")
     build_command = ["build", f"--tag={tag}", *args, path]
-    if utils.is_buildkit_enabled():
-        build_command.insert(0, "buildx")
+    build_command.insert(0, "buildx")
```
kdmccormick, line 7: Adding a comment about the `buildx` specifier.
```suggestion
    # Docker v19-v22 makes the new BuildKit builder available, but only if `buildx` is specified.
    # Starting in Docker v23, BuildKit is the default, and `buildx` can be omitted.
    build_command.insert(0, "buildx")
```

tutor/env.py
```diff
@@ -54,7 +54,8 @@ def _prepare_environment() -> None:
             ("HOST_USER_ID", utils.get_user_id()),
             ("TUTOR_APP", __app__.replace("-", "_")),
             ("TUTOR_VERSION", __version__),
-            ("is_buildkit_enabled", utils.is_buildkit_enabled),
+            # For backwards compatability. Remove after Quince.
```
regisb, line 57: I'm thinking we should add some sort of keyword in the comments to indicate future maintainers when to remove this. Something like `DEPR v18`. What do you think?

That keyword should also be added to tutor.images.build (see change below).

regisb (changes requested):
Just a very minor comment and we should be good to go. Thanks!

tutor/env.py
kdmccormick: @regisb Good thinking. Though I think `DEPR v18` would leave room for ambiguity--is it deprecated in v18, or should it be removed in v18, or removed after v18?

How about `REMOVE-AFTER-Vxx`? For example, for this template filter, we're leaving it in for Quince (v16) but it'll be gone in Redwood, so it'd be `REMOVE-AFTER-V16`.

regisb: "REMOVE-AFTER" looks great to me!

kdmccormick: Great, I just pushed a new commit: https://github.com/overhangio/tutor/pull/913/commits/551c6312ad3a526770fa61906e3346b28aff818c

kdmccormick:
@regisb Did you want to re-review this, or is it good to go?

regisb (approved):
thanks Kyle!

Issue #909, opened 2023-10-02
ORA2 file upload in CMS triggers "InvalidCacheBackendError: The connection 'ora2-storage' doesn't exist" exceptions


FatemeKhodayari:
**Bug description**
Uploading a file to an open assessment unit fails with a 500 error to `upload_url`:

![image](https://github.com/overhangio/tutor/assets/55655542/5f064f67-3afa-4601-a502-5e2054943ba5)

The logs from the CMS container are as follows:
```log
Traceback (most recent call last):
  File "/openedx/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/openedx/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/openedx/venv/lib/python3.8/site-packages/sentry_sdk/integrations/django/views.py", line 84, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
  File "/opt/pyenv/versions/3.8.15/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/openedx/venv/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/openedx/venv/lib/python3.8/site-packages/django/views/decorators/clickjacking.py", line 50, in wrapped_view
    resp = view_func(*args, **kwargs)
  File "/openedx/edx-platform/./cms/djangoapps/contentstore/views/preview.py", line 75, in preview_handler
    resp = instance.handle(handler, req, suffix)
  File "/openedx/venv/lib/python3.8/site-packages/xblock/mixins.py", line 84, in handle
    return self.runtime.handle(self, handler_name, request, suffix)
  File "/openedx/edx-platform/xmodule/x_module.py", line 1033, in handle
    return super().handle(block, handler_name, request, suffix=suffix)
  File "/openedx/venv/lib/python3.8/site-packages/xblock/runtime.py", line 1081, in handle
    results = handler(request, suffix)
  File "/openedx/venv/lib/python3.8/site-packages/xblock/mixins.py", line 63, in wrapper
    response = func(self, request_json, suffix)
  File "/openedx/venv/lib/python3.8/site-packages/openassessment/xblock/submission_mixin.py", line 504, in upload_url
[...]

PR #914, opened 2023-10-07
fix: ora2 uploads in cms

FatemeKhodayari:
The ORA2 cache settings seem to never have existed in CMS, causing ORA2 file uploads fail on CMS side. As the cache settings are shared between both CMS and LMS, the settings are moved from `common_lms.py` to `common_all.py`. This issue was initially raised [here](https://discuss.openedx.org/t/palm-2-ora-file-upload-failure/11332).

Closes https://github.com/overhangio/tutor/issues/909.

### DIFF
--- a/changelog.d/20231007_153252_fateme.khodayari97.md
+++ b/changelog.d/20231007_153252_fateme.khodayari97.md
@@ -1,0 +1,2 @@
+- [Bugfix] Fix ORA2 file uploads in CMS. As the cache settings are shared between CMS and LMS, the settings are moved 
+from `common_lms.py` to `common_all.py`. (by @FatemeKhodayari)
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -77,6 +77,11 @@ CACHES = {
         "BACKEND": "django_redis.cache.RedisCache",
         "LOCATION": "redis://{% if REDIS_USERNAME and REDIS_PASSWORD %}{{ REDIS_USERNAME }}:{{ REDIS_PASSWORD }}{% endif %}@{{ REDIS_HOST }}:{{ REDIS_PORT }}/{{ OPENEDX_CACHE_REDIS_DB }}",
     },
+    "ora2-storage": {
+        "KEY_PREFIX": "ora2-storage",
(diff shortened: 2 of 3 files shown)


regisb (approved):
Looks good to me! Did you verify that the changes work on your platform?

FatemeKhodayari:
To be honest, I tried to test it by creating tutor executable using `make bundle` and then `dist/tutor local launch`. But it gives me these errors:

```log
Error applying action: func=<function discover_module.<locals>.load at 0x7f873ee484c0> contexts=['plugins']'
Error applying action: func=<function discover_module.<locals>.load at 0x7f873ee489d0> contexts=['plugins']'
⚠️  Failed to enable plugin 'mfe': plugin 'mfe' is not installed.
```

regisb:
Here's how I would test the changes:

1. Install my local fork of tutor: `pip install -e .`
2. Update my environment: `tutor config save`
3. Restart the platform: `tutor local reboot`
4. Try to upload files from the Studio
5. Try to upload files from the LMS

Thanks for taking the time to verify that your changes work. This is a very important step.

FatemeKhodayari:
Thanks for the test steps but I'm still getting the `⚠️  Failed to enable plugin 'mfe': plugin 'mfe' is not installed.` error. Having this, I can't see any of the courses in the learning MFE, thus can't upload files in LMS.

regisb:
To install the mfe plugin, run: `tutor plugins install mfe`. Then, `tutor config save` and `tutor local reboot`.

FatemeKhodayari:
Thanks for the help. I finally could test and unfortunately I had missed the `ORA2_FILEUPLOAD_ROOT` in cms which I added in commit https://github.com/overhangio/tutor/pull/914/commits/bc53a3a30232b4530df2d609773ae7effeb79d5b . I didn't add a changelog entry for this commit as I wasn't sure if its needed or not. 

Also I have a question. While adding `ORA2_FILEUPLOAD_ROOT` to cms, I realized that in lms there's a `DATA_DIR` that is not present in cms. What is it used for? Does this must be added to cms as well?

regisb:
> Also I have a question. While adding ORA2_FILEUPLOAD_ROOT to cms, I realized that in lms there's a DATA_DIR that is not present in cms. What is it used for? Does this must be added to cms as well?

I'm not quite sure anymore. Please don't change it if you don't need it :)

PR #915, opened 2023-10-09
feat: auto-mount edx-platform python requirements

regisb:
These changes make to possible to run:

    tutor mounts add /path/to/my-xblock

The xblock directory with then be auto-magically bind-mounted in the "openedx" image at build time, and the lms*/cms* containers at run time.

This makes it effectively possible to work as a developer on edx-platform requirements.

We take the opportunity to move some openedx-specific code to a dedicated module.

Close https://github.com/openedx/wg-developer-experience/issues/177

### DIFF
--- a/docs/tutorials/edx-platform.rst
+++ b/docs/tutorials/edx-platform.rst
@@ -1,0 +1,177 @@
+.. _edx_platform:
+
+Working on edx-platform as a developer
+======================================
+
+Tutor supports running in development with ``tutor dev`` commands. Developers frequently need to work on a fork of some repository. The question then becomes: how to make their changes available within the "openedx" Docker container? 
+
+For instance, when troubleshooting an issue in `edx-platform <https://github.com/openedx/edx-platform>`__, we would like to make some changes to a local fork of that repository, and then apply these changes immediately in the "lms" and the "cms" containers (but also "lms-worker", "cms-worker", etc.)
+
+Similarly, when developing a custom XBlock, we would like to hot-reload any change we make to the XBlock source code within the containers.
+
+Tutor provides a simple solution to these questions. In both cases, the solution takes the form of a ``tutor mounts add ...`` command.
+
+Working on the "edx-platform" repository
+----------------------------------------
+
+Download the code from the upstream repository::
+
+    cd /my/workspace/edx-plaform
+    git clone https://github.com/openedx/edx-platform .
+
+Check out the right version of the upstream repository. If you are working on the `current "zebulon" release <https://docs.openedx.org/en/latest/community/release_notes/index.html>`__ of Open edX, then you should checkout the corresponding branch::
+
+    # "zebulon" is an example. You should put the actual release name here.
+    # I.e: aspen, birch, cypress, etc.
+    git checkout open-release/zebulon.master
+
+On the other hand, if you are working on the Tutor :ref:`"nightly" <nightly>` branch then you should checkout the master branch::
+
+    git checkout master
+
+Then, mount the edx-platform repository with Tutor::
+
+    tutor mounts add /my/workspace/edx-plaform
+
+This command does a few "magical" things 🧙 behind the scenes:
+
+1. Mount the edx-platform repository in the image at build-time. This means that when you run ``tutor images build openedx``, your custom repository will be used instead of the upstream. In particular, any change you've made to the installed requirements, static assets, etc. will be taken into account.
+2. Mount the edx-platform repository at run time. Thus, when you run ``tutor dev start``, any change you make to the edx-platform repository will be hot-reloaded.
+
+You can get a glimpse of how these auto-mounts work by running ``tutor mounts list``. It should output something similar to the following::
+
+    $ tutor mounts list
+    - name: /home/data/regis/projets/overhang/repos/edx/edx-platform
+    build_mounts:
+    - image: openedx
+        context: edx-platform
+    - image: openedx-dev
+        context: edx-platform
+    compose_mounts:
+    - service: lms
+        container_path: /openedx/edx-platform
+    - service: cms
+        container_path: /openedx/edx-platform
+    - service: lms-worker
+        container_path: /openedx/edx-platform
+    - service: cms-worker
+        container_path: /openedx/edx-platform
+    - service: lms-job
+        container_path: /openedx/edx-platform
+    - service: cms-job
+        container_path: /openedx/edx-platform
+
+Working on edx-platform Python dependencies
+-------------------------------------------
+
+Quite often, developers don't want to work on edx-platform directly, but on a dependency of edx-platform. For instance: an XBlock. This works the same way as above. Let's take the example of the `"edx-ora2" <https://github.com/openedx/edx-ora2>`__ package, for open response assessments. First, clone the Python package::
+
+    cd /my/workspace/edx-ora2
+    git clone https://github.com/openedx/edx-ora2 .
+
+Then, check out the right version of the package. This is the version that is indicated in the `edx-platform/requirements/edx/base.txt <https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/edx/base.txt>`__. Be careful that the version that is currently in use in your version of edx-platform is **not necessarily the head of the master branch**::
+
+    git checkout <my-version-tag-or-branch>
+
+Then, mount this repository::
+
+    tutor mounts add /my/workspace/edx-ora2
+
+Verify that your repository is properly bind-mounted by running ``tutor mounts list``::
+
+    $ tutor mounts list
+    - name: /my/workspace/edx-ora2
+    build_mounts:
+    - image: openedx
+        context: mnt-edx-ora2
+    - image: openedx-dev
+        context: mnt-edx-ora2
+    compose_mounts:
+    - service: lms
+        container_path: /mnt/edx-ora2
+    - service: cms
+        container_path: /mnt/edx-ora2
+    - service: lms-worker
+        container_path: /mnt/edx-ora2
+    - service: cms-worker
+        container_path: /mnt/edx-ora2
+    - service: lms-job
+        container_path: /mnt/edx-ora2
+    - service: cms-job
+        container_path: /mnt/edx-ora2
+
+You should then re-build the "openedx" Docker image to pick up your changes::
+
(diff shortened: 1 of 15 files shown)


docs/tutorials/edx-platform.rst
```diff
@@ -0,0 +1,124 @@
...
+-------------------------------------------
+
+Quite often, developers don't want to work on edx-platform directly, but on a dependency of edx-platform. For instance: an XBlock. This works the same way as above. Let's take the example of the `"edx-ora2" <https://github.com/openedx/edx-ora2>`__ package, for open response assessments. First, clone the Python package::
+
+    cd /my/workspace/edx-ora2
+    git clone https://github.com/openedx/edx-ora2 .
+
+Then, check out the right version of the package. This is the version that is indicated in the ``edx-platform/requirements/edx/base.txt``. Be careful that the version that is currently in use in your version of edx-platform is **not necessarily the latest version**.
+
+    git checkout <my-version-tag-or-branch>
```
kdmccormick, line 74:
```suggestion
::
    git checkout <my-version-tag-or-branch>
```
so that this renders as code rather than a quotation.

docs/tutorials/edx-platform.rst
```diff
@@ -0,0 +1,124 @@
+.. _edx_platform:
+
+Working on edx-platform as a developer
```
kdmccormick, line 3: This is fantastic!! 🎉 

I do think this guide should be linked to from the [Open edX development guide](https://docs.tutor.overhang.io/dev.html), if not entirely absorbed by it. I also think we ought to avoid confusion by removing the old [XBlock and edx-platform plugin development](https://docs.tutor.overhang.io/dev.html#xblock-and-edx-platform-plugin-development) section.

Furthermore, I recommend de-emphasizing the [Override docker-compose volumes](https://docs.tutor.overhang.io/dev.html#override-docker-compose-volumes) and [Manual bind-mount to any directory](https://docs.tutor.overhang.io/dev.html#override-docker-compose-volumes) sections. Perhaps those will still be useful to some advanced developers, but I think that most developers will only need to know the new instructions you wrote here. I don't want the average developer to get overwhelmed trying to understand the instructions which only now apply to power-users.

docs/tutorials/edx-platform.rst
```diff
@@ -0,0 +1,124 @@
...
+    - service: cms-worker
+        container_path: /openedx/requirements/edx-ora2
+    - service: lms-job
+        container_path: /openedx/requirements/edx-ora2
+    - service: cms-job
+        container_path: /openedx/requirements/edx-ora2
+
+It is quite possible that your package is not automatically recognized and bind-mounted by Tutor. In such a case, you will need to create a :ref:`Tutor plugin <plugin_development_tutorial>` that implements the :py:data:`tutor.hooks.Filters.EDX_PLATFORM_PYTHON_PACKAGES` patch::
+
+    tutor.hooks.Filters.EDX_PLATFORM_PYTHON_PACKAGES.add_item("my-package")
```
kdmccormick, line 105: If you include the import, then the reader could copy-paste this code block and have a mostly-working patch.
```suggestion

    from tutor import hooks
    hooks.Filters.EDX_PLATFORM_PYTHON_PACKAGES.add_item("my-package")
```

docs/tutorials/edx-platform.rst
```diff
@@ -0,0 +1,124 @@
...
+
+After you implement and enable that plugin, ``tutor mounts list`` should display your directory among the bind-mounted directories.
+
+You should then re-build the "openedx" Docker image to pick up your changes::
+
+    tutor images build openedx
```
kdmccormick, line 111:
```suggestion
    tutor images build openedx-dev
```

In order to apply to ``tutor dev start``, I believe you'll need to specifically build the dev image.

tutor/plugins/openedx.py
```diff
@@ -0,0 +1,130 @@
+from __future__ import annotations
```
kdmccormick, line 1: I like how this refactoring is going 👍🏻

tutor/hooks/catalog.py
```diff
@@ -258,6 +258,26 @@ def your_filter_callback(some_data):
...
+    #: filter. Add your own edx-platform dependencies to this filter to make it easier for
+    #: users to work on edx-platform dependencies.
+    #:
+    #: See the list of all edx-platform base requirements here:
+    #: https://github.com/openedx/edx-platform/blob/master/requirements/edx/base.txt
+    #:
+    #: :parameter list[str] names: Add here simple names ("my-package") or regular
+    #:   expressions.
+    EDX_PLATFORM_PYTHON_PACKAGES: Filter[list[str], []] = Filter()
+
```
kdmccormick, line 280: Two things:

1. This filter hook is openedx-specific, unlike every other hook in catalog.py. Are you sure you want to add this hook here, given that you are extracting openedx-specific code into its own module in this very same PR?

2. As devs migrate over from devstack, they will need to hack on packages for services other than edx-platform. I guess we could adopt a pattern where each service plugin defines its own version of this filter (e.g., `DISCOVERY_PYTHON_PACKAGES: ... = Filter()`). Alternatively, we could make this filter's name more generic, e.g. `KNOWN_PYTHON_PACKAGES: Filter[list[str], []]`, and each plugin could decide how to process the filter. This would also ensure that catalog.py doesn't have openedx-specific stuff in it.

tutor/plugins/openedx.py
```diff
@@ -0,0 +1,130 @@
...
+    return volumes
+
+
+# Auto-magically bind-mount xblock directories and some common dependencies.
+hooks.Filters.EDX_PLATFORM_PYTHON_PACKAGES.add_items(
+    [
+        r".*[xX][bB]lock.*",
+        "edx-enterprise",
+        "edx-ora2",
+        "edx-search",
```
kdmccormick, line 65:
```suggestion
    [
        r".*[xX][bB]lock.*",
        "edx-enterprise",
        "edx-ora2",
        "edx-search",
        "platform-plugin-.*",
```
The `platform-plugin-` prefix is a convention that some teams used when I was at edX. I think it's nice, and if you don't hate it, then I'd request you add it to the list :)

kdmccormick (changes requested):
This is going to be big :) I still need to test it, but in the meantime, a few suggestions.

tutor/hooks/catalog.py
kdmccormick: > As devs migrate over from devstack, they will need to hack on packages for services other than edx-platform. 

As an example, I know @rgraber has been needing to run course-discovery with a local Python package in order to work on the event bus, and had found the old Tutor workflow difficult.

regisb: A couple notes for myself:

1. Moving the filter to a separate module would make it difficult to document it. Also, developers would have to import a separate module, e.g: `tutor.openedx.hooks`.
2. We could change the filter signature to `Filter[list[tuple[str, str]], []]` to make it slightly more generic. But it would still be openedx-specific, so we should still move it to a separate module.

I don't have a great answer for now. I'll think about it some more.

regisb: I may have found a good solution. It yields the following documentation:

image

Let me know what you think.

kdmccormick: @regisb I'm not sure that solves the right problems. Tutor's docs already have plenty of Open-edX-specific details, and that's OK (probably good, actually. Otherwise, the docs would be confusingly abstract).

The issues I see with EDX_PLATFORM_PYTHON_PACKAGES is that (1) it makes it harder for us to isolate an Open-edX-agnostic core of Tutor's _codebase_, and (2) it prevents micro-services like Credentials and Course Discovery from using this improvement.

What about this?
```python
    #: List of python package names that can be potentially installed into a particular image.
    #: ...
    #: :parameter list[str] names: Add here simple names ("my-package") or regular
    #:   expressions.
    #: :parameter [str] image: Name of the image (e.g., "openedx") which should install
    #:   the packages, if mounted.
    IMAGES_PYTHON_PACKAGES: Filter[list[str], [str]] = Filter()
```

One wrinkle is that package names would need to be added to this filter for both prod and dev images (e.g., openedx and openedx-dev). We could, of course, write a low-priority filter which adds every package from openedx to openedx-dev.

regisb: > The issues I see with EDX_PLATFORM_PYTHON_PACKAGES is that (1) it makes it harder for us to isolate an Open-edX-agnostic core of Tutor's codebase, and (2) it prevents micro-services like Credentials and Course Discovery from using this improvement.

Right, I agree. Which is why I moved the filter to a dedicated openedx module in my latest changes. I wasn't so keen on doing that because I figured it would make it more difficult to document the new hooks, but the solution is acceptable (I think?).

Would you prefer I keep the hook in the tutor.hooks.catalog module?

regisb: Oooook, I had misunderstood you. Your proposition makes sense, I'll update my implementation.

docs/tutorials/edx-platform.rst
regisb: You are absolutely right. I did the following:

- Add a link to the tutorial at the top of the Open edX development guide.
- Removed the old xblock and edx-platform plugin development section
- Added a warning to the "override docker-compose volumes" section.
- Moved some content to the tutorial.

tutor/hooks/catalog.py
regisb: I have yet again updated my implementation. It feels quite generic and it can be used by any plugin developer -- though it's definitely very advanced usage.

regisb:
I feel like the current implementation has now reached a decent state, and I'd like to merge this now, such that Palm users can use that feature in the future. I know it's a bit of a rush, but Quince is scheduled for tomorrow...

pomegranited:
Hi @regisb, this will be a great for developers, thank you!

But I tried using this on my `tutor dev` stack, but it doesn't seem to be mounting my requested dirs. Does this only work for `tutor local`?

regisb:
@pomegranited what repos did you try to mount? Did the mount fail at build or run-time?

pomegranited:
@regisb My apologies -- I re-read your updated dev docs and found my error: I missed that tutor dev uses the `openedx-dev` image, so that's what I need to rebuild if I want to mount edx-platform (not `openedx`).

PR #917, opened 2023-10-10
Fix: Save configs on enable/disable plugins

CodeWithEmad:
Before this, we had to run `tutor config save` after enable/disable a plugin.
Closes https://github.com/openedx/wg-developer-experience/issues/176


tutor/commands/plugins.py
```diff
@@ -135,15 +136,13 @@ def list_command(show_enabled_only: bool) -> None:
 @click.argument("plugin_names", metavar="plugin", nargs=-1, type=PluginName())
 @click.pass_obj
 def enable(context: Context, plugin_names: list[str]) -> None:
-    config = tutor_config.load_minimal(context.root)
+    config = tutor_config.load_full(context.root)
```
regisb, line 139: Aaaaand you just broke your config.yml! If you load the full configuration, then your config.yml will also store all configuration settings, including the default ones. This means that they will not be updated whenever we modify the default values in config/defaults.yml.

Only the minimal configuration must be used to save the config.yml yml file -- whereas the full configuration is needed to update the environment.

This is one of the trickiest parts of the tutor code base, sorry you got bit :-/ I'm not sure what's the best way to bypass this issue. Maybe call the `save` command, which does things properly?

CodeWithEmad: > Aaaaand you just broke your config.yml! If you load the full configuration

😄 Yeah, you're right.

> sorry you got bit :-/ 

No worries 😄  I've always preferred a good code mystery over easy-mode anyway!

> Maybe call the save command, which does things properly?

I faced something weird. for the `enable` command, I managed to invoke `save` using this:

```
@click.command(help="Enable a plugin")
@click.argument("plugin_names", metavar="plugin", nargs=-1, type=PluginName())
@click.pass_context
def enable(context: click.Context, plugin_names: list[str]) -> None:
    if not plugin_names:
        fmt.echo_error("No plugin names provided.")
        return
    for plugin in plugin_names:
        plugins.load(plugin)
        fmt.echo_info(f"Plugin {plugin} enabled")
    tutor_config.save_enabled_plugins(config)
    tutor_config.save_config_file(context.obj.root, config)
    context.invoke(config_save_command, env_only=True)
```
But for the `disable` command, this approach fails. it tries to render parts of the removed plugin, even though it's not a part of `config.yml`. for example, after I disabled the `mfe` plugin: 

```
Error rendering patch 'caddyfile':
{{ MFE_HOST }}{$default_site_port} {
    respond / 204
    request_body {
        max_size 2MB
    }
    import proxy "mfe:8002"
}

Error rendering template apps/caddy/Caddyfile
Error: Missing configuration value: 'MFE_HOST' is undefined
```

Any thoughts?

regisb: Can you please add a unit test that reproduces the issue? Otherwise it's difficult for us to troubleshoot.

regisb:
Sorry @CodeWithEmad, I forgot to ask you to open this PR on top of nightly. This is a breaking change, so I'd rather introduce it only in nightly/quince.

regisb (changes requested):
noooooooo! I had left my comments as "pending" for a week... I hate it when github does that to me.

tutor/commands/plugins.py
```diff
@@ -133,17 +135,18 @@ def list_command(show_enabled_only: bool) -> None:
 
 @click.command(help="Enable a plugin")
 @click.argument("plugin_names", metavar="plugin", nargs=-1, type=PluginName())
-@click.pass_obj
-def enable(context: Context, plugin_names: list[str]) -> None:
-    config = tutor_config.load_minimal(context.root)
+@click.pass_context
+def enable(context: click.Context, plugin_names: list[str]) -> None:
+    if not plugin_names:
```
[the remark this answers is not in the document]
CodeWithEmad: Sure.

CodeWithEmad:
Oh. The devil's in the detail, indeed.

tutor/commands/plugins.py
```diff
@@ -153,22 +153,21 @@ def enable(context: Context, plugin_names: list[str]) -> None:
...
-            hooks.Actions.PLUGIN_UNLOADED.do(plugin, context.root, config)
+            hooks.Actions.PLUGIN_UNLOADED.do(plugin, context.obj.root, config)
             disabled.append(plugin)
             fmt.echo_info(f"Plugin {plugin} disabled")
     if disabled:
-        tutor_config.save_config_file(context.root, config)
-        fmt.echo_info(
-            "You should now re-generate your environment with `tutor config save`."
-        )
+        ENV_PATCHES_DICT.clear()
```
regisb, line 168: Not quite :) One of the advantages of having an event-driven API is that we don't need to import dependencies. In particular, we don't want to spread the use of `ENV_PATCHES_DICT` outside of its own module (tutor.plugins). So, we should try not to use that filter here.

Instead, we should add a callback to the PLUGIN_UNLOADED action. That callback will clear and re-compute the ENV_PATCHES_DICT. It will be called automatically every time a plugin is unloaded. Thus, we don't have to worry about it outside of its own module.

This is a bit difficult to explain... Am I making sense?

CodeWithEmad: Yes. Totally. You mean something like this:

``` python
@hooks.Actions.PLUGIN_UNLOADED.add()
def _clear_plugin_patches(plugin: str, root: str, _config: Config) -> None:
    """
    After a plugin is disabled, we clear the patches.
    """
    ENV_PATCHES_DICT.clear()
```

regisb: Yes. And the callback should also call a cache-filling function.

CodeWithEmad: I apologize, but I'm feeling lost again. The issue was resolved after I cleared the `ENV_PATCHES_DICT`, and the disable command is now functioning properly. If by "cache-filling" you mean executing `.iterate()` and updating the `ENV_PATCHES_DICT`, I believe that `_update_enabled_plugins_on_unload` will take care of that. However, if I am mistaken, I would appreciate some assistance.

tutor/commands/plugins.py
```diff
@@ -153,22 +153,20 @@ def enable(context: Context, plugin_names: list[str]) -> None:
 @click.argument(
     "plugin_names", metavar="plugin", nargs=-1, type=PluginName(allow_all=True)
 )
-@click.pass_obj
-def disable(context: Context, plugin_names: list[str]) -> None:
-    config = tutor_config.load_minimal(context.root)
+@click.pass_context
```
regisb, line 156: Unless I'm mistaken, we no longer have to pass the context here, right? Same for the `enable` command above.

CodeWithEmad: yes.

tutor/commands/plugins.py
CodeWithEmad: Looks like I need to spend more time on tutor codebase :)

regisb:
Closed in favour of #957

PR #921, opened 2023-10-12
fix: fix `ulimits` error for elasticsearch in Docker rootless mode

### LINKED ISSUE #876: Docker rootless: Error when setting ulimits on elasticsearch
**Bug description**

Tutor can't start elasticsearch on rootless docker due to `ulimits` section. `ulimits` section cannot be removed with `docker-compose.override.yml` once set.

**How to reproduce**
 - Run `$ tutor config save`
 - Run `$ tutor dev start elasticsearch`

The error below shows up:

```
Error response from daemon: failed to create shim task: OCI runtime create failed:
runc create failed: unable to start container process: error during container init:
error setting rlimits for ready process: error setting rlimit type 8: operation not permitted: unknown
```

Removing the `ulimits` section fixes the issue.

**Environment**
Ubuntu 22.04 with rootless docker:

```
Docker version 23.0.0, build e92dd87
Docker Compose version v2.19.1
tutor, version 16.0.2
```

Hardware: CPU i7=12700H with 32 GB of RAM.

**Suggest solutions**

 A. Fix the issue by introducing a new variable ELASTICSEARCH_SET_ULIMITS which can be set to False on rootless docker installations
 B. Set ELASTICSEARCH_SET_ULIMITS to False by default
 C. Include ulimit only on `.prod` deployments
 D. Other suggestions?

I'm planning to fix the issue myself. I've finally started to learn Tutor.

OmarIthawi:
### Description

Unset `ulimits.memlock` for elasticsearch for Docker rootless devstacks.

Related issue:

 - Fixes #876 

### Error

When running `tutor dev launch` in rootless docker, everything works except for elasticsearch. 

```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting rlimits for ready process: error setting rlimit type 8: operation not permitted: unknown
```

### TODO

 - [x] Test locally
 - [x] Rebase and open against `master`

### DIFF
--- a/changelog.d/20231116_081945_i_rootless_docker.md
+++ b/changelog.d/20231116_081945_i_rootless_docker.md
@@ -1,0 +1,2 @@
+- [Improvement] Fix `ulimits` error for elasticsearch in Docker rootless mode (by @OmarIthawi)
+
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -3,6 +3,7 @@ import os
 import tempfile
 import unittest
 from io import StringIO
+import subprocess
 from typing import List, Tuple
 from unittest.mock import MagicMock, mock_open, patch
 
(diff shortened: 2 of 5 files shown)


OmarIthawi:
@regisb this now ready for review. I've tested it locally and it works like a charm :)

tutor/commands/dev.py
```diff
@@ -23,6 +24,13 @@ def __init__(self, root: str, config: Config):
             tutor_env.pathjoin(self.root, "local", "docker-compose.override.yml"),
             tutor_env.pathjoin(self.root, "dev", "docker-compose.override.yml"),
         ]
+
+        if not utils.is_docker_rootless():
```
regisb, line 28: Let's not create a separate file just for this use case. Instead, the dev/docker-compose.yml file should include a section:

```
  {% if RUN_ELASTICSEARCH -%}
  elasticsearch:
    ulimits: {}
  {%- endif %}
```

tutor/templates/local/docker-compose.yml
```diff
@@ -57,10 +57,6 @@ services:
       - bootstrap.memory_lock=true
       - discovery.type=single-node
       - "ES_JAVA_OPTS=-Xms{{ ELASTICSEARCH_HEAP_SIZE }} -Xmx{{ ELASTICSEARCH_HEAP_SIZE }}"
-    ulimits:
```
regisb, line 60: Your change will have the effect of disabling ulimits in production, but that's not what we want, right?

regisb (changes requested):
Sorry for the long delay!

tutor/templates/local/docker-compose.yml
OmarIthawi: I'm not sure if that's the result. Please let me know how did you find that?

tutor/commands/dev.py
OmarIthawi:
```
elasticsearch:
    ulimits: {}
```

Won't work and throws the same exception. 

Setting the `ulimit` to `False` or anything else won't matter. To address your idea, the `if` statement should be added within the production file. But I think we're converging to an agreement!

OmarIthawi:
@regisb I've refactored the solution to be `dev`-only without needing a new docker-compose file.

Please let me know if that works.

regisb (approved):
Awesome!

OmarIthawi:
Thanks @regisb!! This is my first Tutor contribution :)

PR #925, opened 2023-10-20
fix: include full-complement paramters in ``JWT_PRIVATE_SIGNING_JWK``

kdmccormick:
Addresses the breaking upstream change in this commit: https://github.com/openedx/edx-platform/commit/92731be0dc69d82cc4ec5b9c511e86df9aa9e632

image

Relevant discussions:
* https://discuss.openedx.org/t/tutor-nightly-login-error-rsa-key-must-include-all-parameters-if-any-are-present-besides-d/11479
* https://openedx.slack.com/archives/C04BJV9RCPN/p1697818202936339
* https://openedx.atlassian.net/wiki/spaces/COMM/pages/3890380898/Next+Release+Redwood

See Changelog entry for details.

________________________________________

@timmc-edx , could you sanity check this?

HEADS UP REVIEWERS: I don't exactly grok how JWTs and JWKs work and whether this fix is safe. I hope one of you do. If not, we should find someone who does.


tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -189,6 +189,9 @@
         "n": "{{ jwt_rsa_key.n|long_to_base64 }}",
         "p": "{{ jwt_rsa_key.p|long_to_base64 }}",
         "q": "{{ jwt_rsa_key.q|long_to_base64 }}",
+        "dq": "{{ jwt_rsa_key.dq|long_to_base64 }}",
+        "dp": "{{ jwt_rsa_key.dp|long_to_base64 }}",
+        "qi": "{{ jwt_rsa_key.invq|long_to_base64 }}",
```
timmc-edx, line 194: Some notes on confirming that this is correct:

RSA private numbers are represented with slightly different names in different libraries. n, e, d, p, and q always seem to have the same names. [dp and dq aren't documented](https://www.pycryptodome.org/src/public_key/rsa#Crypto.PublicKey.RSA.RsaKey) in pycryptodome (which is what Tutor is using here) but apparently they're present and have a matching definition (d mod (p - 1) and likewise for q.) For qi we can compare with PyJWT's [translation from the `cryptography` library](https://github.com/jpadilla/pyjwt/blob/777efa2f51249f63b0f95804230117723eca5d09/jwt/algorithms.py#L316) and see that cryptography's `iqmp` has the [same definition](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateNumbers) as PyCryptodome's `invq`.

(The JWK spec doesn't actually define what `qi` is, but instead refers the reader to several print books.)

timmc-edx (approved):
Looks good!

regisb:
Let's merge this change and test it in quince.

regisb:
Wait... My understanding was that this change was required in Quince. Did I get that correctly @kdmccormick?

kdmccormick:
@regisb We need the change on Nightly. I don't *think* we need it in Quince since the breaking edx-platform changed merged after the cut, but it's totally OK to have it in Quince regardless.

PR #927, opened 2023-10-24
feat: introduce course discovery repository settings

Faraz32123:


tutor/templates/config/defaults.yml
```diff
@@ -20,6 +20,7 @@ DOCKER_IMAGE_PERMISSIONS: "{{ DOCKER_REGISTRY }}overhangio/openedx-permissions:{
 DOCKER_IMAGE_REDIS: "docker.io/redis:7.0.11"
 DOCKER_IMAGE_SMTP: "docker.io/devture/exim-relay:4.96-r1-0"
 EDX_PLATFORM_REPOSITORY: "https://github.com/openedx/edx-platform.git"
+COURSE_DISCOVERY_REPOSITORY: "https://github.com/openedx/course-discovery.git"
```
regisb, line 23: I believe there is some confusion. This setting will be used only when the discovery plugin is enabled, thus the setting should be part of the discovery plugin. Please close this PR and move the setting to the corresponding plugin PR.

regisb (changes requested):

PR #928, opened 2023-10-25
fix: run Celery tasks asynchronously in dev mode

kdmccormick:
In `tutor dev`, Celery tasks were running eagerly (in-process) rather than being delegated to the worker containers for asynchronous processing. This difference in behavior between `tutor dev` and `tutor local` meant that certain code paths and race-condition bugs would not surface when testing in dev.

The cause is that CELERY_ALWAYS_EAGER is set to True in the upstream devstack settings file. That's because Devstack does not support async LMS/CMS workers; it runs all those tasks in-process.

The fix is to simply override CELERY_ALWAYS_EAGER to False. We do this in the common settings file for simplicity, but the change should not affect production mode, which has always had CELERY_ALWAYS_EAGER equal False.

For unit tests, we override CELERY_ALAYS_EAGER back to True, because many of them are written to assume in-process tasks.

### DIFF
--- a/changelog.d/20231025_122158_kyle_celery_async.md
+++ b/changelog.d/20231025_122158_kyle_celery_async.md
@@ -1,0 +1,1 @@
+- [Bugfix] In development mode, Celery tasks will now be delegated to worker containers for asynchronous execution rather than eagerly exected by the lms/cms app processes. This matches the behavior of production mode and should enable more robust testing & debugging of Celery tasks. Unit tests, however, will still eagerly execute Celery tasks in-process (by @kdmccormick). 
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -235,5 +235,8 @@ CODE_JAIL = {
     "user": None,
 }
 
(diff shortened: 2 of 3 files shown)


kdmccormick:
I needed this fix in order to test the upcoming Content Libraries V2 changes, which moves all of the update-from-library operations into Celery tasks.

FYI @ormsbee.

regisb (approved):
This is a major change. I'm assuming you tested this and it works, right? It will make troubleshooting celery tasks more difficult for developers (no `pdb` for them). But on the other hand it will get us closer to production. If you think that this will be an overall improvement for developer experience, then please merge.

kdmccormick:
> It will make troubleshooting celery tasks more difficult for developers (no pdb for them). 

@regisb Let me check this out. I _think_ devs should be able to set breakpoints in Celery tasks and then interact with them using `tutor dev start [lms|cms]-worker`, but I haven't confirmed that yet. As I mentioned in Slack, I've been having trouble with breakpoints in general recently.

kdmccormick:
This PR has been beaten out by: https://github.com/overhangio/tutor/pull/1128

In retrospect, I think that's OK. When developers want to test async code, they can easily switch over to `tutor local`.

PR #929, opened 2023-10-27
chore: update deprecation warning for Django 4.2

cmltaWt0:
We are going to upgrade to Django 4.2 for Quince.
Issue: https://github.com/openedx/wg-build-test-release/issues/315
PR to edx-platform: https://github.com/openedx/edx-platform/pull/33516 (will be merged only after Tutor preparation for Django4.2)

Here I'm just updating to the new `50`/`51` DeprecationWarnings.
Also to be able to smoothly move from Django 3.2 to 4.2 I propose to add try/except block (for now).

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -149,9 +149,16 @@ LOGGING["loggers"]["blockstore.apps.bundles.storage"] = {"handlers": ["console"]
 
 # These warnings are visible in simple commands and init tasks
 import warnings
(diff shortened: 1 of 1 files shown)


regisb:
If I understand correctly, we should get rid of the lower `except` block after https://github.com/openedx/edx-platform/pull/33516 is merged, right?

cmltaWt0:
> If I understand correctly, we should get rid of the lower `except` block after [openedx/edx-platform#33516](https://github.com/openedx/edx-platform/pull/33516) is merged, right?

Yeap 👍

PR #932, opened 2023-11-06
feat: dev added to extras_require

CodeWithEmad:
We can use this to install tutor development packages inside ci jobs, with one line. it can be helpful for [this](https://github.com/overhangio/tutor-mfe/pull/155)


regisb (approved):
Looking great!

PR #940, opened 2023-11-14
feat: add surveyreport questions

Alec4r:
## DESCRIPTION
This PR creates a new hook Action that allows tutor plugins to interact with the configuration at the time of the interactive questionnaire that happens during `tutor local launch`.

## REQUIRED PLUGINS
None. The creation of the Action does not require anything from new or existing plugins. Current plugins that do not use this action will see absolutely no changes to the functionality of `tutor local launch`.

## MORE INFO
This Action was first needed when trying to create a plugin to send automatic and anonymized reports of usage.
For more context on that see:
- https://github.com/eduNEXT/tutor-contrib-surveyreport
- https://openedx.atlassian.net/wiki/spaces/OEPM/pages/3872948238/Problem+Very+few+are+going+to+generate+and+send+the+Survey+Report.

## EXAMPLE:

A plugin using this Action to add a new question can be seen at:
https://github.com/eduNEXT/tutor-contrib-surveyreport/pull/1

### DIFF
--- a/tutor/hooks/catalog.py
+++ b/tutor/hooks/catalog.py
@@ -57,6 +57,12 @@ class Actions:
     #: :parameter str name: docker-compose project name.
     COMPOSE_PROJECT_STARTED: Action[[str, Config, str]] = Action()
 
+    #: Triggered after all interactive questions have been asked.
+    #: You should use this action if you want to add new questions.
+    #:
+    #: :parameter dict config: project configuration.
+    CONFIG_INTERACTIVE: Action[[Config]] = Action()
+
     #: This action is called at the end of the tutor.config.load_full function.
     #: Modifying this object will not trigger changes in the configuration.
     #: For all purposes, it should be considered read-only.
--- a/tutor/interactive.py
+++ b/tutor/interactive.py
@@ -3,7 +3,7 @@ from typing import List, Optional
 import click
 
 from . import config as tutor_config
-from . import env, exceptions, fmt
+from . import env, exceptions, fmt, hooks
 from .types import Config, get_typed
 
 
@@ -145,6 +145,8 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
             config,
             defaults,
         )
+
+    hooks.Actions.CONFIG_INTERACTIVE.do(config)
 
 
 def ask(question: str, key: str, config: Config, defaults: Config) -> None:
--- a/changelog.d/20231214_141533_alecar.main_survey_report_plugin.md
+++ b/changelog.d/20231214_141533_alecar.main_survey_report_plugin.md
@@ -1,0 +1,1 @@
+- [Feature] add `CONFIG_INTERACTIVE` action that allows tutor plugins to interact with the configuration at the time of the interactive questionnaire that happens during tutor local launch. (by @Alec4r).


tutor/interactive.py
```diff
@@ -145,6 +145,40 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
             config,
             defaults,
         )
+    if plugins.is_loaded("surveyreport"):
```
felipemontoya, line 148: This PR in the current form is adding to the tutor core a custom way of asking questions that only apply to the survey report plugin. I don't think that would be the way of doing this feature.
Tutor is build on plugability, we need to create a new filter right here that we can later extend from the survey report.

tutor/interactive.py
```diff
@@ -145,6 +145,40 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
...
+        ask_survey_report_questions(config=config, defaults=defaults)
+
+
+def ask_survey_report_questions(config: Config, defaults: Config) -> None:
```
felipemontoya, line 152: This is the function that should go in the survey plugin

tutor/templates/config/base.yml
```diff
@@ -8,6 +8,8 @@ OPENEDX_SECRET_KEY: "{{ 24|random_string }}"
 PLUGINS:
   # The MFE plugin is required
   - mfe
+  # SurveyReport plugin is required
+  - surveyreport
```
felipemontoya, line 12: We can not allow the survey report to become a requirement for every tutor install. We have a proposal to make it part of tutor full, but first we need to finish up the plugin and make it completely optiona. Only then we can create a separate request to include it in https://github.com/overhangio/tutor/blob/master/requirements/plugins.txt

felipemontoya (changes requested):
We need to refactor this as a new filter that other plugins can also build on top of

felipemontoya:
@regisb we sat with @Alec4r in a call giving this PR a big refactor and we'd like you input about the ergonomics of the new Filter. This is probably the first filter that is being contributed so we could not find an example to draw from, so please forgive our lack of understanding.

We are looking at adding the filter at the end of the `ask_questions` [function](https://github.com/overhangio/tutor/blob/3b2373f6f14de595200a41966cd27c27369368b5/tutor/interactive.py#L147).

#### Option 1
would be to add something that lets passes the context to the filter and we expect the plugins to run the code that asks the questions. Something like:

at tutor core:
```python
utils = [ask, ask_bool, ask_choice]
config = ASK_INTERACTIVE_QUESTIONS.apply(config, defaults, utils)
```
at the plugin:
```python
@hooks.Filter.ASK_INTERACTIVE_QUESTIONS.add()
def cool_question_function(config, defaults, utils):

    utils.ask("do you want to automatically send a survey report to Axim every six months? pretty please :)", "AUTO_SEND_SURVEY_REPORT", config, defaults)

    return config
```

#### Option 2

would be to a control loop at the core that asks the question that are mostly string configured at the plugin.
Something like:

at tutor core:
```python
 for item in hooks.Filter.ASK_INTERACTIVE_QUESTIONS.get_items():
    ask(item[2], item[3], config, defaults)
```
at the plugin:
```python
hooks.Filters.ASK_INTERACTIVE_QUESTIONS.add_items(
    [
        (
            "surveyreport-namespace",
            "do you want to automatically send a survey report to Axim every six months? pretty please :)",
             "AUTO_SEND_SURVEY_REPORT",
        ),
        (
            "surveyreport-namespace",
            "Are you sure?",
             "AUTO_SEND_SURVEY_REPORT_CONFIRMATION",
        ),
    ]
)
```

Any and all guidance is very appreciated.

regisb:
I think I may have put you on the wrong lead. Instead of a filter, maybe create a `INTERACTIVE_CONFIGURATION` action that would take as argument the current configuration?

```
@hooks.Action.INTERACTIVE_CONFIGURATION.add()
def _enable_surveyreport(config):
    utils.ask("...")
    ...

def ask_questions(config, run_for_prod):
    hooks.Action.INTERACTIVE_CONFIGURATION.do(config)
```

Alec4r:
@regisb @felipemontoya I have added the actions to tutor core and also I openned a PR in surveyreport plugin https://github.com/eduNEXT/tutor-contrib-surveyreport/pull/1 adding the questions.

felipemontoya (approved):
Thanks a lot @Alec4r. This is now very much in line with what I was expecting we needed to do in the core of tutor.

The intervention seems lightweight and has an explanation at the definition of the new Action. Would you please go ahead and review this @regisb?

@Alec4r I think we need to update the PR description to make clear that this is no longer surveyreport specific but it is only opening up a space for plugins to ask question during the interactive questionnaire.

tutor/hooks/catalog.py
```diff
@@ -123,6 +123,12 @@ def run_this_on_start(root, config, name):
     #: :parameter str root: absolute path to the project root.
     PROJECT_ROOT_READY: Action[str] = Action()
 
+    #: Triggered after all interactive questions have been asked.
+    #: You should use this action if you want to add new questions.
+    #:
+    #: :parameter dict config: project configuration.
+    INTERACTIVE_CONFIGURATION: Action[[Config]] = Action()
```
regisb, line 130: Can you please move this action such that it's listed in alphabetical order? Also, we should rename it to `CONFIG_INTERACTIVE`.

regisb (changes requested):
Who doesn't love simplicity? Thanks @Alec4r! Can you please just add a changelog entry?

regisb:
Sorry about the delay guys, I was busy with the Quince release.

tutor/hooks/catalog.py
Alec4r: @regisb Done

Alec4r:
> Who doesn't love simplicity? Thanks @Alec4r! Can you please just add a changelog entry?

I think is done, let me know if you have another comment.

regisb (approved):

Issue #936, opened 2023-11-14
The branching model of "nightly" must be improved


regisb:
The current branching model makes it very difficult to create releases branches on top of nightly.

When we work on a new release, we need the latest changes from the "nightly" branch. So we create the release branches on top of "nightly". But the "nightly" branches contain code that is specific to nightly mode. Namely:

- `__version_suffix__ = "nightly"`
- `OPENEDX_COMMON_VERSION: "master"`

So we need to override these values when we create the release branches. And we need to put them back in when we merge back the release branches in "nightly", after the release.

This process is cumbersome for plugin maintainers. To resolve this issue, we would have to automatically figure out that we are running "nightly" on installation. For instance, we could run:

    git clone --branch=nightly https://github.com/overhangio/tutor.git
    cd tutor/
    pip install -e .[nightly]

And then somehow detect the "nightly" requirements, and set `__version_suffix__` and `OPENEDX_COMMON_VERSION` accordingly.

Problems:

1. I don't know how to reliably detect that we installed the nightly requirements.
2. This means that we expect the user to type "nightly" twice, which is not very convenient.

Further thinking is required :monocle_face: ...

regisb:
Marking this as resolved, for now. When making a new release, the only change we now need to resolve is the version suffix in tutor core. All other conflicts should now have disappeared.

PR #944, opened 2023-11-20
feat: simplify nightly management

regisb:
Bumping the `OPENEDX_COMMON_VERSION` in the master branch usually creates a conflict when we merge the change in the nightly branch. To avoid this conflict, we add some logic to the `OPENEDX_COMMON_VERSION`.

This change should be invisible for most users.

This partially addresses issue #936.


regisb:
FYI @wahajArbisoft @kdmccormick

PR #957, opened 2023-12-08
fix: save config and env on `plugins enable/disable`

regisb:
This replaces https://github.com/overhangio/tutor/pull/917/

### DIFF
--- a/changelog.d/20231011_000434_codewithemad_save_configs.md
+++ b/changelog.d/20231011_000434_codewithemad_save_configs.md
@@ -1,0 +1,1 @@
+- [Improvement] You don't have to run `tutor config save` every time you enable or disable a plugin anymore. (by @CodeWithEmad)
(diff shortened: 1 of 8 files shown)


regisb:
@fghaas Would it make sense to add a `--no-save-env` (or similar) option to `tutor plugins enable/disable`?

fghaas:
Good question. _Most of the time,_ users would want automatic `config save`, particularly if the plugin requires a `config.yml` variable to be initialized. So I think Tutor does the "right" thing by default in v17, except if we're applying an excessively strict interpretation of [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Would a toggle to turn it off be useful, or would it be more likely to be a footgun? I really don't know.

regisb:
Let's do nothing for now. If you (or anyone else) later think it's important we can decide to add a toggle.

fghaas:
That works for me!

PR #958, opened 2023-12-12
fix: wrap django 5 imports in try-except block

mariajgrimaldi:
### Description
This PR wraps Django5 warning imports in try-except blocks to avoid failures in django3 that's still in use in edx-platform's master branch. For more info on the report see: https://openedx.slack.com/archives/CGE253B7V/p1702402901252609

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -150,9 +150,12 @@ LOGGING["loggers"]["blockstore.apps.bundles.storage"] = {"handlers": ["console"]
 # These warnings are visible in simple commands and init tasks
 import warnings
 
-from django.utils.deprecation import RemovedInDjango50Warning, RemovedInDjango51Warning
-warnings.filterwarnings("ignore", category=RemovedInDjango50Warning)
-warnings.filterwarnings("ignore", category=RemovedInDjango51Warning)
+try:
+    from django.utils.deprecation import RemovedInDjango50Warning, RemovedInDjango51Warning
+    warnings.filterwarnings("ignore", category=RemovedInDjango50Warning)
+    warnings.filterwarnings("ignore", category=RemovedInDjango51Warning)
+except ImportError:
+    pass
 
 warnings.filterwarnings("ignore", category=DeprecationWarning, module="wiki.plugins.links.wiki_plugin")
 warnings.filterwarnings("ignore", category=DeprecationWarning, module="boto.plugin")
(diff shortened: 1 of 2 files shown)


tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -150,9 +150,12 @@
 # These warnings are visible in simple commands and init tasks
 import warnings
 
-from django.utils.deprecation import RemovedInDjango50Warning, RemovedInDjango51Warning
-warnings.filterwarnings("ignore", category=RemovedInDjango50Warning)
-warnings.filterwarnings("ignore", category=RemovedInDjango51Warning)
+try:
+    from django.utils.deprecation import RemovedInDjango50Warning, RemovedInDjango51Warning
```
regisb, line 154: Could you please just add a comment that includes the "REMOVE-AFTER-V18" string? and explain why this try/except was still needed in v17.

regisb (changes requested):
Thanks for the very quick PR! This doesn't require a new release because it fixes the nightly branch (which will automatically get the fix). I'll merge as soon as you add the extra comment.

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -150,9 +150,12 @@
...
+    warnings.filterwarnings("ignore", category=RemovedInDjango50Warning)
+    warnings.filterwarnings("ignore", category=RemovedInDjango51Warning)
+except ImportError:
+    pass
```
kdmccormick, line 158:
```suggestion
except ImportError:
    # REMOVE-IN-V18:
    # In Quince, edx-platform uses Django 5. But on master, edx-platform still uses Django 3.
    # So, Tutor v17 needs to silence these warnings, whereas Tutor v17-nightly fails to import them.
    # Once edx-platform master is upgraded to Django 5, the try-except wrapper can be removed.
    pass
```

friendly edit -- just adding a REMOVE-IN comment

kdmccormick:
Merged via https://github.com/overhangio/tutor/commit/aa01e3095d9c7da532a8714e63d7e081dffedbdb

Thanks for the fix @mariajgrimaldi !

kdmccormick:
@regisb Oof, I just realized I merged this to nightly instead of master 🤦🏻 

I guess that fine, though, since we don't need this change on master? Let me know if you'd rather I make the same change on master too, for consistency's sake.

regisb:
That's fine, there is no need to backport if the feature is not strictly necessary. But next time, please rebase instead of adding a merge commit.

kdmccormick:
Will do.

PR #973, opened 2024-01-02
chore: install compiled base dependencies

CodeWithEmad:
If there's no weird reason behind it, this will fix the installation of the compiled base dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
     packages=find_packages(exclude=["tests*"]),
     include_package_data=True,
     python_requires=">=3.8",
-    install_requires=load_requirements("base.in"),
+    install_requires=load_requirements("base.txt"),
     extras_require={
         "dev": load_requirements("dev.txt"),
         "full": load_requirements("plugins.txt"),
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -57,7 +57,7 @@ def is_requirement(line: str) -> bool:
     packages=find_packages(exclude=["tests*"]),
     include_package_data=True,
     python_requires=">=3.8",
-    install_requires=load_requirements("base.in"),
+    install_requires=load_requirements("base.txt"),
```
regisb, line 60: It is generally not recommended to pin requirements in Python projects. This is because when a user runs `pip install tutor someotherreq`, installation will fail if the requirements are not exactly the same. I can't find the Python docs reference page anymore, though.

regisb:
As explained in my comment above, I strongly suggest we do not make this change. What issue is this PR supposed to resolve?

CodeWithEmad:
There were no issues addressed by this PR. I thought to myself that if a package in the `base.in` had a breaking change in the future, we wouldn't want to install the latest version.

PR #976, opened 2024-01-10
feat!: default importdemocourse to the new course & lib

kdmccormick:
## Description (will be copied into commit)

[We're overhauling the Demo Course repo!](https://github.com/openedx/openedx-demo-course/pull/47). That PR will restructure the repo, rebuild the Demo Course from scratch, and a Demo Content Library to accompany the course.

This PR, which should merge right after that one, modifies the existing `tutor (local|dev|k8s) do importdemocourse` command accordingly:
* The `--repo-dir` option now defaults to a new subdirectory: demo-course/course.
* The `--library-tar` option is added, defaulting to the library archive: dist/demo-content-library.tar.gz
* Content libraries require an owner user in order to be imported, so `--library-owner` is added as an option. Omitting the option means that the course will import fine, but the library import will be skipped and a warning will be printed. This is OK, as the course works fine without the library.

💥 BREAKING CHANGE: The new course is significantly different than the old one. Importing the old demo course remains possible, but with a couple extra arguments:  `tutor local do importdemocourse --repo-dir . --version open-release/palm.4`.

## Testing

Run:

```bash
tutor local do importdemocourse --version kdmccormick/new-demo-course --library-owner <any admin user>
```

Fire up Tutor and go to Studio. On your course listing, you should see the Demo Course, and under your library listing, you should see "Respiratory System Question Bank 1" (that's the demo library). Go to LMS, "Discover New", and choose Axim/DemoX (the one on the right).

image

Enroll. The outline should look like this:

image


tutor/commands/jobs.py
```diff
@@ -120,37 +120,54 @@ def create_user_template(
...
-    template = f"""
-# Import demo course
-git clone {repo} --branch {version} --depth 1 /tmp/course
-python ./manage.py cms import ../data /tmp/course/{repo_dir}
-
-# Re-index courses
-./manage.py cms reindex_course --all --setup"""
+    template = f"git clone {repo} --branch {version} --depth 1 /tmp/course"
+    if library_owner:
+        template += f"\nyes | ./manage.py cms import_content_library /tmp/course/{library_tar} {library_owner}"
```
regisb, line 166: Note to self: if we go with this approach, we should add a `-y/--yes` option to the `import_content_library` management command. We can take care of it after this PR is merged.

tutor/commands/jobs.py
```diff
@@ -120,37 +120,54 @@ def create_user_template(
...
+    else:
+        template += "\necho 'WARNING: Skipped demo library import because --library-owner was not provided.'"
```
regisb, line 168: This should be: `fmt.echo_alert("Skipped demo...")`.

regisb:
Very much looking forward to this change and seeing the old demo course go away!

tutor/commands/jobs.py
kdmccormick: agreed

tutor/commands/jobs.py
kdmccormick: sounds good, but this will be obviated by https://github.com/overhangio/tutor/pull/976#discussion_r1457952785

tutor/commands/jobs.py
```diff
@@ -120,37 +120,54 @@ def create_user_template(
...
+@click.option(
+    "-l",
+    "--library-tar",
+    default="dist/demo-content-library.tar.gz",
+    show_default=True,
+    help="Git relative .tar.gz path to import content library from",
+)
+@click.option(
+    "-L",
+    "--library-owner",
```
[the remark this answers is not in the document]
kdmccormick: @regisb I applied this in the latest commit--ready for another look!

changelog.d/20240110_101228_kyle_importnewdemocourse.md
```diff
@@ -0,0 +1,2 @@
+- 💥 [Feature] `tutor local do importdemocourse` will now import the new [Open edX Demo Course](https://github.com/openedx/openedx-demo-course), which has been rebuilt from scratch. If you wish to import the old demo course instead (which no longer receives updates), you can run: `tutor local do importdemocourse --repo-dir . --version open-release/palm.4`.
```
regisb, line 1: We should replace "palm.4" by "quince.1".

tutor/commands/jobs.py
```diff
@@ -124,14 +124,14 @@ def create_user_template(
...
-    default="https://github.com/openedx/edx-demo-course",
+    default="https://github.com/openedx/openedx-demo-course",
     show_default=True,
     help="Git repository that contains the course to be imported",
 )
 @click.option(
     "-d",
     "--repo-dir",
-    default="",
+    default="demo-course/course",
```
regisb, line 134: This change will cause some existing commands to fail, when users try to import a different custom course that is not stored in a subdirectory:

    tutor local do importdemocourse --repo https://github.com/me/mycourse

I wonder if we could simplify the CLI by auto-detecting the location of the "course.xml" file? This would make the import script much more complex. But I see no other option if we want to preserve ease-of-use and compatibility with the new demo course structure.

DawoudSheraz: Or maybe, for now, if the repo path ends with openedx-demo-course, treat the default as demo-course/course, otherwise ""?

changelog.d/20240110_101228_kyle_importnewdemocourse.md
sarina: Does this mean that you need to be on `quince.2` to get the new version of the course, or will it be possible to be on an older version of the platform but still import the new course?

kdmccormick: > We should replace "palm.4" by "quince.1".

Right, or even `quince.master`.

> Does this mean that you need to be on quince.2 to get the new version of the course, or will it be possible to be on an older version of the platform but still import the new course?

Older versions of Open edX and Tutor will *default* to the old demo course, but it is _possible_ to import the new demo course into them with the right arguments:

```
tutor local do importdemocourse --repo-dir demo-course/course --version master
```

I'll update this message to be clearer about this.

tutor/commands/jobs.py
kdmccormick: @regisb I am OK with that complexity. I'll implement this logic:

* If the `--repo-dir` argument is provided, use it.
* Else:
    * `find` the course.xml file
    * If there is exactly one course.xml, then use that directory.
    * Else if there are 0 or 2+ course.xml files, raise an exception.

@DawoudSheraz The old and new demo course are both stored in the same repo, "openedx-demo-course". "edx-demo-course" has redirected to "openedx-demo-course" a couple years.

kdmccormick:
Closed in favor of https://github.com/overhangio/tutor/pull/998

PR #984, opened 2024-01-16
fix: allow to setup redis maxmemory

Ian2012:
### Description

In some of our production environments, we have experienced an incremental use of Redis capacity. After some research we found that this [code](https://github.com/openedx/edx-platform/blob/open-release/nutmeg.3/common/lib/xmodule/xmodule/modulestore/split_mongo/mongo_connection.py#L248) (with no major changes since the nutmeg release) was storing course structure data in Redis without TTL as it the expected cache layer was `memcached` and it's configured by default to use the LRU eviction policy.

The above code stored the data in the following key format: `course_structure.{{db}}{{version_id}}` so there was a lot of outdated/unused data (due to changes in the courses) stored in Redis indefinitely.

**Note**: Although the majority of the keys stored in Redis were associated with sessions, those already had a TTL configured and didn't represent the majority of the data. It was estimated (using a sample) that every key in the `course_structure` namespace has 0.3 MB of data (for small courses) and there was around 54k keys there.

To configure an eviction policy in Redis we need to configure two params:
- `maxmemory`: Setups the Redis memory usage limit.
- `maxmemory-policy`: Setups the eviction policy used by Redis when the `maxmemory` limit is reached.

See [Key eviction](https://redis.io/docs/reference/eviction/) for more information.

The problem was discussed [here](https://discuss.openedx.org/t/course-structure-cache-never-expire-in-juniper-3/4336) in the Juniper release, but the problem is the same.

This PR solves that issue by configuring the above parameters.

### DIFF
--- a/tutor/templates/apps/redis/redis.conf
+++ b/tutor/templates/apps/redis/redis.conf
@@ -39,3 +39,9 @@ auto-aof-rewrite-percentage 100
 auto-aof-rewrite-min-size 64mb
 aof-load-truncated yes
 aof-use-rdb-preamble yes
+
+################################ MEMORY - EVICTION ###############################
+
+# https://redis.io/docs/reference/eviction/
+maxmemory {{ REDIS_MAXMEMORY }}
+maxmemory-policy allkeys-lru


regisb:
This PR points to a serious issue: that the course structure cache can grow without bounds when many changes are made to the courses. The `TIMEOUT=7200` parameter of the course_structure_cache is bypassed by the `timeout=None` argument in edx-platform.

Am I getting this right?

While I recognize that this is a serious issue, and I thank you for suggesting a fix, your approach has some important drawbacks:

1. The max memory threshold will work for some users, but it will inevitably have to be tweaked for others, with a value that is very hard to guess.
2. Enabling eviction for all keys may have some unintended effects which may be critical.

I think that the "right" fix would be to not override the default timeout, as was suggested by @ormsbee [here](https://discuss.openedx.org/t/course-structure-cache-never-expire-in-juniper-3/4336/3?u=regis). But I'm open to other approaches.

ormsbee:
I didn't realize that Redis wouldn't evict something unless it had an explicit TTL set on it. I think it's fine to put a timeout on the structure cache entries, but it should be a long timeout (a week maybe?). I definitely wouldn't want to go to the default `7200` seconds.

regisb:
Thanks for commenting Dave. We can definitely bump the current timeout from 7200s to 1 week. Would you be open to a PR that removed the `None` timeout?

I can confirm that `course_structure*` keys have no expiration time, as opposed to others:

```
$ tutor local exec redis bash
$ redis-cli -n 1
127.0.0.1:6379[1]> KEYS *
...
85) "default:1:celery-task-meta-f7cbb83b-f03f-4391-8ba1-cca0e5d314bd"
86) "course_structure:1:650d788f8e457ec4d9e18d64"
87) "course_structure:1:649d8994d3f9cafd41d4090f"
88) "course_structure:1:650d788f5998d7ab16e18d67"
89) "course_structure:1:64ff3dbf2918040af8062d4b"
90) "course_structure:1:649d89a5c6af476799d40911"
127.0.0.1:6379[1]> EXPIRETIME course_structure:1:649d89a5c6af476799d40911
(integer) -1
127.0.0.1:6379[1]> EXPIRETIME default:1:common.djangoapps.xblock_django.models.XBlockConfiguration.common.djangoapps.xblock_django.api.disabled_xblocks
(integer) 1705998245
```

ormsbee:
@regisb: I was thinking this over last night. If having any timestamp is sufficient to trigger Redis's LRU eviction policy, let's just set it for something that's absurdly high, like a year. I have on one occasion been saved by having structure cache entries > 1 month old. Tag me on that for a PR, and I'll merge it.

regisb:
> If having any timeout is sufficient to trigger Redis's LRU eviction policy

I might be wrong, but I don't think this is the case. In Redis, LRU eviction will only work if there is a defined maximum memory value. I would like to avoid defining such a value. If we do define such a value, then LRU eviction will kick in even for keys that do not have an expiration time. 

Thus, defining a very large timeout will not resolve our issue.

Ian2012:
@regisb You are right, we would need to define the `maxmemory` parameter in Redis for the LRU eviction to work.

ormsbee:
So if someone makes a PR that sets the timeout for structures to a week, I will merge it. But I really think that we do need to put some value for `maxmemory` and make Redis do LRU. We shouldn't be deploying our cache in a mode where it just explodes when it fills up.

Ian2012:
@regisb Would you agree on creating a patch instead for it? So people can define whatever parameter they need for their redis installation. Also, this fix can be more easily backported to other tutor versions instead of edx-platform versions.

What do you think?

regisb:
Sorry for my slow response...

> But I really think that we do need to put some value for maxmemory and make Redis do LRU. We shouldn't be deploying our cache in a mode where it just explodes when it fills up.

The flip side of that is that for smaller platforms, unused and useless cached entries would just accumulate until reaching the maxmemory, which would be a waste of resources.

> Would you agree on creating a patch instead for it?

Yes, absolutely. We need to make it possible to customize Redis settings.

Here's what I propose:

1. In edx-platform, remove the `None` timeout such that the cache key uses the default timeout.
2. In edx-platform, bump the default timeout of `course_structure_cache` from 7200 to 1 week.
3. In Tutor, also bump the default timeout of `course_structure_cache` from 7200 to 1 week.
4. In Tutor, make it possible to customize Redis settings by introducing a dedicated patch.

That way, the default would be to not have a maxmemory for Redis, but people would be able to set it based on their requirements using a simple patch.

ormsbee:
That plan works for me, and I think it puts us in a better place than we are now.

> > But I really think that we do need to put some value for maxmemory and make Redis do LRU. We shouldn't be deploying our cache in a mode where it just explodes when it fills up.
> 
> The flip side of that is that for smaller platforms, unused and useless cached entries would just accumulate until reaching the maxmemory, which would be a waste of resources.

Just to be clear, I mean to say that Redis should have a `maxmemory` limit set for operational safety reasons, regardless of what we decide the TTL for any particular cache entries are. We could set it to a conservative number for smaller installs, and let larger ones override it with higher values.

regisb:
> Just to be clear, I mean to say that Redis should have a maxmemory limit set for operational safety reasons, regardless of what we decide the TTL for any particular cache entries are. We could set it to a consertive number for smaller installs, and let larger ones override it with higher values.

Yes, I agree, but I'm really wondering what this conservative number should be. Maybe 8G?

According to my experiments, we can hardcode `maxmemory 8gb` in redis.conf, and operators will be able to override it using just a patch. This would allow us to avoid introducing a new configuration setting. @Ian2012 would that work for you?

I opened an upstream PR here: https://github.com/openedx/edx-platform/pull/34210 I'm going to cherry-pick the commit in Tutor.

Ian2012:
I agree on the patch but the default seems too high. 4gb seems like a more reasonable default for most installations.

regisb:
OK, let's go with 4GB.

regisb:
Closed by #1001.

Issue #975, opened 2024-01-10
Template rendering failure when installing theme plugin


ravikhetani:
**Bug description**
Hi, I was trying to build the Openedx image with a theme plugin, and an exception was thrown in `/tutor/tutor/env.py` in the  `def render_template` method, for fonts in my themes `lms/static/fonts` directory. My fonts are in the OpenType format (.otf). I can't see anything documented in the openedx documents, or tutor documents that suggests that OpenType fonts can't be used.

Having looked at env.py https://github.com/overhangio/tutor/blob/fdd230f353026e949305978a48065fdf17de44df/tutor/env.py#L18, I can see that it's not included in the binary file extensions, preventing `render_template` from trying to render the fonts. Any reason why it's not included in the list of extensions?

**How to reproduce**
Add Opentype fonts to custom theme plugin, and try to build the openedx image

regisb:
Hi @ravikhetani! It should be easy enough to add the ".otf" extension to the `BIN_FILE_EXTENSIONS`. Alternatively, we could create a new filter were we would include all files that should be copied, and not rendered. This filter would include all the files with an extension in `BIN_FILE_EXTENSIONS`. 

Would you like to open a PR? Ideally, we should implement both solutions, but feel free to tackle just one. If you'd rather not open a PR, we can take care of it.

ravikhetani:
@regisb I think yes to both - I'll submit a PR for the quick fix of adding it to the lest of extensions, but I think the filter is the proper solution, as I can imagine other supported image formats etc that people may want to add.

regisb:
Closed by #985.

Abdul-Muqadim-Arbisoft:
@regisb I will work on making the file copying filter

Abdul-Muqadim-Arbisoft:
This pull request introduces a new filter, IS_FILE_RENDERED, designed to enhance the flexibility of file handling within Tutor plugins. The new IS_FILE_RENDERED filter allows plugin developers to specify files that should not be rendered but instead copied . For our specific scenario we can add path to the respective .otf files through the filter we can return  false for  rendering it and it would only be copied

PR #985, opened 2024-01-18
chore: Add .webp and .otf file extensions to list of binary file extensions

ravikhetani:
Short term solution to address issue raised in #975, which was causing tutor to throw exceptions when enabling theming plugins, which included file extensions not specified by `BIN_FILE_EXTENSIONS` in `env.py`


regisb (changes requested):
Thanks for the PR! Please rebase your changes on top of master and run `make format` to properly format your code.

regisb (approved):
Thanks for the PR!

Issue #989, opened 2024-01-22
`tutor plugins enable xxx` does not update the environment


regisb:
**Bug description**

Enabling a plugin should properly update the environment.

**How to reproduce**

When running:

    tutor plugins enable cairn

I would expect that the environment is updated, as per [this new feature](https://github.com/overhangio/tutor/pull/957). But it turns out it's not the case. The following command does not output anything:

    grep cairn ~/.local/share/tutor/env/local/docker-compose.yml

**Additional context**

This is a caching issue. The `ENV_PATCHES_DICT` variable is a cache that is filled once all plugins are loaded -- but not when a new plugin is loaded. This is an issue that was discussed before here: https://github.com/overhangio/tutor/pull/917#discussion_r1409064275

Turns out, cache management is even more difficult that we anticipated. This issue also shows up in the MFE plugin, which has to manually implement, cache management, which is ugly: https://github.com/overhangio/tutor-mfe/blob/de526369f0a882c459ac9d20575f8c23ea0f7fab/tutormfe/plugin.py#L93

Still, we do want to preserve caching features for performance reasons. So here's what I suggest:

1. Create a `hooks.Actions.PLUGINS_CACHE_CLEAR` action.
2. Create a new `tutor/hooks/utils.py` module, with Apache 2.0 license.
3. In that new module, create a `@plugins_cache` decorator that will be a wrapper on top of `functools.lru_cache`.
4. Every time a function is decorated with `@plugins_cache`, a callback to the PLUGINS_CACHE_CLEAR must be added, such that the cache is cleared when that action is called.
5. Import `hooks.utils.plugins_cache` in `hooks/__init__.py` such that the function is available in plugins.
6. Make sure that the `plugins_cache` decorator is documented in https://docs.tutor.edly.io/reference/api/hooks/index.html
7. Migrate `ENV_PATCHES_DICT` to a `_env_patches_dict` function that is decorated with `hooks.plugins_cache`.

PR #996, opened 2024-01-30
Fix env save on plugins

regisb:
Environment was not updated correctly on `tutor plugins enable ...`
because of a caching issue. To bypass this issue, we improve the caching
mechanism and introduce a new `lru_cache` decorator.

Close https://github.com/overhangio/tutor/issues/989.


kdmccormick (approved):
An elegant solution to a gnarly problem, nice work.  I am also glad you went with `PLUGIN_LOADED` rather than `PLUGIN_CACHE_CLEAR`--the former sounds like a reusable action hook; the latter sounds more like an indirect function call.

I think you buried the lede on what was wrong before, though. Was it that you were hooking into `PLUGINS_LOADED` instead of `PLUGIN_LOADED`?

DawoudSheraz (approved):
A single difference of 'S' in `PLUGIN_LOADED` and `PLUGINS_LOADED` causing such a pain -- nice one on this fix.

regisb:
> I think you buried the lede on what was wrong before, though. Was it that you were hooking into PLUGINS_LOADED instead of PLUGIN_LOADED?

Not quite! I knew that I could use PLUGIN_LOADED before. But that would have meant calling the `_fill_patches_cache` function for every loaded plugin. In this new implementation, we do clear the cache many times needlessly, but that's OK, because we only recompute the cache when needed.

PR #1001, opened 2024-02-08
fix: infinite cache growth

regisb:
See the discussion here: https://github.com/overhangio/tutor/pull/984
And the upstream PR here: https://github.com/openedx/edx-platform/pull/34210

The tl;dr is that the Redis course structure cache was growing without bounds. While the upstream fix should resolve that issue, we decided that Tutor should have a maxmemory limit and an eviction policy set for operational safety.

    tutor local run cms ./manage.py cms shell -c "from django.core.cache import caches; c = caches['course_structure_cache']; [c.expire(key, 604800) for key in c.keys('*')]"

FYI @Ian2012 @ormsbee

### DIFF
--- a/changelog.d/20240208_111236_regis_infinite_course_cache.md
+++ b/changelog.d/20240208_111236_regis_infinite_course_cache.md
@@ -1,0 +1,4 @@
+- 💥[Bugfix] Prevent infinite growth of course structure cache in Redis. (by @regisb)
+  - Redis is now configured with a maximum memory size of 4GB. If this is too low for your platform, you should increase this value using the new "redis-conf" patch.
+  - Make sure that course structure cache keys have an actual timeout.
+- [Feature] Introduce the "redis-conf" patch. (by @regisb)
--- a/docs/reference/patches.rst
+++ b/docs/reference/patches.rst
@@ -376,6 +376,13 @@ File: ``apps/openedx/settings/lms/production.py``
 
 Python-formatted LMS settings in production. Values defined here override the values from :patch:`openedx-lms-common-settings`.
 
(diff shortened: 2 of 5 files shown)


regisb:
Let's wait until the [upstream PR](https://github.com/openedx/edx-platform/pull/34210) is merged, then update the commit sha1 in the dockerfile.

VirtualDiegox:
@regisb just to be sure, this fix should be set also in the k8s templates? specifically in volumes.yml? I faced the same issue and i had to resize the vpc manually before running the command to expire the existing keys

regisb:
@VirtualDiegox As far as I know, this fix does not require any change made to volumes.yml. Well, we could increase the volume size to 4GB, but that would not be required by this change. Did you have something else in mind?

regisb:
@VirtualDiegox I'm wondering if you are facing the issue that we actually solved here. I.e: the infinite growth of the course structure cache. How much space is your Redis instance using after you clear the course structure cache?

VirtualDiegox:
@regisb Yes I think i was facing that issue too, but our environment it's based on the k8s deployment, that's why I asked if the volume size in the k8s template should be increased too, right now our Redis volume is 4gb. The command to clear the course structure cache doesn't work if the space of the volume is full, so I solved it temporarily with a plugin that addionally to the changes that you made, create a new volume with 4gb instead of 1gb and override the redis deployment in order to use it. If you are wondering why create new volume instead of modifying the existing one it is due permissions issues in the cluster.

regisb:
Right, I understand. But after you clear the course structure cache, how much memory is your Redis volume using, out of the 4GB maximum space?

PR #1005, opened 2024-02-20
Change command to reset build context

misilot:
When I tried to run the specified command, docker told me to run `docker context use default`

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -194,7 +194,7 @@ Now build again::
 
     tutor images build
 
-All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
+All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker contextuse default``. 
 
 .. note::	
 	Setting a too low value for maximum parallelism will result in longer build times.


regisb:
What's the full error message you got?

misilot:
> ERROR: run `docker context use default` to switch to default context

regisb:
What's you Docker version?

regisb:
That's strange. The `docker buildx use` command is not deprecated in Docker 25.0.3, as far as I know: https://github.com/docker/buildx/blob/master/docs/reference/buildx_use.md
https://docs.docker.com/engine/release-notes/25.0/
https://docs.docker.com/build/release-notes/

I can't reproduce the issue on my machine. I'm running Docker 25.0.3 as well, with `docker buildx version==v0.12.1`. I'm wondering if you have a "default" builder at all? You can check with:

    docker buildx ls

DawoudSheraz:
While there are some issues with 'default' instance on my system, I am also seeing the mentioned error. I am also using Docker 25.0.3, on Mac. 

```
> docker buildx ls           
NAME/NODE       DRIVER/ENDPOINT  STATUS   BUILDKIT PLATFORMS
dawoudsheraz    docker-container                   
  dawoudsheraz0 desktop-linux    inactive          
default *                        error             
desktop-linux   docker                             
  desktop-linux desktop-linux    running  v0.12.5  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6

Cannot load builder default *: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

> docker buildx use desktop-linux
ERROR: run `docker context use desktop-linux` to switch to context desktop-linux

> docker context use desktop-linux
desktop-linux
Current context is now "desktop-linux"

```

regisb:
@misilot do you run macOS?

misilot:
@regisb, yes I am running macOS

hamza-56:
The command `docker buildx use` still exists. However, to revert to the default context, we need to use the `context` command. When switching to a non-default context, we can utilize `buildx`.

https://github.com/docker/buildx/pull/39/files

DawoudSheraz:
> The command `docker buildx use` still exists. However, to revert to the default context, we need to use the `context` command. When switching to a non-default context, we can utilize `buildx`.
> 
> https://github.com/docker/buildx/pull/39/files

Right. I think we can make the docs explicit that we need to use `docker context use default` specifically to switch back to default but can use buildx to switch to non-default contexts.

regisb:
Builders and contexts are different things, so don't think that we should be recommending to switch context by default. 

I just saw this part of the [docker buildx use](https://docs.docker.com/reference/cli/docker/buildx/use/) documentation:

> Alternatively, a context name can be used to switch to the default builder of that context.

If I understand correctly, `docker buildx` will first check if the argument corresponds to a different context name. Thus, I am able to reproduce the issue by switching to a different context, first:

```
$ docker context create pouac --docker from=default
pouac
Successfully created context "pouac"
{09:25:40:caffeine}~$ docker context ls
NAME        DESCRIPTION                               DOCKER ENDPOINT               ERROR
default *   Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   
pouac                                                 unix:///var/run/docker.sock   
$ docker context use pouac
pouac
Current context is now "pouac"
$ docker buildx use max4cpu
$ docker buildx use default
ERROR: run `docker context use default` to switch to default context
```

@misilot what is the output of `docker context ls` for you?

misilot:
```
➜  ~ docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                 KUBERNETES ENDPOINT   ORCHESTRATOR
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux *     moby                Docker Desktop                            unix:///Users/tmisilo/.docker/run/docker.sock
```

Is what I have currently on my laptop (macos)

regisb:
This confirms my assumption. Docker recommends that you run `docker context use` instead of `docker buildx use` when you have a custom context that you are currently using. As a consequence, I don't think that `docker context use` should be the general recommendation.

misilot:
Sounds good. I'll go with Docker is just creating custom contexts on it's own since I believe those are the default from an install :)

DawoudSheraz:
Based on the conversations above, it seems no further action is required. Closing this out. Feel free to re-open/open a new one if some changes in docs are still needed.

PR #1007, opened 2024-02-22
fix: merge conflicts due to edx-platform patches in nightly

regisb:
When building the nightly images, some patches fail because they come from the master branch. To address this, we apply certain patches only if we are not in nightly mode.


regisb:
Merging this now to resolve CI build.

Issue #1006, opened 2024-02-21
Tutor Python 3.11 upgrade


DawoudSheraz:
- [x] Update Python to 3.11 in Dockerfile for nightly and ensure the image builds as expected. The changes would be merged to nightly. This task would rely on upgrade of edx-platform to Python 3.11
- [x] Ensure tutor codebase is compatible with Python 3.12

See https://github.com/overhangio/tutor/issues/1004 for the context.

DawoudSheraz:
Resolves by https://github.com/overhangio/tutor/pull/1008

DawoudSheraz:
Attempted building Open edX docker file on Python 3.12. I had to make the following changes /tutor/templates/build/openedx/Dockerfile

L27-L30 (Python and Pyenv)
```
ARG PYTHON_VERSION=3.12.2
ENV PYENV_ROOT /opt/pyenv
RUN git clone https://github.com/pyenv/pyenv $PYENV_ROOT --branch v2.3.36 --depth 1
```

L96 - L102 (setuptools and pip update)
```
RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
    pip install \
    # https://pypi.org/project/setuptools/
    # https://pypi.org/project/pip/
    # https://pypi.org/project/wheel/
    setuptools==69.1.1 pip==24.0 wheel==0.41.2
```

However, the build failed when installing Open edX base requirements with the error:

```
 Downloading numpy-1.22.4.zip (11.5 MB)
#40 115.6      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 11.5/11.5 MB 1.6 MB/s eta 0:00:00
#40 116.3   Installing build dependencies: started
#40 124.2   Installing build dependencies: finished with status 'done'
#40 124.2   Getting requirements to build wheel: started
#40 124.4   Getting requirements to build wheel: finished with status 'error'
#40 124.4   error: subprocess-exited-with-error
#40 124.4   
#40 124.4   × Getting requirements to build wheel did not run successfully.
#40 124.4   │ exit code: 1
#40 124.4   ╰─> [33 lines of output]
#40 124.4       Traceback (most recent call last):
#40 124.4         File "/openedx/venv/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
#40 124.4           main()
[...]

PR #1008, opened 2024-02-23
build: add python 3.12 checks alongside 3.8

DawoudSheraz:
For #1006 

### Description

- Adds Python 3.12 checks in CI alongside 3.8
- Fix `make docs` in Python 3.12

`make docs` in Python 3.12 was failing for various reasons for variety of items (Action, Filter, FilterCallbackFunc). Mainly, the generic `type` does not get translated to `py:class` and is instead picked up as `py:data`. There were a few issues on Sphinx and some related items on Stackoverflow. Everyone handled it differently in some capacity. One suggestion was to ignore some of the type params but it did not result in docs building as expected (https://github.com/overhangio/tutor/pull/1008#issuecomment-1968906467). The appropriate fix for "Action & Filter" was to add entries in autodoc_type_aliases. FilterCallbackFunc was needed to be ignore because it does not translate to a py:class.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,6 +51,8 @@ nitpick_ignore = [
     # python 3.10
     ("py:class", "NoneType"),
     ("py:class", "click.core.Command"),
+    # Python 3.12
+    ("py:class", "FilterCallbackFunc"),
 ]
 # Resolve type aliases here
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases
@@ -58,6 +60,15 @@ autodoc_type_aliases: dict[str, str] = {
     # python 3.10
     "T": "tutor.core.hooks.actions.T",
     "T2": "tutor.core.hooks.filters.T2",
+    # # python 3.12
+    "L": "tutor.core.hooks.filters.L",
+    "FilterCallbackFunc": "tutor.core.hooks.filters.FilterCallbackFunc",
+    # https://stackoverflow.com/questions/73223417/type-aliases-in-type-hints-are-not-preserved
+    # https://github.com/sphinx-doc/sphinx/issues/10455
+    # https://github.com/sphinx-doc/sphinx/issues/10785
+    # https://github.com/emdgroup/baybe/pull/67
+    "Action": "tutor.core.hooks.actions.Action",
+    "Filter": "tutor.core.hooks.filters.Filter",
 }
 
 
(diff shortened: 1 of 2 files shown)


DawoudSheraz:
The docs build issue mainly happens because Sphinx is not able to correctly identify `type` params. It tends to translate each param variation into a separate `py:class` entity, thus leading into errors when running in nitpick mode. One of the ways was to ignore the patterns via `nitpick_ignore_regex` but it resulted in docs not building as expected. 

- https://docs.plasmapy.org/en/stable/contributing/doc_guide.html#reference-target-not-found
- https://github.com/sphinx-doc/sphinx/issues/10974
- https://github.com/sphinx-doc/sphinx/issues/10785 

image

changelog.d/20240229_151232_dawoud.sheraz_py312.md
```diff
@@ -0,0 +1 @@
+- [Feature] Add Python 3.12 CI and `make docs` compatibility (by @DawoudSheraz)
```
regisb, line 1: CI-only changes don't warrant a changelog. 

Basically, to decide whether a change needs a changelog entry, we need to place ourselves in the position of the typical Tutor user: platform administrator or plugin developer. Whenever they upgrade, we expect that they have a look at the changelog entries. We don't want them to worry unnecessarily about changes that won't impact them. But we do want to highlight the features that they might want to use and the breaking changes.

docs/conf.py
```diff
@@ -51,13 +51,24 @@
...
+    # Python 3.12
+    ("py:class", "FilterCallbackFunc"),
 ]
 # Resolve type aliases here
 # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases
 autodoc_type_aliases: dict[str, str] = {
     # python 3.10
     "T": "tutor.core.hooks.actions.T",
     "T2": "tutor.core.hooks.filters.T2",
+    # # python 3.12
```
regisb, line 63: See? this is why I hate Sphinx...

regisb (changes requested):
Remove changelog entry and we are good to go.

changelog.d/20240229_151232_dawoud.sheraz_py312.md
DawoudSheraz: Ok.

regisb (approved):
lgtm!

PR #1009, opened 2024-02-27
docs: add git clone eof error troubleshooting

Danyal-Faheem:
**Issue**

- Git throws an EOF error while cloning large repositories on slow/unstable networks.

```
RUN mkdir -p /openedx/edx-platform &&     git clone https://github.com/openedx/edx-platform.git --branch open-release/quince.1 --depth 1 /openedx/edx-platform:
Cloning into '/openedx/edx-platform'...
fatal: the remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
```

- This error is thrown while cloning `edx-platform` during the build process of the `openedx`/`openedx-dev` image.
https://github.com/overhangio/tutor/blob/13c420cb0a722b90bf7bb7984a3ee44eb150114d/tutor/templates/build/openedx/Dockerfile#L41-L43

**Changes**

- Added a troubleshooting guide for the issue thrown by git while cloning large repositories on slow/unstable networks.
- Since this issue is subjective to every users network connection, I have instead just added a troubleshooting guide to reduce resources in the buildkit which does fix the issue.

This error has been reported by multiple users, for example [here](https://discuss.openedx.org/t/the-remote-end-hung-up-unexpectedly-early-eof-index-pack-failed/12094) and [here](https://discuss.openedx.org/t/problem-when-tutor-images-build-openedx/12237). I myself have faced this as well.
The max I could try it out was on a 25 Mb/s wired connection and it still threw the error.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -169,6 +169,8 @@ This issue should only happen in development mode. Long story short, it can be s
 
 If you'd like to learn more, please take a look at `this Github issue <https://github.com/overhangio/tutor/issues/302>`__.
 
+.. _high_resource_comsumption:
+
 High resource consumption on ``tutor images build`` by docker 
 -------------------------------------------------------------
 
@@ -194,15 +196,14 @@ Now build again::
 
 All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
 
-fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed when running `tutor images build ...`
-------------------------------------------------------------------------------------------------------------------------------
-
-This issue can occur due to problems with your network while cloning edx-platform which is a fairly large repository.
-
-First, try to run the same command once again to see if it works as your network connection can sometimes drop during the build process.
-
-If that does not work, you can follow the tutorial above for `High resource consumption <#high-resource-consumption-on-tutor-images-build-by-docker>`_ to limit the number of concurrent build steps so that your network is not being shared between multiple layers at once.
-
 .. note::	
-	The build for the image will be considerably slow now.
+	Setting a too low value for maximum parallelism will result in longer build times.
 
+fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed when running ``tutor images build ...``
+--------------------------------------------------------------------------------------------------------------------------------
+
+This issue can occur due to problems with the network connection while cloning edx-platform which is a fairly large repository.
+
+First, try to run the same command once again to see if it works as the network connection can sometimes drop during the build process.
+
+If that does not work, follow the tutorial above for :ref:`High resource consumption <high_resource_consumption>` to limit the number of concurrent build steps so that the network connection is not being shared between multiple layers at once.
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -169,7 +169,7 @@ This issue should only happen in development mode. Long story short, it can be s
 
 If you'd like to learn more, please take a look at `this Github issue <https://github.com/overhangio/tutor/issues/302>`__.
 
-.. _high_resource_comsumption:
+.. _high_resource_consumption:
 
 High resource consumption on ``tutor images build`` by docker 
 -------------------------------------------------------------
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -193,3 +193,16 @@ Now build again::
     tutor images build
 
 All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
+
+fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed
+----------------------------------------------------------------------------------------
+
+This issue can occur due to problems with your network while cloning edx-platform which is a fairly large repository.
+
+You should first of all try to rerun the command you were trying to run to see if it works.
+
+If that does not work, you can follow the tutorial above for `High resource consumption <#high-resource-consumption-on-tutor-images-build-by-docker>`_ and set `max-parallelism = 1` so that while edx-platform is being cloned, the network is not being utilized by another command.
+
+.. note::	
+	The build for the image will be considerably slow now as each step will be built one by one.
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.rst
```diff
@@ -193,3 +193,12 @@ Now build again::
...
 All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
+
+fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed
+----------------------------------------------------------------------------------------
+
+This issue can occur due to problems with your network while cloning edx-platform which is a fairly large repository.
+
+You should first of all try to rerun the command you were trying to run to see if it works.
+
+If that does not work, you can follow the tutorial above for `High resource consumption <#high-resource-consumption-on-tutor-images-build-by-docker>`_ and set `max-parallelism = 1` so that while edx-platform is being cloned, the network is not being utilized by another command.
```
DawoudSheraz, line 204: Should a caution be added here that this can increase build time?

Danyal-Faheem: I've added a short note at the end detailing this.

docs/troubleshooting.rst
```diff
@@ -193,3 +193,15 @@ Now build again::
...
+
+.. note::	
+	The build time for the image will be considerably slow now as each step will be built one by one.
```
DawoudSheraz, line 207: nit: build time --> build

DawoudSheraz (approved):

docs/troubleshooting.rst
Danyal-Faheem: Changed, thank you!

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
     tutor images build
 
 All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
+
+fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed
+----------------------------------------------------------------------------------------
+
+This issue can occur due to problems with your network while cloning edx-platform which is a fairly large repository.
```
regisb, line 200: In what situation does this error occur? Unless I'm mistaken, this happens whenever we run `tutor images build ...` command, right? We should tell the reader that, such that they feel confident that the error that we are describing matches their experience.

Also, try to avoid the "you" pronoun whenever possible. For instance, some of the sentences can be rewritten as:

> This issue can occur due to network problems while cloning edx-platform which is a fairly large repository.

> First, try to run again the same command to see if it works.

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
...
+
+You should first of all try to rerun the command you were trying to run to see if it works.
```
regisb, line 202: This is good advice, but it does not inspire confidence. How many times should the user try to re-run the same command? Why should they do that?

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
...
+
+If that does not work, you can follow the tutorial above for `High resource consumption <#high-resource-consumption-on-tutor-images-build-by-docker>`_ and set `max-parallelism = 1` so that while edx-platform is being cloned, the network is not being utilized by another command.
```
regisb, line 204: Setting max-parallelism=1 is not great advice. I would suggest just forwarding the user to the above tutorial, and they can figure out by themselves what is the right paralellism value.

regisb (changes requested):

docs/troubleshooting.rst
Danyal-Faheem: I've updated the title with the sample command.

I've removed the "you" pronoun as well.

docs/troubleshooting.rst
Danyal-Faheem: I've added that the command should be run once again and why it should be run again.

docs/troubleshooting.rst
Danyal-Faheem: I've removed that part and just forwarded the user to the other tutorial.

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
     tutor images build
 
 All build commands should now make use of the newly configured builder. To later revert to the default builder, run ``docker buildx use default``. 
+
+fatal: the remote end hung up unexpectedly / fatal: early EOF / fatal: index-pack failed when running `tutor images build ...`
```
regisb, line 197: In RST, verbatim strings are surrounded by double ticks: ``` ``tutor images build`` ```

docs/troubleshooting.rst
regisb: You left out the "your network" :) (and below as well)

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
...
+------------------------------------------------------------------------------------------------------------------------------
+
+This issue can occur due to problems with your network while cloning edx-platform which is a fairly large repository.
+
+First, try to run the same command once again to see if it works as your network connection can sometimes drop during the build process.
+
+If that does not work, you can follow the tutorial above for `High resource consumption <#high-resource-consumption-on-tutor-images-build-by-docker>`_ to limit the number of concurrent build steps so that your network is not being shared between multiple layers at once.
```
regisb, line 204: Let's use a ref for internal linking. (see instances of `:ref:`)

docs/troubleshooting.rst
```diff
@@ -193,3 +193,16 @@ Now build again::
...
+
+.. note::	
+	The build for the image will be considerably slow now.
```
regisb, line 207: This note does not belong here, but to the "High resource consumption" section above. Also, it should be reworded as: "Setting a too low value for maximum parallelism will result in longer build times."

regisb (changes requested):

docs/troubleshooting.rst
Danyal-Faheem: Done. Thanks for letting me know!

docs/troubleshooting.rst
Danyal-Faheem: Replaced with `:ref:`

docs/troubleshooting.rst
Danyal-Faheem: Updated and moved the note in the previous section

regisb (approved):
:+1:

PR #1010, opened 2024-02-29
feat: Celery worker concurrency setting

arbrandes:
This allows the user to configure how many Celery workers are spawned independently of how many CPUs there are in the system. The default is to spawn as many workers as there are CPUs, which in some cases can consume too many resources.

The setting should be particularly useful to people running Tutor for development on Linux machines, where reducing the concurrency to "1" can reduce RAM usage significantly.

### Testing

1. Before running this branch, launch a Tutor environment and count how many celery process there are.  With something like:

    ````
    pgrep celery | wc -l
    ````

    You should get twice the number of CPUs on the system - one set for each of LMS and CMS - plus two parent processes.  (On my machine, which has 12 real cores + 12 virtual ones, the number comes out to 26.)

3. Stop the environment, install this branch, and set:
 
    ````
    tutor config save --set OPENEDX_LMS_CELERY_WORKERS=1 --set OPENEDX_CMS_CELERY_WORKERS=1
    ````

4. Relaunch the environment.  The worker containers should be recreated.

5. Check the number of celery processes.  There should now be just 4.

It's worth checking RAM usage, too.  Before, my dev environment would take up 8 or more gigabytes of RAM.  After, it takes less than 3.


tutor/templates/k8s/deployments.yml
```diff
@@ -141,7 +141,7 @@ spec:
       containers:
         - name: cms-worker
           image: {{ DOCKER_IMAGE_OPENEDX }}
-          args: ["celery", "--app=cms.celery", "worker", "--loglevel=info", "--hostname=edx.cms.core.default.%%h", "--max-tasks-per-child", "100", "--exclude-queues=edx.lms.core.default"]
+          args: ["celery", "--app=cms.celery", "worker", "--loglevel=info", "--hostname=edx.cms.core.default.%%h", "--concurrency={{ OPENEDX_CMS_CELERY_WORKERS }}", "--max-tasks-per-child", "100", "--exclude-queues=edx.lms.core.default"]
```
dkaliberda, line 144: The `--concurrency` argument specifies the number of worker processes. This is not an ideal practice in Kubernetes environments because:
1. Kubernetes prefers to manage scalability and replication at the container orchestration level, using the `replicas` field in a Deployment to manage the number of pod instances. 
2. Setting `--concurrency` inside a container limits the scalability to the process level inside the pod, rather than allowing Kubernetes to manage multiple pods across nodes for better fault tolerance and load distribution.
3. It violate "one process per container" principle. This is important because with multiple processes in the same container, it is harder to troubleshoot the container because logs from different processes will be mixed together, and it is harder to manage the processes lifecycle, etc.

tutor/templates/k8s/deployments.yml
```diff
@@ -250,7 +250,7 @@ spec:
       containers:
         - name: lms-worker
           image: {{ DOCKER_IMAGE_OPENEDX }}
-          args: ["celery", "--app=lms.celery", "worker", "--loglevel=info", "--hostname=edx.lms.core.default.%%h", "--max-tasks-per-child=100", "--exclude-queues=edx.cms.core.default"]
+          args: ["celery", "--app=lms.celery", "worker", "--loglevel=info", "--hostname=edx.lms.core.default.%%h", "--concurrency={{ OPENEDX_LMS_CELERY_WORKERS }}", "--max-tasks-per-child=100", "--exclude-queues=edx.cms.core.default"]
```
dkaliberda, line 253: The same problem

tutor/templates/local/docker-compose.yml
```diff
@@ -158,7 +158,7 @@ services:
     environment:
       SERVICE_VARIANT: lms
       DJANGO_SETTINGS_MODULE: lms.envs.tutor.production
-    command: celery --app=lms.celery worker --loglevel=info --hostname=edx.lms.core.default.%%h --max-tasks-per-child=100 --exclude-queues=edx.cms.core.default
+    command: celery --app=lms.celery worker --loglevel=info --hostname=edx.lms.core.default.%%h --concurrency={{ OPENEDX_LMS_CELERY_WORKERS }} --max-tasks-per-child=100 --exclude-queues=edx.cms.core.default
```
dkaliberda, line 161: The docker-compose also provides mechanisms for managing replicas.

tutor/templates/local/docker-compose.yml
```diff
@@ -177,7 +177,7 @@ services:
     environment:
       SERVICE_VARIANT: cms
       DJANGO_SETTINGS_MODULE: cms.envs.tutor.production
-    command: celery --app=cms.celery worker --loglevel=info --hostname=edx.cms.core.default.%%h --max-tasks-per-child 100 --exclude-queues=edx.lms.core.default
+    command: celery --app=cms.celery worker --loglevel=info --hostname=edx.cms.core.default.%%h --concurrency={{ OPENEDX_CMS_CELERY_WORKERS }} --max-tasks-per-child 100 --exclude-queues=edx.lms.core.default
```
dkaliberda, line 180: The docker-compose also provides mechanisms for managing replicas.

tutor/templates/local/docker-compose.yml
arbrandes: In theory, yes.  I'd be glad to review a PR that does that instead.  I just need a way to reduce RAM usage for development. ;)

DawoudSheraz:
Not sure if it is Mac thing but I can't see anything against `pgrep celery` in LMS/CMS worker, using bash (both dev and local). However, I can see the max_concurrency is set to number of allocated CPUs for both LMS and CMS workers using `celery --app=lms.celery inspect stats`.

DawoudSheraz (approved):

docs/dev.rst
```diff
@@ -60,6 +60,20 @@ Now, use the ``tutor dev ...`` command-line interface to manage the development
...
+Conserving system resources
+---------------------------
+
+When developing, you'll usually want the development environment to take up as little system resources as possible.  To do so, we recommend setting the following Tutor configuration and relaunching your environment::
+
+  tutor config save \
+    --set OPENEDX_CMS_CELERY_WORKERS=1 \
+    --set OPENEDX_LMS_CELERY_WORKERS=1 \
+    --set OPENEDX_CMS_UWSGI_WORKERS=1 \
+    --set OPENEDX_LMS_UWSGI_WORKERS=1 \
```
regisb, line 72: I'd rather avoid asking users to manually set these values. Instead, we should automatically default to workers=1 in development. Can we do that? For instance by overriding the celery config in development?

docs/dev.rst
```diff
@@ -60,6 +60,20 @@ Now, use the ``tutor dev ...`` command-line interface to manage the development
...
+    --set ELASTICSEARCH_HEAP_SIZE=100m
```
regisb, line 73: Same here: can we automatically set this value in development?

docs/configuration.rst
```diff
@@ -149,6 +149,11 @@ This defines the version that will be pulled from just the Open edX platform git
 
 By default, there are 2 `uwsgi worker processes <https://uwsgi-docs.readthedocs.io/en/latest/Options.html#processes>`__ to serve requests for the LMS and the CMS. However, each worker requires upwards of 500 Mb of RAM. You should reduce this value to 1 if your computer/server does not have enough memory.
 
+- ``OPENEDX_LMS_CELERY_WORKERS`` (default: ``"0"``)
+- ``OPENEDX_CMS_CELERY_WORKERS`` (default: ``"0"``)
```
regisb, line 153: Adding new configuration settings to Tutor core is a personal trigger of mine :firecracker: Do we really want to make changes to the default production values? If yes, can we:

1. propose better defaults?
2. make these custom changes possible via a patch instead of two new configuration settings?

arbrandes: Yeah, I understand the reluctance to add new config items.  It's just that in this case, it was the most straightforward way to achieve what I was after.  There's precedent, too: `OPENEDX_LMS_UWSGI_WORKERS` is there for very similar reasons.

Regarding the defaults, I'm not actually changing them: I'm just making them explicit, where before they were explicit.  (The implicit default is to scale the workers to however many CPUs you have, and that's what "0" means.)

As for using patches, I wouldn't mind except for the fact that, as mentioned above, this is just doing what `OPENEDX_LMS_UWSGI_WORKERS` does, except for Celery workers.  If we have that configuration, I don't see why we shouldn't have this one.

All of this said, I really like the idea of changing certain things automatically for development environments, whether they have corresponding config items or not.  For instance, after I issued this PR it came to my attention that Tutor's [importing `*` from devstack.py](https://github.com/overhangio/tutor/blob/master/tutor/templates/apps/openedx/settings/lms/development.py#L3) for the development settings, and that means that we aren't using workers at all!  (See https://github.com/openedx/edx-platform/blob/master/lms/envs/devstack.py#L35.)  So why is `tutor dev` even firing up workers?

In any case, the latter sounds like it warrants a separate PR.  My question regarding _this_ PR, though, is whether we do or do not want `OPENEDX_LMS_CELERY_WORKERS`.  It might not make sense to change this in a Kubernetes setting, but I'm willing to defend that it _does_ on any Docker deployment where you have more CPUs than you have RAM (so to speak).

arbrandes: How about this for not starting workers at all in dev mode?  https://github.com/overhangio/tutor/pull/1041

The question remains whether we still want to let people configure the number of Celery workers manually.  (I say we let them.)

regisb: I love the fact that we can disable workers in dev. I commented on #1041. 

Let's now focus on the possibility to customize the number of celery runners. I agree that this would be a useful feature. If we really have to, we'll introduce new configuration values, but I'd like to see if we can avoid it. For instance, could we avoid that by creating a celery [config file](https://docs.celeryq.dev/en/stable/userguide/configuration.html#id11)? This file would include a `{{ patch("edx-platform-celery-config") }}` statement. That way, we wouldn't have to create new configuration settings for every celery parameter.

DawoudSheraz:
@arbrandes Hi, what's the plan for this PR? Thanks

arbrandes:
I can look into adding a Celery conf file patch, but since I'm not using this PR (as opposed to the one that disables workers in dev mode), it'll probably take a while to get to.

docs/configuration.rst
igobranco: @regisb
Agree! There should be an option to disable workers in dev, it could be disabled by default on dev mode. Personally, I like of having a config file with a patch. By default the config file should have the minimum config to start.
Nevertheless, I feel that won't resolve every additional configuration:
- https://github.com/celery/celery/issues/2566
- https://docs.celeryq.dev/en/stable/reference/cli.html
- https://docs.celeryq.dev/en/stable/userguide/configuration.html this won't have a way to disable gossip, mingle, heartbeat and configure (vertical) autoscale.

regisb: Please see my answer here: https://github.com/overhangio/tutor/issues/1126#issuecomment-2385620266
I propose that all remaining comments are made on issue #1126

arbrandes:
I'm closing this because it seems we're gonna go with a patch/config file solution.  The conversation should continue on #1126.

PR #978, opened 2024-01-12
build: add Github action to auto-add new issues and PRs to project

DawoudSheraz:
### Description
Adds a Github action to auto-add new PRs and issues to Overhang.io project https://github.com/orgs/overhangio/projects/4.

Requires a token with repo and project scopes in env variable ADD_TO_PROJECT_SECRET_TOKEN

- https://github.com/actions/add-to-project?tab=readme-ov-file#inputs
- https://stackoverflow.com/questions/62530078/github-auto-assign-issue-to-project

PR #1011, opened 2024-03-05
build: remove auto-add for PRs

DawoudSheraz:
The auto-add for Issues and PRs was added in https://github.com/overhangio/tutor/pull/978. However, it did not work as expected for PRs. It was failing on PRs created from fork and increasing the noise.
The reason was that when PRs were created from forks, the Personal access token from main repo was not passed. This results in GA not getting the appropriate auth token to auto-add the PR to the project. Additionally,  the workflows on PR needed to be triggered manually (both auto-add and CI), thus keeping the manual intervention just like before.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -1,9 +1,6 @@
-name: Auto Add Issues and Pull Requests to Project
+name: Auto Add Issues to Project
 
 on:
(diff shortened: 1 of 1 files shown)


regisb:
Note: this same PR was already applied to all plugins.

regisb (approved):
:+1:

PR #1015, opened 2024-03-14
Install tutor-indigo by default with tutor[full]

hinakhadim:
As `tutor-indigo` is enabled in `quince` by default but it is not installed when run `pip install tutor[full]`. `tutor plugins list` show this warning: `⚠️  Failed to enable plugin 'indigo': plugin 'indigo' is not installed.`

### **Before:**
It is not installing `tutor-indigo`. I did fresh install of tutor by `pip install tutor[full]`
Screenshot 2024-03-14 at 9 58 45 AM
Screenshot 2024-03-14 at 10 00 43 AM

### **Now:**
It installs `tutor-indigo` by default as other plugins are being installed. 
Screenshot 2024-03-14 at 10 35 24 AM

### DIFF
--- a/requirements/plugins.txt
+++ b/requirements/plugins.txt
@@ -5,6 +5,7 @@ tutor-credentials>=17.0.0,<18.0.0
 tutor-discovery>=17.0.0,<18.0.0
 tutor-ecommerce>=17.0.0,<18.0.0
 tutor-forum>=17.0.0,<18.0.0
+tutor-indigo>=17.0.0,<18.0.0
 tutor-jupyter>=17.0.0,<18.0.0
 tutor-mfe>=17.0.0,<18.0.0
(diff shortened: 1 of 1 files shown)


DawoudSheraz (approved):

requirements/plugins.txt
```diff
@@ -11,3 +11,4 @@ tutor-minio>=17.0.0,<18.0.0
 tutor-notes>=17.0.0,<18.0.0
 tutor-webui>=17.0.0,<18.0.0
 tutor-xqueue>=17.0.0,<18.0.0
+tutor-indigo>=17.0.0,<18.0.0
```
regisb, line 14: Could you please move indigo such that the list of plugins is sorted?

hinakhadim: done

Issue #997, opened 2024-01-31
`tutor mounts add /path-to-openedx-scorm-xblock` doesn't automatically install scorm-xblock


Danyal-Faheem:
**Bug description**

Using the latest mounting command `tutor mounts add /path-to-xblock`, I tried to install [scorm-xblock](https://github.com/overhangio/openedx-scorm-xblock) in my tutor install. This was on a fresh install of tutor and only the default plugins were installed as well. However, tutor did not install scorm-xblock after I had mounted it using the above mentioned command and did not show up in the Advanced tab on a studio course even after I had added `scorm` to the advanced settings list.

**How to reproduce**

On a fresh install of tutor quince 17.0, you can follow the following steps to reproduce the command.

- Clone [scorm-xblock](https://github.com/overhangio/openedx-scorm-xblock).
- Run `tutor mounts add path-to-scorm-xblock`
- Run `tutor images build openedx-dev`
- Run `tutor dev stop && tutor dev start -d`
- Create a new course or run `tutor dev do importdemocourse`. Under Advanced Module List In the advanced settings of the course, add `scorm` and save changes. Now, while adding a new unit, you should be able to see Scorm Module as an option in the advanced tab but that is not the case here.

The output of `tutor mounts list` for me is:
```
- name: path-to/openedx-scorm-xblock
  build_mounts:
  - image: openedx
    context: mnt-openedx-scorm-xblock
  - image: openedx-dev
    context: mnt-openedx-scorm-xblock
  compose_mounts:
  - service: openedx
    container_path: /mnt/openedx-scorm-xblock
  - service: openedx-dev
    container_path: /mnt/openedx-scorm-xblock
```
The output of `tutor config printvalue OPENEDX_EXTRA_PIP_REQUIREMENTS` for me is:

`[]`

openedx-scorm-xblock does show for me at the `/mnt` path inside the LMS container.
That is also the solution I am currently utilizing which is to follow the following steps:

- `tutor dev run lms bash`
- `cd /mnt`
- `pip install -e ./openedx-scorm-xblock`
- `exit`
- `tutor dev stop && tutor dev start -d`

After these steps, openedx-scorm-xblock works as intended and hot-reloading works as well as it is installed in editable mode.

Surprisingly enough, while running `tutor images build openedx-dev` after adding the xblock as a mount, it seems that tutor does try to install `openedx-scorm-xblock` in editable mode as can be seen in the screenshot below.

[...]

PR #1016, opened 2024-03-15
fix: ensure mounted installable packages are installed as expected

DawoudSheraz:
Fixes https://github.com/overhangio/tutor/issues/997

When xblocks and other installed packages are mounted, they are installed during image build. The dirs are copied over and then pip install takes place (resulting in egg-file generation). However, the dirs are not updated in host directories. When running tutor, the host files are re-mounted at run time. Due to missing egg-info, the installed packages no longer appear in pip list. This PR adds a small step during initialization that runs pip install on the mounted dirs if egg-info is missing.

### Testing
- Checkout the branch & install.
- Clone scorm or any other xblock (as mentioned in https://github.com/overhangio/tutor/issues/997)
- mount the xblock
- Re-build open edx image. This might not be needed if you already have recent image build. In that case, running `tutor dev do init -l lms` would be enough
- Bash into lms and run `pip list | grep mnt`. The mounted xblock(s) will show up.

### DIFF
--- a/tutor/templates/jobs/init/mounted-directories.sh
+++ b/tutor/templates/jobs/init/mounted-directories.sh
@@ -1,20 +1,21 @@
 # The initialization job contains various re-install operations needed to be done
-# on mounted directories (edx-platform, /mnt/*xblock)
-# 1. /mnt/*xblock
-# Whenever xblocks are mounted, during the image build time, they are copied over to container and
-# installed. This results in egg_info generation for the mounted xblocks. However, the egg_info is not carried
+# on mounted directories (edx-platform, /mnt/*xblock, /mnt/<edx-ora, search, enterprise>)
+# 1. /mnt/*
+# Whenever xblocks or other installable packages are mounted, during the image build, they are copied over to container
+# and installed. This results in egg_info generation for the mounted directories. However, the egg_info is not carried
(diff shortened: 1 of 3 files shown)


tutor/templates/jobs/init/mounted-directories.sh
```diff
@@ -1,7 +1,23 @@
...
+# installed. This results in egg_info generation for the mounted xblocks. However, the egg_info is not carried
+# over to host. When the containers are launched, the host directories without egg_info are mounted on runtime
+# and disappear from pip list.
+# 2. edx-platform
 # When a new local copy of edx-platform is bind-mounted, certain build
 # artifacts from the openedx image's edx-platform directory are lost.
 # We regenerate them here.
 
+
+for mounted_xblock in /mnt/*xblock/; do
```
regisb, line 14: We should not iterate just on xblocks, because we can mount other apps in this directory. See the list of auto-mounted directories here: https://github.com/overhangio/tutor/blob/f211b982fd8d970a538eb8452e16259a5eaae92b/tutor/plugins/openedx.py#L62

I think it should be safe to iterate on all directories that include a setup.py file.

Danyal-Faheem (approved):
LGTM! 👍 
I've tested it out on Tutor version 17.0.2 and it's working perfectly.

PR #1021, opened 2024-03-19
Adds CONFIG_USER Filter

abonnell:
[Context](https://discuss.openedx.org/t/tutor-config-unique-filter-not-overwriting-16-1-4/11712/2)

Implements a `CONFIG_USER` Filter that users can hook into and replace key values in their `config.yml` files

Similar in usage to the `CONFIG_UNIQUE` filter but it will always replace the value of a given key with the new value being passed in instead of skipping over it

Example usage in a tutor plugin, being used to fetch secrets from AWS Secrets Manager and put that dictionary of key:value pairs into config.yml to be ingested by tutor patches
```
@hooks.Filters.CONFIG_USER.add()
def fetch_secrets_from_aws(config):
    secrets = s.do_config()

    for secret in secrets:
        config.append(
            (secret, secrets[secret])
        )
        
    return config
```

### DIFF
--- a/changelog.d/20240319_140241_andrew.bonnell_add_config_user_filter.md
+++ b/changelog.d/20240319_140241_andrew.bonnell_add_config_user_filter.md
@@ -1,0 +1,1 @@
+- [Feature] Introduces the CONFIG_USER Filter. (by @abonnell)
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -111,6 +111,12 @@ def get_user(root: str) -> Config:
         config = get_yaml_file(path)
     upgrade_obsolete(config)
     update_with_env(config)
+
+    extra_config: list[tuple[str, ConfigValue]] = []
+    extra_config = hooks.Filters.CONFIG_USER.apply(extra_config)
+    for name, value in extra_config:
+        config[name] = value
+    
     return config
 
 
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -112,9 +112,7 @@ def get_user(root: str) -> Config:
     upgrade_obsolete(config)
     update_with_env(config)
 
-    extra_config: list[tuple[str, ConfigValue]] = []
-    extra_config = hooks.Filters.CONFIG_USER.apply(extra_config)
-    for name, value in extra_config:
+    for name, value in hooks.Filters.CONFIG_USER.iterate():
         config[name] = value
     
     return config
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -114,7 +114,6 @@ def get_user(root: str) -> Config:
 
     for name, value in hooks.Filters.CONFIG_USER.iterate():
         config[name] = value
-    
     return config
 
 
--- a/tutor/hooks/catalog.py
+++ b/tutor/hooks/catalog.py
@@ -257,9 +257,12 @@ class Filters:
     #:   names must be prefixed with the plugin name in all-caps.
     CONFIG_UNIQUE: Filter[list[tuple[str, Any]], []] = Filter()
 
-    #: Declare unique configuration settings that must be saved in the user ``config.yml`` file. This is where
-    #: you should declare passwords and randomly-generated values that are different from one environment to the next.
-    #: Callbacks using this Filter will overwrite the values of existing keys
+    #: Used to declare unique key:value pairs in the ``config.yml`` file that will be overwritten on ``tutor config save``.
+    #: This is where you should declare passwords and other secrets that need to be fetched live from an external secrets
+    #: store. Most users will not need to use this filter but it will allow you to programmatically fetch and set secrets
+    #: from an external secrets store such as AWS Secrets Manager via boto3.
+    #:
+    #: Values passed in to this filter will overwrite existing values in the ``config.yml`` file.
     #:
     #: :parameter list[tuple[str, ...]] items: list of (name, value) new settings. All
     #:   names must be prefixed with the plugin name in all-caps.
--- a/tutor/hooks/catalog.py
+++ b/tutor/hooks/catalog.py
@@ -257,6 +257,14 @@ class Filters:
     #:   names must be prefixed with the plugin name in all-caps.
     CONFIG_UNIQUE: Filter[list[tuple[str, Any]], []] = Filter()
 
+    #: Declare unique configuration settings that must be saved in the user ``config.yml`` file. This is where
+    #: you should declare passwords and randomly-generated values that are different from one environment to the next.
+    #: Callbacks using this Filter will overwrite the values of existing keys
+    #:
+    #: :parameter list[tuple[str, ...]] items: list of (name, value) new settings. All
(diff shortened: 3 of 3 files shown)


tutor/hooks/catalog.py
```diff
@@ -257,6 +257,14 @@ def your_filter_callback(some_data):
     #:   names must be prefixed with the plugin name in all-caps.
     CONFIG_UNIQUE: Filter[list[tuple[str, Any]], []] = Filter()
 
+    #: Declare unique configuration settings that must be saved in the user ``config.yml`` file. This is where
```
regisb, line 260: The docstring should be updated. Please try to emphasise that most people should not need this, and describe in which specific scenarios this filter should be used.

tutor/config.py
```diff
@@ -111,6 +111,12 @@ def get_user(root: str) -> Config:
         config = get_yaml_file(path)
     upgrade_obsolete(config)
     update_with_env(config)
+
+    extra_config: list[tuple[str, ConfigValue]] = []
+    extra_config = hooks.Filters.CONFIG_USER.apply(extra_config)
+    for name, value in extra_config:
```
regisb, line 117: We have some syntactic sugar to write this more simply:

```
for name, value in hooks.Filters.CONFIG_USER.iterate():
    ...
```

regisb (changes requested):
thanks for rising up to the challenge! I just have a couple minor comments. Did you try out this change and verify that it works for your use case?

tutor/config.py
abonnell: Updated to
```
def get_user(root: str) -> Config:
    """
    Get the user configuration from the tutor root.

    Overrides from environment variables are loaded as well.
    """
    convert_json2yml(root)
    path = config_path(root)
    config = {}
    if os.path.exists(path):
        config = get_yaml_file(path)
    upgrade_obsolete(config)
    update_with_env(config)

    for name, value in hooks.Filters.CONFIG_USER.iterate():
        config[name] = value
    
    return config
    ```
    
    That's much cleaner, thanks for the tip!

File: tutor/hooks/catalog.py (hunk 1 above)
Comment (abonnell) [reply tutor/hooks/catalog.py]: Updated to for extra specificity

```
    #: Used to declare unique key:value pairs in the ``config.yml`` file that will be overwritten on ``tutor config save``.
    #: This is where you should declare passwords and other secrets that need to be fetched live from an external secrets
    #: store. Most users will not need to use this filter but it will allow you to programmatically fetch and set secrets
    #: from an external secrets store such as AWS Secrets Manager via boto3.
    #:
    #: Values passed in to this filter will overwrite existing values in the ``config.yml`` file.
    #:
    #: :parameter list[tuple[str, ...]] items: list of (name, value) new settings. All
    #:   names must be prefixed with the plugin name in all-caps.
    CONFIG_USER: Filter[list[tuple[str, Any]], []] = Filter()
```

**Comment by abonnell** [comment]
> thanks for rising up to the challenge! I just have a couple minor comments. Did you try out this change and verify that it works for your use case?

Yes! Its working pretty cleanly, no real issues with it within the scope of what its trying to do!

**Comment by abonnell** [comment]
@regisb have you had a chance to review after my changes? Thanks! Apologies if I made changes for the review incorrectly, my org doesn't use github and I'm not super familiar with its review process

**Comment by DawoudSheraz** [comment]
@abonnell Hi. I will add a review on Monday. It has been on my todo, but did not get time to look at it. Apologies for the delay. Thanks

File: tutor/config.py [hunk 3]
```diff
@@ -111,6 +111,10 @@ def get_user(root: str) -> Config:
         config = get_yaml_file(path)
     upgrade_obsolete(config)
     update_with_env(config)
+
+    for name, value in hooks.Filters.CONFIG_USER.iterate():
+        config[name] = value
```
DawoudSheraz, line 116: question: If we are fetching the secrets from a secret manager, this will hit the manager every time the config is loaded. Should the consumer of CONFIG_USER be caching results to avoid duplicate/unnecessary calls?

changelog.d/20240319_140241_andrew.bonnell_add_config_user_filter.md
```diff
@@ -0,0 +1 @@
+- [Feature] Introduces the CONFIG_USER Filter. (by @abonnell)
```
DawoudSheraz, line 1: nit: You can add some more details of the filter here. 1-2 liner explaining what can be expected from this. I understand the info is there in docstring but changelog are important for the larger community.

tutor/config.py
abonnell: Thanks for the feedback! In my mind it was going to be on the code that's fetching secrets to implement caching, and then it will return a full list of key:value tuples to the filter itself so as to save complexity on tutor's side (and give users more flexibility in how they actually implement this filter). Thoughts?

DawoudSheraz: Yes, makes sense. When I say consumer, I also meant the code that would be using the filter.

abonnell: Ah then yes, I would say the code that is feeding the filter should cache results but ultimately that will be down to the specific implementations of devs and their org's needs/requirements

For the example on the root of the PR `s.do_config()` is reaching out to a singular AWS Secrets Manager secret with many keys in the JSON payload, so its just a single request for us at the end of the day, but other orgs could cache timestamps for the values and compare those as needed prior to making the request!

changelog.d/20240319_140241_andrew.bonnell_add_config_user_filter.md
abonnell: https://github.com/overhangio/tutor/pull/1021/commits/d0b3229f16bb612582bb431a418c0ff8b5ea775c

Updated changelog message to be a bit more descriptive

abonnell (commented):
Updated changelog, hook decription, provided clarity to dawoud

changelog.d/20240319_140241_andrew.bonnell_add_config_user_filter.md
```diff
@@ -0,0 +1 @@
+- [Feature] Introduces the CONFIG_USER Filter. Used to declare unique key:value pairs in config.yml that will be overwritten when running tutor config save. Useful for injecting secrets from an external secrets manager into edx, or other values that may change over time that you need to programmatically fetch. (by @abonnell)
```
DawoudSheraz, line 1: nit: Useful for injecting secrets from an external secrets manager into edx, or other values that may change over time that you need to programmatically fetch --> The filter can be used for injecting secrets from an external secrets manager into Open edX deployment, or other values that may change over time and need to be fetched programmatically

DawoudSheraz (approved):

abonnell:
Apologies @DawoudSheraz , missed a formatting change. Double checked `make test` and it seems good now

abonnell (commented):
Updates with Dawoud's feedback as well

regisb (approved):
lgtm!

PR #1022, opened 2024-03-20
chore: update redis

DawoudSheraz:
- Update redis to latest available version.

I updated the redis, pulled in latest images (when running launch), enabled discovery, and poked around a bit on Studio and Discovery. I did not see anything unusual in Redis logs. Only one error, some PID save, occurs at start of container. Though that error is also present for 7.2.1.

### DIFF
--- a/changelog.d/20240320_131420_dawoud.sheraz_redis_update.md
+++ b/changelog.d/20240320_131420_dawoud.sheraz_redis_update.md
@@ -1,0 +1,1 @@
+- [Improvement] Update Redis to 7.2.4 (by @dawoudsheraz)
--- a/changelog.d/20240320_131420_dawoud.sheraz_redis_update.md
+++ b/changelog.d/20240320_131420_dawoud.sheraz_redis_update.md
@@ -1,1 +1,1 @@
-- [Improvement] Update Redis to 7.2.4 (by @dawoudsheraz)
+- [Security] Update Redis to 7.2.4 (by @dawoudsheraz)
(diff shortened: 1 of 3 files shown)


tutor/templates/config/defaults.yml
```diff
@@ -22,7 +22,7 @@ DOCKER_IMAGE_MONGODB: "docker.io/mongo:4.4.25"
 DOCKER_IMAGE_MYSQL: "docker.io/mysql:8.1.0"
 DOCKER_IMAGE_PERMISSIONS: "{{ DOCKER_REGISTRY }}overhangio/openedx-permissions:{{ TUTOR_VERSION }}"
 # https://hub.docker.com/_/redis/tags
-DOCKER_IMAGE_REDIS: "docker.io/redis:7.2.1"
+DOCKER_IMAGE_REDIS: "docker.io/redis:7.2.4"
```
regisb, line 25: You should also update the default value of `DOCKER_IMAGE_REDIS` in the tutor docs. (and this value is outdated by the way)

changelog.d/20240320_131420_dawoud.sheraz_redis_update.md
```diff
@@ -0,0 +1 @@
+- [Improvement] Update Redis to 7.2.4 (by @dawoudsheraz)
```
regisb, line 1: This is a security upgrade, so it should be labelled as `[Security]`.

regisb (changes requested):

regisb (approved):
thanks Dawoud!

PR #1026, opened 2024-03-26
fix: clean translations before pulling them

kdmccormick:
Blocked by https://github.com/openedx/edx-platform/pull/34428

Some more discussion here: https://openedx.slack.com/archives/C04R6TUJB7T/p1711482071806759?thread_ts=1711478732.261049&cid=C04R6TUJB7T
___________________

## Description

If one is using a bind-mounted edx-platform directory, they may already have pulled translations. This will cause the Docker build to fail with:

    CommandError: "/openedx/edx-platform/conf/plugins-locale/plugins" should be empty before running atlas pull.

To fix this, we simply clean out the edx-platform translation target directories before pulling in the Dockerfile.


kdmccormick:
Ack, this fails with:
```
 > [production 18/34] RUN make clean_translations:                                  
0.194 git clean -fdX conf/locale  # Removes core edx-platform translations but keeps Esperanto (eo) test translations
0.196 fatal: not a git repository (or any of the parent directories): .git
0.196 make: *** [Makefile:57: clean_translations] Error 128
```
This is happening because we intentionally avoid copying `.git` in at this layer, as that would bloat the layer and cause its cache to be invalidated with any git state change whatsoever.

I'm going to turn this into a draft, a propose an alternate upstream implementation of `make clean_translations` which doesn't rely on git.

kdmccormick:
The edx-platform PR linked above merged, so `make clean_translations` now works in the Dockerfile.

I tested this locally and confirmed that it fixes the error.

DawoudSheraz (approved):

Issue #1020, opened 2024-03-19
Upgrade to Mongo 7


kdmccormick:
(Originally mentioned here: https://discuss.openedx.org/t/tutor-cs-comment-service-mongo-and-ruby-upgrade/12472)

edx-platform and cs_comments_service are now both being tested with Mongo 7 (https://github.com/openedx/edx-platform/pull/34213), so we can now upgrade Tutor to use Mongo 7. We should apply the change to Tutor Nightly, the sooner the better. It would be ideal to do it before the Redwood cut (late April), although the true deadline is the Redwood release (June 9).

The line that needs to be changed: https://github.com/overhangio/tutor/blob/nightly/tutor/templates/config/defaults.yml#L23

DawoudSheraz:
https://discuss.openedx.org/t/tutor-edx-platform-mongo-7-upgrade/12512 also confirms that it is safe to upgrade Mongo as both cs_comments_service and edx-platform are running on Mongo 7.

regisb:
We would also (probably) need to update the upgrade scripts, in tutor/commands/upgrade, to make sure that existing MongoDb 4 databases are successfully migrated to v7.

DawoudSheraz:
Since there have been major releases since 4.4, the upgrade would need to go through them before running on Mongo 7. Therefore, during upgrade command, mongo would need to follow 4.4 --> 5.0 --> 6.0 --> 7.0 to work correctly. 

- https://www.mongodb.com/community/forums/t/mongodb-upgrade-from-4-2-and-4-4-to-6-0/205176
- https://www.mongodb.com/community/forums/t/restore-backup-from-mongo-4-2-to-7/242906

There are some other changes across various versions:
- db.adminCommand needs confirm: true when upgrading and setting compatibility to v7 (https://www.mongodb.com/docs/manual/reference/command/setFeatureCompatibilityVersion/#confirm)
- Starting with mongo 6.1, the `--nojournal` commandline argument is no longer needed as journal is enabled by default (https://www.mongodb.com/docs/manual/reference/configuration-file-settings-command-line-options-mapping/#changelog)
- mongo binary is not available in docker images starting with v6. mongosh should be used in its place (https://github.com/docker-library/mongo/issues/558, https://github.com/docker-library/mongo/pull/553#issuecomment-1192008008)

PR #1029, opened 2024-03-29
mongodb update

DawoudSheraz:
Resolves https://github.com/overhangio/tutor/issues/1020

Because mongo is going from 4.4 --> 7, starting containers directly wouldn't work out of box. The container will keep on restarting with featureCompatibility error. Before running the containers, run `tutor <dev/local/k8s> upgrade --from=quince` to ensure Mongo is upgraded properly. See https://github.com/overhangio/tutor/issues/1020#issuecomment-2024794174 for more context for certain changes.

### DIFF
--- a/changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
+++ b/changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
@@ -1,10 +1,10 @@
 - 💥[Feature] Update MongoDB to v7.0.7 (by @dawoudsheraz)
 
-MongoDB is updating from 4.4 to 7. Since there have been major releases since 4.4, the upgrade would need to go through them before running Mongo 7. Mongo would need to follow 4.4 --> 5.0 --> 6.0 --> 7.0 upgrade path to work correctly.
-The container will keep on restarting with featureCompatibility error if the upgrade path is not followed. 
+  MongoDB is updating from 4.4 to 7. Since there have been major releases since 4.4, the upgrade would need to go through them before running Mongo 7. Mongo would need to follow 4.4 --> 5.0 --> 6.0 --> 7.0 upgrade path to work correctly.
+  The container will keep on restarting with featureCompatibility error if the upgrade path is not followed. 
 
-To upgrade mongo, run the following command based in the appropriate environment:
-```
-tutor <dev|local|k8s> upgrade --from=quince
-``` 
-For k8s only, the above command will not perform the upgrade automatically. Instead, the command will output will print out a series of commands that would need to be run manually to carry out the upgrade.
+  To upgrade mongo, run the following command based in the appropriate environment:
+  ```
+  tutor <dev|local|k8s> upgrade --from=quince
+  ``` 
+  For k8s only, the above command will not perform the upgrade automatically. Instead, the command will output a series of commands that would need to be run manually to carry out the upgrade.
--- a/changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
+++ b/changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
@@ -1,0 +1,10 @@
+- 💥[Feature] Update MongoDB to v7.0.7 (by @dawoudsheraz)
+
+MongoDB is updating from 4.4 to 7. Since there have been major releases since 4.4, the upgrade would need to go through them before running Mongo 7. Mongo would need to follow 4.4 --> 5.0 --> 6.0 --> 7.0 upgrade path to work correctly.
+The container will keep on restarting with featureCompatibility error if the upgrade path is not followed. 
+
+To upgrade mongo, run the following command based in the appropriate environment:
+```
+tutor <dev|local|k8s> upgrade --from=quince
+``` 
+For k8s only, the above command will not perform the upgrade automatically. Instead, the command will output will print out a series of commands that would need to be run manually to carry out the upgrade.
--- a/tutor/commands/upgrade/common.py
+++ b/tutor/commands/upgrade/common.py
@@ -49,7 +49,7 @@ def get_mongo_upgrade_parameters(
     """
     Helper utility to get parameters required during mongo upgrade.
     """
-    mongo_version = int(docker_version.split('.')[0])
+    mongo_version = int(docker_version.split(".")[0])
     admin_command: dict[str, int | str] = {
         "setFeatureCompatibilityVersion": compatibility_version
     }
--- a/tutor/commands/upgrade/common.py
+++ b/tutor/commands/upgrade/common.py
@@ -49,7 +49,7 @@ def get_mongo_upgrade_parameters(
     """
     Helper utility to get parameters required during mongo upgrade.
     """
-    mongo_version = int(docker_version[0])
+    mongo_version = int(docker_version.split('.')[0])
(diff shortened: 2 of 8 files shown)


Danyal-Faheem:
I've tested these changes locally on `tutor k8s` with https off and they working just fine for me after making this one small change:

Removing `--nojournal` from the arguments [here](https://github.com/DawoudSheraz/tutor/blob/0f3ec665214ba73ac3d93a7bf0f3a903cf5c7234/tutor/templates/k8s/deployments.yml)

I've tried the following steps for testing:

1. Create super user
2. Import demo course
3. Remove a subsection from demo course
4. Add a subsection and unit to the demo course
5. Add a scorm module
6. Change start date and publish my changes
7. All the changes were reflected on the LMS

DawoudSheraz:
> I've tested these changes locally on `tutor k8s` with https off and they working just fine for me after making this one small change:
> 
> Removing `--nojournal` from the arguments [here](https://github.com/DawoudSheraz/tutor/blob/0f3ec665214ba73ac3d93a7bf0f3a903cf5c7234/tutor/templates/k8s/deployments.yml#L355)
> 
> I've tried the following steps for testing:
> 
> 1. Create super user
> 2. Import demo course
> 3. Remove a subsection from demo course
> 4. Add a subsection and unit to the demo course
> 5. Add a scorm module
> 6. Change start date and publish my changes
> 
> All the changes were successfully reflected on the LMS

Thanks @Danyal-Faheem , much appreciated.

DawoudSheraz:
@regisb @kdmccormick Reminder for the review, thanks 🙂 .

tutor/commands/upgrade/compose.py
```diff
@@ -168,7 +179,17 @@ def upgrade_mongodb(
         )
         return
 
+    # Just need the main release version for certain checks
+    mongo_version = int(to_docker_version[0])
+    admin_command: t.Dict[str, t.Any] = {
```
regisb, line 184: This is actually a `dict[str, str]` and there is no need to `import typing as t`. Same below in the k8s module.

tutor/commands/upgrade/compose.py
```diff
@@ -168,7 +179,17 @@ def upgrade_mongodb(
...
+        "setFeatureCompatibilityVersion": to_compatibility_version
+    }
+    if mongo_version == 7:
+        # Explicit confirmation is required to upgrade to 7 from 6
+        # https://www.mongodb.com/docs/manual/reference/command/setFeatureCompatibilityVersion/#confirm
+        admin_command.update({"confirm": 1})
```
regisb, line 190: Just making sure I understand what this means: this also works in headless, non-interactive mode, right? I mean that the user is not explicitly asked to manually confirm by typing "y" on their keyboard, right?

tutor/commands/upgrade/k8s.py
```diff
@@ -171,13 +183,23 @@ def upgrade_mongodb(
             "responsibility to upgrade your MongoDb instance to {to_docker_version}."
         )
         return
+    mongo_version = int(to_docker_version[0])
```
regisb, line 186: This makes a lot of shared code between compose.py and k8s.pyu. Can we move it to the commands/upgrade/common.py module?

tutor/templates/k8s/deployments.yml
```diff
@@ -352,7 +352,7 @@ spec:
       containers:
         - name: mongodb
           image: {{ DOCKER_IMAGE_MONGODB }}
-          args: ["mongod", "--nojournal", "--storageEngine", "wiredTiger"]
```
regisb, line 355: Is this a breaking change? What's the impact of removing that option?

changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] Update MongoDB to v7.0.7 (by @dawoudsheraz)
```
regisb, line 1: Please add a line telling nightly users to run the upgrade command.

tutor/templates/k8s/deployments.yml
DawoudSheraz: I am not sure if we can classify it as a breaking change. Starting with Mongo 6.1, the journal module is always enabled https://www.mongodb.com/docs/manual/reference/configuration-file-settings-command-line-options-mapping/#changelog and the nojournal param is no longer available.

tutor/commands/upgrade/compose.py
DawoudSheraz: When "confirm": 1 is added, the value is int and then quality checks fail.

regisb: Then we need the typing import, but not `Any`, because we can use `t.Union[int, str]` as a value type.

changelog.d/20240329_164354_dawoud.sheraz_mongo_update.md
```diff
@@ -0,0 +1,10 @@
...
+
+MongoDB is updating from 4.4 to 7. Since there have been major releases since 4.4, the upgrade would need to go through them before running Mongo 7. Mongo would need to follow 4.4 --> 5.0 --> 6.0 --> 7.0 upgrade path to work correctly.
+The container will keep on restarting with featureCompatibility error if the upgrade path is not followed. 
+
+To upgrade mongo, run the following command based in the appropriate environment:
+```
+tutor <dev|local|k8s> upgrade --from=quince
+``` 
+For k8s only, the above command will not perform the upgrade automatically. Instead, the command will output will print out a series of commands that would need to be run manually to carry out the upgrade.
```
regisb, line 10: typo: "will output will print out"

regisb (approved):
looking good! Just a minor typo in the changelog.

tutor/commands/upgrade/common.py
```diff
@@ -41,6 +43,23 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
     )
 
 
+def get_mongo_upgrade_parameters(
+    docker_version: str, compatibility_version: str
+) -> tuple[int, dict[str, int | str]]:
+    """
+    Helper utility to get parameters required during mongo upgrade.
+    """
+    mongo_version = int(docker_version[0])
```
kdmccormick, line 52: Minor tweak to ensure that this function doesn't break at Mongo v10
```suggestion
    mongo_version = int(docker_version.split('.')[0])
```

tutor/commands/upgrade/compose.py
```diff
@@ -46,6 +46,9 @@ def upgrade_from(context: click.Context, from_release: str) -> None:
     if running_release == "palm":
         running_release = "quince"
 
+    if running_release == "quince":
+        upgrade_from_quince(context, config)
+
```
kdmccormick, line 51: Will Nightly users be able to upgrade by running `tutor ... upgrade --from=quince` ? If so, shall we send that message out to #tutor-maintenance once this merges.?

kdmccormick (approved):
Assuming you agree with these comments, this looks good to me. Thanks @DawoudSheraz .

tutor/commands/upgrade/compose.py
DawoudSheraz: Yes 👍🏽

PR #1041, opened 2024-04-16
feat!: don't run Celery workers in dev mode

arbrandes:
Tutor's importing * from devstack.py[1] for the development settings, and that means that we aren't using Celery workers at all in dev mode (see [2]).  This removes them from the dev compose file, thus saving everyboding a significant chunk of RAM.

[1] https://github.com/overhangio/tutor/blob/master/tutor/templates/apps/openedx/settings/lms/development.py#L3
[2] https://github.com/openedx/edx-platform/blob/master/lms/envs/devstack.py#L35

To do so, we rely on Docker compose profiles[3].
    
[3] https://docs.docker.com/compose/profiles/
    
BREAKING CHANGE: the `COMPOSE_PROJECT_STARTED` hook signature had to be changed to accomodate profile selection.

### DIFF
--- a/changelog.d/20240416_162659_arbrandes_no_workers_in_dev.md
+++ b/changelog.d/20240416_162659_arbrandes_no_workers_in_dev.md
@@ -1,0 +1,2 @@
+- 💥[Feature] Use Docker compose profiles to control services. (by @arbrandes) -->
+- [Fix] Don't start Celery workers in dev mode, as they're never used. (by @arbrandes) -->
--- a/tutor/templates/local/docker-compose.yml
+++ b/tutor/templates/local/docker-compose.yml
@@ -170,6 +170,8 @@ services:
       {%- endfor %}
     depends_on:
       - lms
+    profiles:
+      - local
 
   cms-worker:
     image: {{ DOCKER_IMAGE_OPENEDX }}
@@ -189,5 +191,7 @@ services:
       {%- endfor %}
     depends_on:
       - cms
+    profiles:
+      - local
 
   {{ patch("local-docker-compose-services")|indent(2) }}
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -23,8 +23,9 @@ from tutor.types import Config
 
 
 class ComposeTaskRunner(BaseComposeTaskRunner):
-    def __init__(self, root: str, config: Config):
+    def __init__(self, root: str, config: Config, profile: str):
         super().__init__(root, config)
+        self.profile = profile
         self.project_name = ""
         self.docker_compose_files: list[str] = []
         self.docker_compose_job_files: list[str] = []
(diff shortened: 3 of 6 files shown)


changelog.d/20240416_162659_arbrandes_no_workers_in_dev.md
```diff
@@ -0,0 +1,2 @@
+- 💥[Feature] Use Docker compose profiles to control services. (by @arbrandes) -->
+- [Fix] Don't start Celery workers in dev mode, as they're never used. (by @arbrandes) -->
```
DawoudSheraz, line 2: We would need some sort of guidelines to understand how the workers can be setup in dev mode, especially now that we are using docker compose profiles.

tutor/templates/local/docker-compose.yml
```diff
@@ -171,6 +171,8 @@ services:
       {%- endfor %}
     depends_on:
       - lms
+    profiles:
```
regisb, line 174: I wasn't familiar with profiles, to be honest. I like the idea of not running workers in dev, but I'd like to avoid making a breaking change to the filter API, and also changes across many python functions.

Instead, could we simply move the lms-worker and cms-worker declarations from docker-compose.yml to docker-compose.prod.yml?

arbrandes: I don't think it's going to be that simple.  There are other places in the code that assume those service definitions exist.  For instance, https://github.com/overhangio/tutor/blob/master/tutor/commands/compose.py#L289.  Plus, I suspect profiles might be useful down the road, woudn't you say so?

If you're dead set against it, though, there is another way: we can just set the worker services in the dev compose override file to a profile like "donotstart".  We get the same thing this PR provides, but without actually supporting different profiles.

kdmccormick:
Haha, I have a competing proposal (although I let it languish, because I was never able to confirm or deny that `pdb` would work on these workers):

* https://github.com/overhangio/tutor/pull/928

My thinking was that it would be *good* to run these tasks asynchronously, because it would expose developers to the same race conditions that inevitably always happen with Celery tasks in production. But I can see the other side of the argument: async tasks are harder to debug and the workers use resources.

Either way, I think we can all agree that we should either use these workers, or turn them off. So I see four paths forward:

1. Use this PR, which will disable workers in dev. Open edX contributors who work on Celery tasks should take note to manually test their code using `tutor local`.
2. Use my PR, which will enable Celery tasks by default for everybody.
3. Create a new Tutor `ASYNC_CELERY_TASKS` configuration flag, enabled by default. Disabling it would remove the Celery workers from docker-compose and configure Django to run the tasks in-process.
4. Same as (3), but implement it as a `tutor-celery` plugin. Like tutor-mfe, it would be installed and enabled by default.

My vote is (1), but I would support any of them.

DawoudSheraz:
@arbrandes Hi, what's the plan for this PR? Thanks

arbrandes:
@DawoudSheraz, good question.  I still used it locally all the time.  I figure I could implement the alternative I describe above that has a smaller footprint (setting a fake docker profile for the workers in the dev override docker-compose), which addresses Regis' concerns and gets us the same immediate effect of not having workers in dev.  (But we lose the option of using profiles for other things in the future.)

I'll issue a separate PR so we can compare.

regisb:
Hey Adolfo, your PR completely fell off my radar, I'm sorry about that. A recent issue (#1126) has drawn my attention on celery again. I found a not-so-hackish way of disabling lms-worker and cms-worker in development, without using profiles. You can check out my PR here: https://github.com/overhangio/tutor/pull/1128

While I do agree that profiles are super interesting, and we should probably use them in the future instead of multiple docker-compose files, I think it's a much bigger change that needs extensive testing. Getting rid of celery workers in dev is an obvious improvement for everyone, and I don't want to wait longer to merge it.

regisb:
Closing in favour of #1128.

PR #1044, opened 2024-04-18
build: don't apply release patches in nightly

kdmccormick:
The Dockerfile applies some patches to edx-platform. These patches are only meant to work on named releases. So, the Dockerfile avoids applying the patches to Nightly by checking if EDX_PLATFORM_VERSION is "master" 

However, people running Nightly will often use a *branch off of master*, causing that conditional to miss, which will lead to build errors like this:

> ERROR: failed to solve: process "/bin/sh -c curl -fsSL https://github.com/openedx/edx-platform/commit/ad201cd664b6c722cbefcbda23ae390c06daf621.patch | git am" did not complete successfully: exit code: 128

As a solution, we simply will remove the patches from Tutor's nightly branch. We can also remove the same conditional from Tutor's master branch.

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -49,13 +49,6 @@ RUN git config --global user.email "tutor@overhang.io" \
 {%- if patch("openedx-dockerfile-git-patches-default") %}
 # Custom edx-platform patches
 {{ patch("openedx-dockerfile-git-patches-default") }}
-{%- elif EDX_PLATFORM_VERSION == "master" %}
-# Patches in nightly node
-{%- else %}
-# Patches in non-nightly mode
-# Prevent course structure cache infinite growth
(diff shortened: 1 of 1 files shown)


regisb:
This is not a super robust solution, as it means we need to open a PR in nightly each time a patch lands in master. I suggest instead to define a `IS_NIGHTLY` variable (I'm open to other names): note that this variable would not be a configuration setting, but a variable available at template rendering time.

Then we can replace `{%- elif EDX_PLATFORM_VERSION == "master" %}` by `{%- elif IS_NIGHTLY %}`.

kdmccormick:
Sounds good! I presume that it's OK to determine the value of `IS_NIGHTLY` by looking at `tutor.__about__.__app__.__version_suffix__`? I can take care of that.

kdmccormick:
Closed in favor of https://github.com/overhangio/tutor/pull/1153

PR #1048, opened 2024-04-22
Improve the "Working on edx-platform Python dependencies" documentation

bradenmacdonald:
I couldn't get the edx-platform to persistently use my bind-mounted version of its dependency `openedx-learning`. I was about to write a giant post on Discourse to ask for help when I happened to notice that the documentation _does_ contain the answer I needed ("What if my edx-platform package is not automatically bind-mounted?"). The problem was, this key section was separated from the related section ("Working on edx-platform Python dependencies") by unrelated topics like how to run the unit tests.

So, this PR:
1) Moves that important section of the documentation up to where it's more relevant, and adds an explicit call out.
2) Also updates tutor to be able to auto-detect `openedx-learning` without a plugin - as it is becoming a key platform dependency and at least at OpenCraft, many developers need to use a bind-mounted version of this dependency.

I didn't include a changelog entry as this is primarily a documentation-only change.

### DIFF
--- a/changelog.d/20240425_105236_braden_bind_mount_docs.md
+++ b/changelog.d/20240425_105236_braden_bind_mount_docs.md
@@ -1,0 +1,1 @@
+- [Improvement] Auto-detect bind mounts of openedx-learning for edx-platform (by @bradenmacdonald)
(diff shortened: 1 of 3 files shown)


DawoudSheraz (approved):
nit: since openedx-learning will be auto-magically mounted once the PR is merged (ignoring the doc changes), we should add a changelog entry for that.

bradenmacdonald:
Sure @DawoudSheraz. I added a changelog entry.

PR #1052, opened 2024-04-29
fix: do not fail on start when there are no persistent volume claims

### LINKED ISSUE #1030: K8s deployment fails when there's no persistent volume claims to apply
**Bug description**

I have following options in my `config.yml`:
```yaml
RUN_ELASTICSEARCH: false
RUN_MONGODB: false
RUN_MYSQL: false
RUN_REDIS: false
RUN_SMTP: false
```
It blocks k8s deployment since there's no persistent volume claims to apply:
```
-x tutor k8s start
kubectl get namespaces openedx
NAME      STATUS   AGE
openedx   Active   2d21h
Namespace already exists: skipping creation.
kubectl apply --kustomize /Users/snglth/Development/Abstract/fc-tutor/env --wait --selector app.kubernetes.io/component=volume
error: no objects passed to apply
Error: Command failed with status 1: kubectl apply --kustomize /Users/snglth/Development/Abstract/fc-tutor/env --wait --selector app.kubernetes.io/component=volume
```
I think that it should not be like that.

**How to reproduce**

Described above

**Environment**

macOS 14.4.1, tutor 17.0.3
**Additional context**

snglth:
Close #1030

My solution seems dirty to me, so I'm very much open to suggestions on how to improve it.

### DIFF
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -296,18 +296,11 @@ def start(context: K8sContext, names: List[str]) -> None:
     names = names or ["all"]
     for name in names:
         if name == "all":
-            # Create volumes
-            kubectl_apply(
-                context.root,
-                "--wait",
-                "--selector",
-                "app.kubernetes.io/component=volume",
-            )
-            # Create everything else except jobs
+            # Create everything except jobs
             kubectl_apply(
                 context.root,
                 "--selector",
-                "app.kubernetes.io/component notin (job,volume,namespace)",
+                "app.kubernetes.io/component notin (job,namespace)",
             )
         else:
             kubectl_apply(
--- a/changelog.d/20240619_122432_illia_k8s_no_volumes_no_fail.md
+++ b/changelog.d/20240619_122432_illia_k8s_no_volumes_no_fail.md
@@ -1,0 +1,1 @@
+[Bugfix] Do not fail on start when there are not persistent volume claims to apply (by @snglth)


tutor/commands/k8s.py
```diff
@@ -296,13 +297,18 @@ def start(context: K8sContext, names: List[str]) -> None:
...
-            # Create volumes
-            kubectl_apply(
-                context.root,
-                "--wait",
-                "--selector",
-                "app.kubernetes.io/component=volume",
-            )
+            # Create volumes if exists
+            volumes_manifest = tutor_env.pathjoin(context.root, "k8s/volumes.yml")
+            volumes_content = pathlib.Path(volumes_manifest).read_text()
```
DawoudSheraz, line 302: any particular reason for using Path here instead of open()?

tutor/commands/k8s.py
```diff
@@ -296,13 +297,18 @@ def start(context: K8sContext, names: List[str]) -> None:
...
     for name in names:
         if name == "all":
-            # Create volumes
-            kubectl_apply(
-                context.root,
-                "--wait",
-                "--selector",
-                "app.kubernetes.io/component=volume",
-            )
+            # Create volumes if exists
```
DawoudSheraz, line 300: nit: this comment can be updated to highlight when volumes are not created or vice versa.

tutor/commands/k8s.py
snglth: Just for brevity. Should I reimplement it using `open()`?

DawoudSheraz: Tutor code is using open or io.open. It is better to keep the new code consistent with the existing pattern rather than having a new one. Thanks

regisb:
@snglth can you please address the PR comments such that we merge your changes before the redwood release?

snglth:
@regisb Comment addressed

tutor/commands/k8s.py
```diff
@@ -296,13 +296,20 @@ def start(context: K8sContext, names: List[str]) -> None:
...
         if name == "all":
-            # Create volumes
-            kubectl_apply(
-                context.root,
-                "--wait",
-                "--selector",
-                "app.kubernetes.io/component=volume",
-            )
+            volumes_manifest = tutor_env.pathjoin(context.root, "k8s/volumes.yml")
+            with open(volumes_manifest) as volumes_manifest_file:
```
regisb, line 300: I'm sorry, I should have read this PR more thoroughly. Reading rendered environment files from tutor is really not a great idea (with some minor exceptions). Also, this quick fix would miss quite a few edge cases, such as volumes defined outside of volumes.yml.

Unfortunately I don't have a great and easy-to-implement alternative to suggest. Ideally, we would just ignore errors when command output includes the "error: no objects passed to apply" string. But there is no straightforward way to achieve this.

I explored alternatives, such as telling kubectl to ignore such errors. In my understanding, it's just not a good user API to crash on missing objects. Ideally, we should see only a warning. But I found no way to change the current behaviour.

Do you have alternative ideas?

regisb: In my experience, this is failing because deployments that depend on volumes won't launch if the volumes don't exist already. But maybe this is no longer the case, or I'm hallucinating. Can you please try it out? I agree it would be a great way to fix this.

snglth: I tested this approach and it worked quite well. Here's the log of `tutor k8s start`: 

```other
kubectl get namespaces openedx
Error from server (NotFound): namespaces "openedx" not found
Namespace does not exist: now creating it...
kubectl apply --kustomize /Users/snglth/Development/Abstract/test-tutor/env --wait --selector app.kubernetes.io/component=namespace
namespace/openedx created
kubectl apply --kustomize /Users/snglth/Development/Abstract/test-tutor/env --selector 'app.kubernetes.io/component notin (job,namespace)'
configmap/caddy-config-84d5m8f228 created
configmap/mfe-caddy-config-g4549gh2g6 created
configmap/openedx-config-6f26g64k46 created
configmap/openedx-settings-cms-7kb9hbbfg7 created
configmap/openedx-settings-lms-645456bbt8 created
configmap/openedx-uwsgi-config-f57k74k5cc created
configmap/redis-config-6588tm476h created
service/caddy created
service/cms created
service/elasticsearch created
service/forum created
service/lms created
service/mfe created
service/minio created
service/mongodb created
service/mysql created
service/redis created
service/smtp created
persistentvolumeclaim/caddy created
persistentvolumeclaim/elasticsearch created
persistentvolumeclaim/minio created
persistentvolumeclaim/mongodb created
persistentvolumeclaim/mysql created
persistentvolumeclaim/redis created
deployment.apps/caddy created
deployment.apps/cms created
deployment.apps/cms-worker created
deployment.apps/elasticsearch created
deployment.apps/forum created
deployment.apps/lms created
deployment.apps/lms-worker created
deployment.apps/mfe created
deployment.apps/minio created
deployment.apps/mongodb created
deployment.apps/mysql created
deployment.apps/redis created
deployment.apps/smtp created
```

Notice that `persistentvolumeclaim` objects are being created before deployments. I also pushed my changes.

regisb: Thank. It's much cleaner this way. Can you please just add a changelog entry? https://docs.tutor.overhang.io/tutor.html#contributing

regisb (commented):
looks great to me! Please just add a changelog entry.

Now, if you have a better idea to handle jobs more cleanly, I'm all ears.. ;)

regisb (approved):
thanks for the changes, and sorry for the long review time!

PR #1060, opened 2024-05-06
feat: expose integer Tutor version parts to templates

kdmccormick:
@michaelwheeler

### DIFF
--- a/tutor/env.py
+++ b/tutor/env.py
@@ -64,6 +64,11 @@ def _prepare_environment() -> None:
             ("HOST_USER_ID", utils.get_user_id()),
             ("TUTOR_APP", __app__.replace("-", "_")),
             ("TUTOR_VERSION", __version__),
+            # We assume that TUTOR_VERSION is always MAJOR.MINOR.PATCH.
+            # We assume that MAJOR and MINOR are both integers.
+            # We make no assumptions about PATCH, and thus do not parse it for plugins to use.
+            ("TUTOR_VERSION_MAJOR", int(__version__.split(".")[0])),
+            ("TUTOR_VERSION_MINOR", int(__version__.split(".")[1])),
             ("TUTOR_BRANCH_IS_MAIN", __version_suffix__ == "main"),
             ("is_docker_rootless", utils.is_docker_rootless),
         ],
--- a/changelog.d/20241107_164816_kyle_tutor_version_parts.md
+++ b/changelog.d/20241107_164816_kyle_tutor_version_parts.md
@@ -1,0 +1,1 @@
+- [Feature] Add integer variables `TUTOR_VERSION_MAJOR` and `TUTOR_VERSION_MINOR` to the template context. These are parsed from the existing `TUTOR_VERSION` string variable, which takes the format `"MAJOR.MINOR.PATCH"`. We add them as a convenience to developers who need to maintain version-agnostic Tutor plugins (by @michaelwheeler and @kdmccormick).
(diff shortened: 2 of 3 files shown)


michaelwheeler:
For some context, this is related to a question that I brought up during the Tutor Users' Group about writing plugins that support multiple Tutor versions. We're currently doing something along the lines of...

```
{% if TUTOR_VERSION.startswith("14.") %}
# Do something for Tutor v14 only...
{% else %}
# Do something for Tutor versions after 14...
{% endif %}
```

But having major/minor/patch versions available to the context as integers would allow more precision and flexibility.

regisb:
@michaelwheeler I'd like to understand better your use case. Have you considered having multiple branches for your plugin, one for every major release?

michaelwheeler:
> Have you considered having multiple branches for your plugin, one for every major release? 

In our team, managing multiple branches for our plugins felt like more overhead than adding an occasional conditional. We also think it will slightly simplify the process of upgrading to a newer named release.

arbrandes:
I think this would useful to have as an option for plugin authors, particularly for cases where the plugins don't change that much between releases.

regisb:
Right. I'm still not sure if it makes sense to add these settings to Tutor core. You do know that you could define these variables in your own plugin by adding a callback to the `ENV_TEMPLATE_VARIABLES` filter, right?

michaelwheeler:
> You do know that you could define these variables in your own plugin by adding a callback to the `ENV_TEMPLATE_VARIABLES` filter, right?

I didn't know that!

It still feels to me like Tutor should provide the variable, since the version is a property of Tutor itself, but if that's undesirable for other reasons then maybe `ENV_TEMPLATE_VARIABLES` can get me where I need to go.

tutor/env.py
```diff
@@ -64,6 +64,9 @@ def _prepare_environment() -> None:
             ("HOST_USER_ID", utils.get_user_id()),
             ("TUTOR_APP", __app__.replace("-", "_")),
             ("TUTOR_VERSION", __version__),
+            ("TUTOR_VERSION_MAJOR", int(__version__.split('.')[0])),
+            ("TUTOR_VERSION_MINOR", int(__version__.split('.')[1]))
+            ("TUTOR_VERSION_PATCH", int(__version__.split('.')[2]))
```
kdmccormick, line 69:
```suggestion
            ("TUTOR_VERSION_MINOR", int(__version__.split('.')[1])),
            ("TUTOR_VERSION_PATCH", int(__version__.split('.')[2])),
```

kdmccormick:
Sorry for the delay @michaelwheeler  I think this is ready to go. Can you review it and try it out with one of your plugins?

michaelwheeler:
> I think this is ready to go. Can you review it and try it out with one of your plugins?

Yep, thanks! I'll try to take a look before the end of next week.

michaelwheeler:
Hey @kdmccormick, I tested this out and things work as expected from my perspective. Thanks!

kdmccormick:
@michaelwheeler Would you have any objection to me removing `TUTOR_VERSION_PATCH` from this PR? You'd still have `TUTOR_VERSION_MAJOR` and `TUTOR_VERSION_MINOR`, but the patch version would remain unavailable to plugin authors without manually parsing `TUTOR_VERSION`.

My thinking is that it would be nice, as maintainers, to have some flexibility with the patch version. We don't currently use alpha/beta suffixes or release candidates (`19.2.3a`, `19.2.rc1`, etc.) but I can't say for sure that we never will, so I'd rather not encourage plugin authors to rely on it always being an integer.

PR #1061, opened 2024-05-08
Enable Authentication for Elasticsearch

nihxdr:
This pull request introduces user authentication for Elasticsearch, enhancing our Open edX instance's connectivity with external Elasticsearch services that require authentication.


DawoudSheraz:
@nihxdr Hi. Can you add more context for this change? How will it effect the existing installations?

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -39,6 +39,9 @@
   {% if ELASTICSEARCH_SCHEME == "https" %}"use_ssl": True,{% endif %}
   "host": "{{ ELASTICSEARCH_HOST }}",
   "port": {{ ELASTICSEARCH_PORT }},
+  {% if ELASTICSEARCH_USER != '' and ELASTICSEARCH_PASSWORD != '' %}
+  "http_auth": ("{{ ELASTICSEARCH_USER}}", "{{ ELASTICSEARCH_PASSWORD }}"),
```
regisb, line 43: If we start supporting http auth in Elasticsearch, then we will have also to update all tutor plugins that connect to Elasticsearch (including forum). This is going to be non-trivial.

DawoudSheraz:
With Elasticsearch going to replace with meilisearch in Sumac https://github.com/overhangio/tutor/pull/1141, I don't think this is needed anymore.

PR #1063, opened 2024-05-08
fix: allow course import from Course Authoring MFE

ormsbee:
This commit removes the `CORS_ALLOW_HEADERS` setting from the LMS/Studio config template. This is set separately by LMS and Studio config files in edx-platform and should not be overridden by Tutor, since the header types to allow is application logic and will not vary between sites. This fixes a bug where course import in Studio using the new Course Authoring MFE was broken in Tutor deployments because it required additional headers to be allowed (`content-range` and `content-disposition`). This had been set correctly in edx-platform, but was being overridden by Tutor.


arbrandes (approved):
Indeed, setting this here only makes it so we have to keep tabs on upstream and update it when it does.  I don't see the point of doing so, so I vote for removing it.

kdmccormick:
For historical context, it looks like we added this override here: https://github.com/overhangio/tutor/pull/743#pullrequestreview-1187960393

kdmccormick (changes requested):
I think this might break in `tutor local` mode. Conversation ongoing in Slack, I'll paste a summary here after.

kdmccormick:
I suspect that:
* CORS_ALLOW_HEADERS conditionally set in production.py based on ENABLE_CORS_HEADERS already being true: https://github.com/openedx/edx-platform/blob/master/lms/envs/production.py#L385
* ENABLE_CORS_HEADERS defaults to False. Tutor overrides it to True, but not using the YAML file; instead, it overrides it to True after production.py is already fully processed
* That is probably why we needed to set CORS_ALLOW_HEADERS here in the first place
* This doesn't affect dev mode because [devstack.py sets CORS_ALLOW_HEADERS unconditionally](https://github.com/openedx/edx-platform/blob/master/lms/envs/devstack.py#L298)
* So, @ormsbee will fix upstream production.py to set CORS_ALLOW_HEADERS unconditionally, which will allow us to remove the extra override here

ormsbee:
Thanks @kdmccormick!

Blocked on https://github.com/openedx/edx-platform/pull/34743

kdmccormick (approved):
Using `tutor local run lms/cms ./manage.py lms/cms print_setting CORS_ALLOW_HEADERS`, I confirmed [my suspicion that we needed to override CORS_ALLOW_HEADERS here](https://github.com/overhangio/tutor/pull/1063#issuecomment-2101251649), and I also confirmed that Dave's edx-platform PR removes the need for us to override it. LGTM pending the upstream PR 👍🏻

ormsbee:
The required edx-platform PR (https://github.com/openedx/edx-platform/pull/34743) merged, so merging this one.

changelog.d/20240508_111720_dave_fix_cors_headers.md
```diff
@@ -0,0 +1 @@
+- [Bugfix] Remove CORS_ALLOW_HEADERS setting from the LMS/Studio config template. This is set separately by LMS and Studio config files in edx-platform and should not be overridden by Tutor, since the header types to allow is application logic and will not vary between sites. This fixes a bug where course import in Studio using the new Course Authoring MFE was broken in Tutor deployments because it required additional headers to be allowed (content-range and content-disposition). This had been set correctly in edx-platform, but was being overridden by Tutor. (by @ormsbee)
```
kdmccormick, line 1:
```suggestion
- [Bugfix] Remove CORS_ALLOW_HEADERS setting from the LMS/Studio config template. This setting, which holds site-agnostic application logic, is now consistently set to a reasonable value upstream by LMS and CMS config. Using the upstream values fixes a bug where course import in Studio using the new Course Authoring MFE was broken in Tutor deployments because it required additional headers to be allowed (content-range and content-disposition) (by @ormsbee)
```
Does this looks good @ormsbee ? I can update your commit message to match when I squash-merge.

PR #1065, opened 2024-05-16
feat!: upgrade mysql charset and collation to utf8mb4

### LINKED ISSUE #938: We must migrate the mysql databases from utf8mb3 to utf8mb4
Currently, Open edX runs with MySQL tables that are encoded with utf8mb3. It is time that we upgrade to utf8mb4 (the default).

For reference:
https://github.com/overhangio/tutor/issues/887
https://discuss.openedx.org/t/palm-upgrade-running-rdbms-in-tutor-mysql-vs-mariadb-utf8-vs-utf8mb4/11030

edX.org is currently upgrading their environment to MySQL 8.0 (and utf8mb4). We should use this opportunity to learn from them any pitfalls they faced during the upgrade.

Danyal-Faheem:
fixes #938.

**Changes**
- Changes the default character set of MySQL to `utf8mb4` and the default collation to `utf8mb4_unicode_ci`.
- Also updated the `dosqlshell` command to use `utf8mb4` as the default character set.

**Testing**
- I have tested these changes on tutor dev and tutor K8s. I followed these steps:
  - Ran `tutor [dev|k8s] launch`
  - Imported the demo course and made sure it was working fine
  - Created admin and student users successfully
  - Created a new course with emojis in the title and in subsection titles and made sure changes were being published and reflected in the LMS

**Future Work**

A new do command `tutor local do change-charset-collation` to allow users to upgrade the charset and collation of their tutor installations will be added in a later release of Tutor. A WIP PR is already up at #1079.


changelog.d/20240520_150131_danyal.faheem_upgrade_mysql_utf8mb4.md
```diff
@@ -0,0 +1,2 @@
+- [Feature] Upgrade default charset and collation of mysql to utf8mb4 and utf8mb4_unicode_ci respectively (by @Danyal-Faheem)
```
DawoudSheraz, line 1: nit: this is a breaking change in a capacity, lets add 💥 in changelog to ensure that.

docs/local.rst
```diff
@@ -141,6 +141,21 @@ The default Open edX theme is rather bland, so Tutor makes it easy to switch to
...
+
+    tutor local do change-charset-collation --all-tables --charset=utf8mb4 --collation=utf8mb4_unicode_ci
+
+Alternatively, if you only want to upgrade certain tables or exclude certain tables, you can use the `status` option. To upgrade the `courseware_studentmodule` and `courseware_studentmodulehistory` tables, run::
+
+    tutor local do change-charset-collation --status=include courseware_studentmodule courseware_studentmodulehistory
+
+Tutor performs pattern matching from the start of the table name so you can just enter the name of the app to include/exclude all the tables under that app. To upgrade all the tables in the database except the ones under the student and wiki apps, run::
+
+    tutor local do change-charset-collation --status=exclude student wiki
```
DawoudSheraz, line 157: We should specify that the change is not reversible.

Danyal-Faheem: I've added a warning note at the top.

changelog.d/20240520_150131_danyal.faheem_upgrade_mysql_utf8mb4.md
Danyal-Faheem: Done!

tutor/commands/upgrade/common.py
```diff
@@ -66,3 +66,55 @@ def get_mongo_upgrade_parameters(
     mv '/openedx/data/ora2/SET-ME-PLEASE (ex. bucket-name)' /openedx/data/ora2/openedxuploads
 fi
 """
+
+def get_mysql_change_charset_query(
```
DawoudSheraz, line 70: nit: we should move this util to another file as we are not using this in upgrade anymore.

docs/local.rst
```diff
@@ -141,6 +141,24 @@ The default Open edX theme is rather bland, so Tutor makes it easy to switch to
...
 Out of the box, only the default "open-edx" theme is available. We also developed `Indigo, a beautiful, customizable theme <https://github.com/overhangio/indigo>`__ which is easy to install with Tutor.
 
+Changing the mysql charset and collation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning:: This change is potentially irreversible. It is recommended to make a backup of the MySQL database. See the :ref:`database dump instructions <database_dumps>` to create a DB dump.
+
+Your database's charset and collation might not support specific characters or emojis. To upgrade your charset and collation of all the tables in your database to utf8mb4 and utf8mb4_unicode_ci respectively, run::
+
+    tutor local do change-charset-collation --all-tables --charset=utf8mb4 --collation=utf8mb4_unicode_ci
```
regisb, line 151: In which situation(s) will users have to run this command? Is it for users upgrading from Quince? Is it optional?

docs/local.rst
```diff
@@ -141,6 +141,24 @@ The default Open edX theme is rather bland, so Tutor makes it easy to switch to
...
+
+Alternatively, if you only want to upgrade certain tables or exclude certain tables, you can use the `status` option. To upgrade the `courseware_studentmodule` and `courseware_studentmodulehistory` tables, run::
```
regisb, line 153: All verbatim comments need double backticks. E.g: ``` ``status`` ```.

tutor/commands/jobs.py
```diff
@@ -308,12 +309,72 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
...
+   
+@click.command(
+    short_help="Upgrade mysql to a specific charset and collation",
+    help=(
+        "Upgrade mysql to a specific charset and collation. You can either upgrade all tables, specify only certain tables to upgrade or specify certain tables to exclude from the upgrade process"
+    ),
+    context_settings={"ignore_unknown_options": True},
+)
+@click.option("--all-tables", is_flag=True, help="change all tables in the openedx database")
+@click.option("-s", "--status", type=click.Choice(['include', 'exclude'], case_sensitive=False), help="Whether to include or exclude certain tables/apps")
```
regisb, line 327: This combination of options is not very intuitive. It means that the end users needs to specify at least one option -- and that is not a great pattern, as options are considered... optional :)

Instead, I suggest that:

1. all tables are process by default.
2. If `--include` is specified, then only those tables are processed.
3. If `--exclude` is specified, then those tables are not processed.

Note that with this approach, `--include` and `--exclude` are mutually compatible (`--exclude` would override `--include`).

tutor/commands/jobs.py
```diff
@@ -308,12 +309,72 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
...
     if args:
         command += " " + shlex.join(args)  # pylint: disable=protected-access
     yield ("lms", command)
 
 
+   
+@click.command(
+    short_help="Upgrade mysql to a specific charset and collation",
+    help=(
+        "Upgrade mysql to a specific charset and collation. You can either upgrade all tables, specify only certain tables to upgrade or specify certain tables to exclude from the upgrade process"
```
regisb, line 322: This docstring is misleading. It implies that it allows the user to upgrade to just any charset/collation. Instead, I suggest that we rename this command to something like `convert-mysql-utf8mb4-charset`.

docs/local.rst
Danyal-Faheem: Yes, they can run it after upgrading from Quince. It is optional as the LMS runs fine without it unless you add specific characters such as emojis that will throw DB errors.

tutor/commands/jobs.py
Danyal-Faheem: The original command was meant to allow users to upgrade to any charset/collation, however that caused issues with the connection string. However, since we have removed that option, I think the `convert-mysql-utf8mb4-charset` makes more sense. I'll update it.

tutor/commands/jobs.py
regisb: > Maybe we can prompt the user if they wish to proceed?

Yes, this sounds perfect. Please make sure to include a `-I/--non-interactive` option for headless scripts.

docs/local.rst
regisb: That explanation should be present in the docs :)

tutor/commands/upgrade/common.py
```diff
@@ -66,3 +66,55 @@ def get_mongo_upgrade_parameters(
...
+                tables_loop: LOOP
+                    FETCH cur INTO _table_name;
+
+                    IF done THEN
+                    LEAVE tables_loop;
+                    END IF;
+
+                    SET FOREIGN_KEY_CHECKS = 0;
+                    Select _table_name;
+                    SET @statement = CONCAT('ALTER TABLE ', _table_name, ' CONVERT TO CHARACTER SET {charset} COLLATE {collation};');
```
ormsbee, line 105: Will this change columns that currently use `utf8mb4_bin` collation to be `utf8mb4_unicode_ci`? If so, that's a problem. The relatively small number of columns that use `utf8mb4_bin` are usually doing so because they explicitly want the values to be case sensitive, because case insensitivity on things like course keys has given us a lot of grief in the past.

tutor/commands/upgrade/common.py
```diff
@@ -66,3 +66,55 @@ def get_mongo_upgrade_parameters(
...
+                Select "The following tables will have their charset and collation updated:";
+                OPEN cur;
+                tables_loop: LOOP
+                    FETCH cur INTO _table_name;
+
+                    IF done THEN
+                    LEAVE tables_loop;
+                    END IF;
+
+                    SET FOREIGN_KEY_CHECKS = 0;
```
ormsbee, line 103: Does this cause any issue when trying to do a partial migration that includes the `course_overviews_courseoverview` table? That's the only table I'm aware of that has a varchar primary key and foreign keys from other tables.

tutor/commands/upgrade/common.py
ormsbee: To clarify that last statement, an example was that MongoDB and the Python layer of CourseKeys are case sensitive, but the key stored in our CourseOverview model is not case sensitive. So at one point we had two course runs in production that had keys that differed only by case (because someone had misspelled when doing the initial creation), and updates to one could clobber the values for the other because they were fighting over the same row in CourseOverview for updates. (The inactive course run was still getting automated push updates from an external system.)

I realize that fixing CourseOverview is well outside the scope of this work, and I'm not trying to suggest we tackle that here. I only put it out there as an example of why apps might want to explicitly specify a sensitive field for storing things like identifier keys.

Danyal-Faheem: > Will this change columns that currently use utf8mb4_bin collation to be utf8mb4_unicode_ci?

It seems this is the case. I think we can add a condition that columns with the `utf8mb4_bin` collation should not be upgraded. Any other specific charsets or collations that we should look out for here? 

Or we can just upgrade all the columns that are using `utf8mb3_general_ci` so that if some other column has their charset/collation fixed by the model in the future, we don't have to update the check for them.

ormsbee: Based on [your findings](https://github.com/overhangio/tutor/issues/938#issuecomment-2090059383), it seems like the only forced collations that are not `utf8mb4_unicode_ci` are:

* `utf8mb3_bin`: This should be mapped into `utf8mb4_bin`, since that's the logically equivalent case-sensitive collation.
* `utf8mb4_bin`: This should be preserved as-is, as discussed.
* `utf8mb4_general_ci`: This would be mapped to `utf8mb4_unicode_ci`, but it looks like it only exists for the `bundles` app, and that app is unused in Redwood and removed entirely just after the Redwood cut, so it's fine to ignore or update it, depending on what's more convenient in the code.

Danyal-Faheem: I've updated the command to perform the following mappings instead:

- `utf8mb3_general_ci` to `utf8mb4_unicode_ci`
- `utf8mb3_*` to `utf8mb4_*` meaning that the `utf8mb3_bin` collations are converted to `utf8mb4_bin` now.

> utf8mb4_bin: This should be preserved as-is, as discussed.

We ignore any columns that already have the `utf8mb4` charset.

> utf8mb4_general_ci: This would be mapped to utf8mb4_unicode_ci, but it looks like it only exists for the bundles app, and that app is unused in Redwood and removed entirely just after the Redwood cut, so it's fine to ignore or update it, depending on what's more convenient in the code.

Yup, trying it out on the nightly release, the bundles app was not there. Therefore, I've ignored it.

tutor/commands/upgrade/common.py
Danyal-Faheem: I tested this out by upgrading only the `course_overviews_courseoverviewtab` table which has a foreign key to `course_overviews_courseoverview` to utf8mb4 and keeping the rest at utf8mb3. 

The LMS and CMS seemed to be working fine for me. I tested out the following things:

- Create new Course
- Publish changes in new course
- Enroll in new course and verify everything is working fine
- Edit an existing course
- Publish new course and verify changes can be seen by students on LMS

ormsbee (approved):

docs/local.rst
```diff
@@ -141,6 +141,33 @@ The default Open edX theme is rather bland, so Tutor makes it easy to switch to
...
+Your database's charset and collation might not support specific characters or emojis. Tutor will run fine without this change unless you explicity use specific characters in your instance.
+
+.. warning:: This change is potentially irreversible. It is recommended to make a backup of the MySQL database. See the :ref:`database dump instructions <database_dumps>` to create a DB dump.
+
+To change the charset and collation of all the tables in the openedx database, run::
+
+    tutor local do convert-mysql-utf8mb4-charset
+
+Alternatively, if you only want to change the charset and collation of certain tables or exclude certain tables, you can use the ``--include`` or ``--exclude`` options. To upgrade the ``courseware_studentmodule`` and ``courseware_studentmodulehistory`` tables, run::
+
```
DawoudSheraz, line 158: nit: please mention that include/exclude take comma-separated values and that values can be table or app names. The examples are already there, we only need to make it explicit.

tutor/commands/jobs.py
```diff
@@ -308,12 +309,75 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
...
+   
+@click.command(
+    short_help="Convert the charset and collation of mysql to utf8mb4.",
+    help=(
+        "Convert the charset and collation of mysql to utf8mb4. You can either upgrade all tables, specify only certain tables to upgrade or specify certain tables to exclude from the upgrade process"
+    ),
+    context_settings={"ignore_unknown_options": True},
+)
+@click.option("--include", is_flag=False, nargs=1, help="Apps/Tables to include in the upgrade process. Requires comma-seperated values with no space inbetween.")
+@click.option("--exclude", is_flag=False, nargs=1, help="Apps/Tables to exclude from the upgrade process. Requires comma-seperated values with no space inbetween.")
```
DawoudSheraz, line 327: nit: in-between.

tutor/commands/jobs.py
```diff
@@ -308,12 +309,75 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
...
+    exclude: str,
+    database: str,
+    non_interactive: bool,
+) -> t.Iterable[tuple[str, str]]:
+    """
+    Do command to upgrade the charset and collation of tables in MySQL
+    
+    Can specify whether to upgrade all tables, or include certain tables/apps or to exclude certain tables/apps
+    """
+    # Make sure user hasn't specified both the options or neither of the options
```
DawoudSheraz, line 343: The comment needs an update.

tutor/commands/upgrade/common.py
Danyal-Faheem: Moved this method to the `tutor/utils.py` file.

docs/local.rst
Danyal-Faheem: Added both of these.

tutor/commands/jobs.py
Danyal-Faheem: Updated.

DawoudSheraz (approved):
Please update PR description and add link to the 2nd PR that contains do command work.

changelog.d/20240614_180451_danyal.faheem_upgrade_mysql_utf8mb4.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] Upgrade default charset and collation of mysql to utf8mb4 and utf8mb4_unicode_ci respectively (by @Danyal-Faheem)
```
DawoudSheraz, line 1: We can mention that if the operator is running MySQL on a dedicated server (RUN_MYSQL=False), they are responsible for doing this change manually on their system/instance.

Danyal-Faheem: Added this small note in the changelog entry.

PR #1074, opened 2024-06-04
fix: permissions error on windows when running `tutor dev start`

### LINKED ISSUE #1034: On Windows `tutor dev start` fails because of "Permission denied"
**Bug description**

When running `tutor dev` commands on Windows, the all.log and tracking.log files are owned by root. But the user ID in the lms and cms containers is 1000. As a consequence, runserver commands fail to start.

This issue is caused by the fact that the permissions service sets the OPENEDX_USER_ID to "{{ HOST_USER_ID }}", which is 0:

```
services:
  permissions:
    environment:
      OPENEDX_USER_ID: "{{ HOST_USER_ID }}"
```

But the lms/cms containers run as user 1000:

```
x-openedx-job-service:
  &openedx-job-service
  ...
    args:
      # Note that we never build the openedx-dev image with root user ID, as it would simply fail.
      APP_USER_ID: "{{ HOST_USER_ID or 1000 }}"
```

To resolve the issue, we should update the permissions container to set `OPENEDX_USER_ID: "{{ HOST_USER_ID or 1000 }}"`.

**How to reproduce**

On windows, run `tutor dev start`. The command will fail with the following error:

```
PermissionError: [Errno 13] Permission denied: '/openedx/data/logs/all.log'
```

**Environment**

Windows.

**Additional context**

See discussion here: https://discuss.openedx.org/t/tutor-dev-host-user-id-on-windows-10-causing-containers-to-exit-errno-13-permission-denied-openedx-data-logs-all-log/12328/3

Danyal-Faheem:
Fixes #1034.

Made the change as suggested by @regisb in the issue.

Tested the change on windows by running `tutor dev start` and `tutor local start`. Both seem to be running fine.

### DIFF
--- a/changelog.d/20240604_221238_danyal.faheem_permissions_host_fix_on_windows.md
+++ b/changelog.d/20240604_221238_danyal.faheem_permissions_host_fix_on_windows.md
@@ -1,0 +1,1 @@
+- [Bugfix] Fix permissions error on windows when running `tutor dev start` (by @Danyal-Faheem)
--- a/tutor/templates/dev/docker-compose.yml
+++ b/tutor/templates/dev/docker-compose.yml
@@ -10,7 +10,7 @@ x-openedx-service:
 services:
   permissions:
     environment:
-      OPENEDX_USER_ID: "{{ HOST_USER_ID }}"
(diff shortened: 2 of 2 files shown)


regisb (approved):
looking good!

DawoudSheraz (approved):

PR #1076, opened 2024-06-07
feat: cache edx-platform checkout in image build

regisb:
Similar to what we need with the MFE image, we leverage Docker's ADD directive to implement caching of the edx-platform git checkout step.

See: https://github.com/overhangio/tutor-mfe/pull/163 https://docs.docker.com/reference/dockerfile/#add


tutor/templates/build/openedx/Dockerfile
```diff
@@ -38,9 +38,9 @@ RUN $PYENV_ROOT/versions/$PYTHON_VERSION/bin/python -m venv /openedx/venv
 FROM minimal as code
 ARG EDX_PLATFORM_REPOSITORY={{ EDX_PLATFORM_REPOSITORY }}
 ARG EDX_PLATFORM_VERSION={{ EDX_PLATFORM_VERSION }}
-RUN mkdir -p /openedx/edx-platform && \
-    git clone $EDX_PLATFORM_REPOSITORY --branch $EDX_PLATFORM_VERSION --depth 1 /openedx/edx-platform
+RUN mkdir -p /openedx/edx-platform
 WORKDIR /openedx/edx-platform
+ADD --keep-git-dir=true $EDX_PLATFORM_REPOSITORY#$EDX_PLATFORM_VERSION .
```
DawoudSheraz, line 43: > In the case where <src> is a remote file URL, the destination will have permissions of 600

curios: can this have an impact on the build and files? I am not sure of default permissions of git clone but saw 755 on my system (on host and within docker image) for various repositories.

regisb: Honestly, I'm not sure. We are using this ADD mechanism in the tutor-mfe plugin, and it's working. I suggest that we add this change, and then add the `--chown` option only if necessary.

DawoudSheraz (approved):

PR #1080, opened 2024-06-17
fix: allow setting ACE enabled channels from the env

GlugovGrGlib:
During the work on the new edx-ace channel for [push notifications](https://github.com/openedx/edx-ace/pull/277), we faced the issue with extending `ACE_ENABLED_CHANNELS` via `ENV_PATCHES` with a new channel.
Eventually, we found that the issue is in hardcoded value in tutor common_all template for `ACE_ENABLED_CHANNELS` introduced with the bugfix for development mode - https://github.com/overhangio/tutor/commit/724c2c84dac6652f10f196e0f0f4fc871ad9e3d0.

By taking `ACE_ENABLED_CHANNELS` from `ENV_TOKENS` with a fallback to `["django_email"]` this variable works for development mode as well as allowing override in local mode.

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -172,7 +172,7 @@ SILENCED_SYSTEM_CHECKS = ["2_0.W001", "fields.W903"]
 # Email
 EMAIL_USE_SSL = {{ SMTP_USE_SSL }}
 # Forward all emails from edX's Automated Communication Engine (ACE) to django.
-ACE_ENABLED_CHANNELS = ["django_email"]
+ACE_ENABLED_CHANNELS = ENV_TOKENS.get("ACE_ENABLED_CHANNELS", ["django_email"])
 ACE_CHANNEL_DEFAULT_EMAIL = "django_email"
 ACE_CHANNEL_TRANSACTIONAL_EMAIL = "django_email"
 EMAIL_FILE_PATH = "/tmp/openedx/emails"
--- a/changelog.d/20240618_001743_glib.glugovskiy_allow_setting_ace_channels_from_env.md
+++ b/changelog.d/20240618_001743_glib.glugovskiy_allow_setting_ace_channels_from_env.md
@@ -1,0 +1,1 @@
+- [Bugfix] Allow setting ACE_ENABLED_CHANNELS from the environment, making it possible to easily override it with other ACE channels. (by @GlugovGrGlib)


GlugovGrGlib:
Hi @DawoudSheraz, could you please review this as you have time, thanks!

DawoudSheraz:
Should this be merged into master, so that it can be released with 18.0.1(whenever it happens)?

GlugovGrGlib:
Hi @DawoudSheraz, thanks I have updated PR to point to master

DawoudSheraz (approved):

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -172,7 +172,7 @@
 # Email
 EMAIL_USE_SSL = {{ SMTP_USE_SSL }}
 # Forward all emails from edX's Automated Communication Engine (ACE) to django.
-ACE_ENABLED_CHANNELS = ["django_email"]
+ACE_ENABLED_CHANNELS = ENV_TOKENS.get("ACE_ENABLED_CHANNELS", ["django_email"])
 ACE_CHANNEL_DEFAULT_EMAIL = "django_email"
 ACE_CHANNEL_TRANSACTIONAL_EMAIL = "django_email"
```
regisb, line 177: How come the `ACE_ENABLED_CHANNELS` setting causes an issue in your environment, but not `ACE_CHANNEL_DEFAULT_EMAIL` or `ACE_CHANNEL_TRANSACTIONAL_EMAIL`?

GlugovGrGlib: @regisb We are working on adding a new ACE channel for push notifications (see https://github.com/openedx/edx-ace/pull/277). We need to extend the list of `ACE_ENABLED_CHANNELS` to include this new channel. However, all email settings should remain unchanged as we are expanding the ace_common app with [new settings](https://github.com/openedx/edx-platform/pull/34971/files#diff-1ed4be7d1a4f0267f287987b3214069590f3a3d5a3cccd5af3e4bee8cf4b3addR30) specifically for configuring mobile push notifications.

regisb: In that case, why not use the "openedx-common-settings" patch at the bottom of this settings file to override the `ACE_ENABLED_CHANNELS` setting?

GlugovGrGlib: As this work done as part of extending OeX core, I am expecting that new edx-ace `push notifications` channel will become the default option for many operators, as FCM, the backend that is used for push notification can be used not only for mobile apps but also for other clients including web. And OeX operators would want to enable `push notifications` along with `django_email` channel. Possibility to set those vars from an env will ease configuration, as most operators already patching LMS/CMS env.

regisb: > as most operators already patching the LMS/CMS env.

This is not what I observe. IMHO the lms/cms.env.yml files are weird relics of a past that is long gone. Settings are typically customised using a Tutor plugin, and it's easier to customise the common Python settings than the env.yml files.

GlugovGrGlib: Do you suggest to include the clear instructions for the operators in the ace_common app to override `ACE_ENABLED_CHANNELS` with the "openedx-common-settings" patch, so to eliminate any confusion about setting `ACE_ENABLED_CHANNELS` instead of this PR?

regisb: No, there shouldn't be any tutor-specific instructions in edx-platform apps. Instructions to override the ACE_* settings should suffice. End-users can refer to [this tutorial](ace_common) to learn how to create a tutor plugin.

Soooo I'm suggesting to do nothing at all :)

GlugovGrGlib: @regisb sorry the link is broken, did you mean https://docs.tutor.edly.io/tutorials/edx-platform-settings.html?

regisb: Sorry -- yes, that's correct. And thus the plugin creation tutorial: https://docs.tutor.edly.io/tutorials/plugin.html#plugin-development-tutorial

PR #1083, opened 2024-06-24
feat: set `EDXAPP_TEST_MONGO_HOST=mongodb` in openedx-dev

kdmccormick:
Currently, we are asking devs to set this env var manually before they run tests.

Why not just set it in the development image for them?

### DIFF
--- a/changelog.d/20240624_105125_kyle_test_mongo_host.md
+++ b/changelog.d/20240624_105125_kyle_test_mongo_host.md
@@ -1,0 +1,1 @@
+[Improvement] Set `EDXAPP_TEST_MONGO_HOST` env var in the openedx-dev image so that it no longer needs to be set by hand when running edx-platform unit tests (by @kdmccormick).
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -276,7 +276,8 @@ RUN pip install -e "/mnt/{{ name }}"
 ENV PYTHONBREAKPOINT=ipdb.set_trace
 
 # Point unit tests at the MongoDB container
-ENV EDXAPP_TEST_MONGO_HOST=mongodb
+ENV EDXAPP_TEST_MONGO_HOST={{ MONGODB_HOST }}
+ENV EDXAPP_TEST_MONGO_PORT_NUM={{ MONGODB_PORT }}
 
 # Recompile static assets: in development mode all static assets are stored in edx-platform,
 # and the location of these files is stored in webpack-stats.json. If we don't recompile
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -275,6 +275,9 @@ RUN pip install -e "/mnt/{{ name }}"
 # Add ipdb as default PYTHONBREAKPOINT
 ENV PYTHONBREAKPOINT=ipdb.set_trace
 
+# Point unit tests at the MongoDB container
+ENV EDXAPP_TEST_MONGO_HOST=mongodb
+
 # Recompile static assets: in development mode all static assets are stored in edx-platform,
 # and the location of these files is stored in webpack-stats.json. If we don't recompile
 # static assets, then production assets will be served instead.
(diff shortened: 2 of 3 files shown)


tutor/templates/build/openedx/Dockerfile
```diff
@@ -275,6 +275,9 @@ RUN pip install -e "/mnt/{{ name }}"
 # Add ipdb as default PYTHONBREAKPOINT
 ENV PYTHONBREAKPOINT=ipdb.set_trace
 
+# Point unit tests at the MongoDB container
+ENV EDXAPP_TEST_MONGO_HOST=mongodb
```
regisb, line 279: I think we want to point to `{{ MONGODB_HOST }}`, and not "mongodb", in case MongoDB runs outside of Tutor. Same for `MONGO_PORT_NUM = {{ MONGODB_PORT }}`. But then, we don't want to make the image build depend on user configuration, so we should not store these settings here...

I don't have a better alternative to offer that does not break compatibility with devstack... I *think* that test settings should be overridable, and that xmodule/modulestore/tests/mongo_connection.py should *not* read `EDXAPP_TEST_MONGO_HOST`. But that would break the current behaviour.

kdmccormick: > But then, we don't want to make the image build depend on user configuration, so we should not store these settings here...

I'm not sure I follow. There are several other points in the Dockerfile that depend on user configuration. As long as the setting has a reasonable default value that most people will use (i.e., `mongodb`), it seems fine to me to build that into the image, no?

> I don't have a better alternative to offer that does not break compatibility with devstack... I think that test settings should be overridable, and that xmodule/modulestore/tests/mongo_connection.py should not read EDXAPP_TEST_MONGO_HOST. But that would break the current behaviour.

Agreed, the tests ought to use the MONGO_HOST and MONGO_PORT_NUM Django settings. The way that they currently ignore the settings and look up the env vars directly is so very wrong.

[Devstack is deprecated](https://github.com/openedx/public-engineering/issues/247), and starting right after that Sumac cut, we are no longer obligated to avoid breaking it. Regardless of whether you accept this PR, let's make the upstream fix this October.

DawoudSheraz: Règis has already added his comment/context. Once those are resolved, this will be good to go.

DawoudSheraz: @kdmccormick Hi, what's plan for this PR?

kdmccormick: @regisb 
> I think we want to point to {{ MONGODB_HOST }}, and not "mongodb", in case MongoDB runs outside of Tutor. Same for MONGO_PORT_NUM = {{ MONGODB_PORT }}.

Done 👍🏻 

> I don't have a better alternative to offer that does not break compatibility with devstack... I think that test settings should be overridable, and that xmodule/modulestore/tests/mongo_connection.py should not read EDXAPP_TEST_MONGO_HOST. But that would break the current behaviour.

I poked around at simplifying this. I think that the only proper fix is to drop the EDXAPP_TEST_MONGO_* env vars *and* get rid of the MONGO_* Django settings, as they are all only used in tests. What prod uses is [DOC_STORE_CONFIG](https://github.com/openedx/edx-platform/blob/688242abd9937559e655166e6c7fc6339f4e1acc/cms/envs/common.py#L1051) -- test settings should override pieces of *that*. Unfortunately that's too much for me to bite off right now, but I'm hoping I can address this (and other bad config patterns) in https://github.com/openedx/edx-platform/issues/35146

In the meantime, I still think this simplification would be valuable to Tutor users. Can you re-review @regisb ?

changelog.d/20240624_105125_kyle_test_mongo_host.md
```diff
@@ -0,0 +1 @@
+[Improvement] Set `EDXAPP_TEST_MONGO_HOST` env var in the openedx-dev image so that it no longer needs to be set by hand when running edx-platform unit tests (by @kdmccormick).
```
DawoudSheraz, line 1: do we need a changelog entry for this, considering this is only meant for dev/testing and wont impact the operators as such?

DawoudSheraz (approved):

changelog.d/20240624_105125_kyle_test_mongo_host.md
kdmccormick: Since Tutor is both an operator tool and a developer tool, I think it is good to make not of the improvements we make for developers too.

PR #1085, opened 2024-06-25
chore: ensure plugins are installed from nightly branches in nightly

DawoudSheraz:
This is one of post-redwood release actions in nightly and updates plugins.txt to install plugins from their nightly branches instead of from PyPI.


regisb (approved):
stupid me! thanks for catching this :)

PR #1086, opened 2024-06-29
feat: cleanup flag

### LINKED ISSUE #967: Clean root/env directory before saving config
**Bug description**
Because the env dir is not cleaned when running `tutor config save`, any changes which delete previously rendered files can leave orphaned files in the env and result in undefined / unpredictable behavior. The bigger the upgrade, the more likely that orphaned files will exist and the more unpredictable the behavior is likely to be.

I feel like we've discussed this before and it was counted as desired behavior, but I can't find the discussion or remember why it would be desirable.

**How to reproduce**
Example: when squashing migration files for Aspects between `v0.65.1` and `v0.66.1` we deleted several Alembic migrations. Upon upgrade, each operator needs to delete their env directory (or at least certain subdirectories) and re-run `tutor config save` otherwise Alembic will find the old migration files and throw an error.

1. Install the `tutor-contrib-aspects==v0.65.1` plugin and enable it
2. `tutor config save`
3. List the files in `{tutor root}/env/plugins/aspects/apps/aspects/migrations/alembic/versions/`
4. Upgrade to `tutor-contrib-aspects==v0.66.1`
5. `tutor config save`
6. List the files in `{tutor root}/env/plugins/aspects/apps/aspects/migrations/alembic/versions/` and note that several files which have been deleted from the project still exist, additionally there maybe compiled python files and `__pycache__` etc created there which may be out of date. Running a tutor init in this state will cause errors.
7. Delete `{tutor root}/env/plugins/aspects/apps/aspects/migrations/alembic/versions/` (or just the entire env directory)
8. `tutor config save`
9. List the files in `{tutor root}/env/plugins/aspects/apps/aspects/migrations/alembic/versions/` and note that there are now the correct number of them.

**Environment**
At least Mac OS and Ubuntu 22.04, Tutor 15/16/17 but probably impacts all environments and Tutor versions.

**Additional context**
This has come up as a consistent pain point with front-end development as well, @brian-smith-tcril may have more context about those issues.

CodeWithEmad:
This will introduce `-c` `--clean` command, to make sure we have a new environment on save.

Close #967

### DIFF
--- a/changelog.d/20240629_120748_codewithemad_cleanup_flag.md
+++ b/changelog.d/20240629_120748_codewithemad_cleanup_flag.md
@@ -1,0 +1,1 @@
+- [Feature] Enhance your workflow with the new `-c` or `--clean` option for the `tutor config save` command! This feature allows you to clean your environment before each save, ensuring that all files and directories within the `env/` folder are deleted, providing you with a fresh environment each time. (by @CodeWithEmad)
--- a/changelog.d/20240629_120748_codewithemad_cleanup_flag.md
+++ b/changelog.d/20240629_120748_codewithemad_cleanup_flag.md
@@ -1,1 +1,1 @@
-- [Feature] Enhance your workflow with the new `-c` or `--clean` option for the `tutor config save` command! This feature allows you to clean your environment before each save, ensuring that all files and directories within the `env/` folder are deleted, providing you with a fresh environment each time. (by @CodeWithEmad)
+- [Feature] Added `-c` or `--clean` option to tutor config save: For plugin developers and advanced users, this option cleans the `env/` folder before saving, ensuring a fresh environment for testing and development. (by @CodeWithEmad)
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -191,9 +191,9 @@ def interactive_configuration(
     interactive: bool,
     run_for_prod: t.Optional[bool] = None,
 ) -> None:
-    click.echo(fmt.title("Interactive platform configuration"))
     config = tutor_config.load_minimal(context.obj.root)
     if interactive:
+        click.echo(fmt.title("Interactive platform configuration"))
         interactive_config.ask_questions(
             config,
             context.obj.root,
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -187,12 +187,18 @@ Press enter when you are ready to continue"""
 
 
 def interactive_configuration(
-    context: click.Context, interactive: bool, run_for_prod: t.Optional[bool] = None
+    context: click.Context,
+    interactive: bool,
+    run_for_prod: t.Optional[bool] = None,
 ) -> None:
     click.echo(fmt.title("Interactive platform configuration"))
     config = tutor_config.load_minimal(context.obj.root)
     if interactive:
-        interactive_config.ask_questions(config, run_for_prod=run_for_prod)
+        interactive_config.ask_questions(
+            config,
+            context.obj.root,
+            run_for_prod=run_for_prod,
+        )
     tutor_config.save_config_file(context.obj.root, config)
     config = tutor_config.load_full(context.obj.root)
     tutor_env.save(context.obj.root, config)
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -136,6 +136,13 @@ class ConfigListKeyValParamType(ConfigKeyValParamType):
 @click.option(
     "-e", "--env-only", "env_only", is_flag=True, help="Skip updating config.yml"
 )
+@click.option(
+    "-c",
+    "--clean",
+    "clean_env",
+    is_flag=True,
+    help="Remove everything in the env directory before save",
+)
 @click.pass_obj
 def save(
     context: Context,
@@ -145,10 +152,13 @@ def save(
     remove_vars: list[tuple[str, t.Any]],
     unset_vars: list[str],
     env_only: bool,
+    clean_env: bool,
 ) -> None:
     config = tutor_config.load_minimal(context.root)
     if interactive:
-        interactive_config.ask_questions(config)
+        interactive_config.ask_questions(config, context.root, clean_env_prompt=True)
+    if clean_env:
+        env.delete_env_dir(context.root)
     if set_vars:
         for key, value in set_vars:
             config[key] = env.render_unknown(config, value)
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -228,7 +228,7 @@ def launch(context: click.Context, non_interactive: bool) -> None:
     click.echo(fmt.title("Interactive platform configuration"))
     config = tutor_config.load_minimal(context.obj.root)
     if not non_interactive:
-        interactive_config.ask_questions(config, run_for_prod=True)
+        interactive_config.ask_questions(config, context.obj.root, run_for_prod=True)
     tutor_config.save_config_file(context.obj.root, config)
     config = tutor_config.load_full(context.obj.root)
     tutor_env.save(context.obj.root, config)
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -225,9 +225,9 @@ def launch(context: click.Context, non_interactive: bool) -> None:
             from_release=tutor_env.get_env_release(context.obj.root),
         )
 
-    click.echo(fmt.title("Interactive platform configuration"))
     config = tutor_config.load_minimal(context.obj.root)
     if not non_interactive:
+        click.echo(fmt.title("Interactive platform configuration"))
         interactive_config.ask_questions(config, context.obj.root, run_for_prod=True)
     tutor_config.save_config_file(context.obj.root, config)
     config = tutor_config.load_full(context.obj.root)
--- a/tutor/interactive.py
+++ b/tutor/interactive.py
@@ -7,7 +7,12 @@ from . import env, exceptions, fmt, hooks
 from .types import Config, get_typed
 
 
-def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
+def ask_questions(
+    config: Config,
+    root: str,
+    run_for_prod: Optional[bool] = None,
+    clean_env_prompt: bool = False,
+) -> None:
     """
     Interactively ask questions to collect configuration values from the user.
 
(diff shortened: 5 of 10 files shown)


kdmccormick:
Thanks @CodeWithEmad ! I can take a look in a couple weeks after the conference and my vacation.

CodeWithEmad:
Thank you, @kdmccormick. There's no rush for it. Enjoy your vacation.

tutor/commands/config.py
```diff
@@ -145,10 +152,13 @@ def save(
     remove_vars: list[tuple[str, t.Any]],
     unset_vars: list[str],
     env_only: bool,
+    clean_env: bool,
 ) -> None:
     config = tutor_config.load_minimal(context.root)
     if interactive:
         interactive_config.ask_questions(config)
+    if clean_env:
```
DawoudSheraz, line 160: Point 4 https://github.com/overhangio/tutor/issues/967#issuecomment-1903535377 does not seem to be there.

CodeWithEmad: You're right. I forgot about it.

CodeWithEmad: I'm not sure if this is the best way to implement it in interactive mode, but I'm open to any suggestions.

tutor/interactive.py
```diff
@@ -148,6 +150,16 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
             config,
             defaults,
         )
+    run_clean = click.confirm(
```
kdmccormick, line 153: This will make it so that the user is always prompted to start fresh, whether or not `-c` was provided. I do not think we want to ask the user this every they launch interactive config. I think it would encourage people to delete their environment too often out of caution, which isn't the indented regular Tutor workflow.

Can you change it so that they are only prompted if `-c` is provided? To make the implementation easier, you can simply abort if they say no. The prompt could be something like "The -c flag you provide will delete your existing Tutor environment. Are you sure you want to do this?"

kdmccormick (changes requested):

tutor/interactive.py
CodeWithEmad: Totally agree. TBH, I really didn't know what exactly we wanted in interactive mode.

tutor/interactive.py
```diff
@@ -148,6 +157,17 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
...
             defaults,
         )
+    if clean_environment:
+        run_clean = click.confirm(
+            fmt.question(
+                "The -c flag you provide will delete your existing Tutor environment. "
+                "Are you sure you want to do this?"
+            ),
+            prompt_suffix=" ",
+            default=False,
```
kdmccormick, line 167: I would default True here, just because the user is already going out of their way to provide the `-c` flag.

tutor/commands/k8s.py
```diff
@@ -225,10 +234,16 @@ def launch(context: click.Context, non_interactive: bool) -> None:
             from_release=tutor_env.get_env_release(context.obj.root),
         )
 
-    click.echo(fmt.title("Interactive platform configuration"))
     config = tutor_config.load_minimal(context.obj.root)
     if not non_interactive:
-        interactive_config.ask_questions(config, run_for_prod=True)
+        click.echo(fmt.title("Interactive platform configuration"))
```
kdmccormick, line 239: Nice fix, moving this this log line into the if.

tutor/commands/config.py
```diff
@@ -145,10 +152,13 @@ def save(
     remove_vars: list[tuple[str, t.Any]],
     unset_vars: list[str],
     env_only: bool,
+    clean_env: bool,
 ) -> None:
     config = tutor_config.load_minimal(context.root)
     if interactive:
-        interactive_config.ask_questions(config)
+        interactive_config.ask_questions(config, context.root)
```
kdmccormick, line 159: When I tested this, `tutor dev clean -ci` didn't ask me if I wanted to remove my env. I think you need to pass the `clean_env` through here.

kdmccormick (changes requested):

tutor/interactive.py
CodeWithEmad: You're absolutely right. changed to True.

tutor/commands/compose.py
```diff
@@ -187,12 +200,24 @@ def interactive_upgrade(
...
     if interactive:
-        interactive_config.ask_questions(config, run_for_prod=run_for_prod)
+        click.echo(fmt.title("Interactive platform configuration"))
+        interactive_config.ask_questions(
+            config,
+            context.obj.root,
+            run_for_prod=run_for_prod,
+            clean_environment=clean_environment,
+        )
+    else:
```
regisb, line 217: I don't understand why this is needed here. Environment deletion will be handled by the save command, right?

tutor/commands/k8s.py
```diff
@@ -225,10 +234,16 @@ def launch(context: click.Context, non_interactive: bool) -> None:
...
+        interactive_config.ask_questions(
+            config, context.obj.root, run_for_prod=True, clean_environment=clean_env
+        )
+    else:
```
regisb, line 243: I'm not 100% sure we want to support a `launch --clean` option. What do ya'll think?

tutor/commands/compose.py
CodeWithEmad: To have a launch -- clean command

CodeWithEmad: Yes, but this will add the clean flag to launch too

tutor/commands/k8s.py
CodeWithEmad: I'm okay with removing this but it's the shorthand for:

`tutor config save --clean && tutor dev|local|k8s launch`

tutor/interactive.py
```diff
@@ -148,6 +157,17 @@ def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
             config,
             defaults,
         )
+    if clean_environment:
+        run_clean = click.confirm(
+            fmt.question(
+                "The -c flag you provide will delete your existing Tutor environment. "
+                "Are you sure you want to do this?"
```
DawoudSheraz, line 164: nits:
- The -c flag deletes and recreates existing Tutor environment {env_path}
- Are you sure you want to proceed with environment deletion

The sentence `Are you sure you want to do this?` looked confusing when reading it in interactive mode. It gave the impression that we want to stop the entire operation, not just cleanup.

CodeWithEmad: Well, since it's a risky function and user might remove its files, we should kinda warn them about the importance of their decision, don't we? @kdmccormick asked to change the original to this, [here](https://github.com/overhangio/tutor/pull/1086#discussion_r1681744804). @regisb do you have something about the clean-up message?

tutor/commands/k8s.py
regisb: We're adding a simple boolean command flag and it ends up affecting 6 different modules and functions. The fact that the `clean` logic has to be handled both by `launch` and `save` makes the feature too complex, IMHO. Since this option is not recommended by default, I would suggest to add it only to the `config save` command, which will make the code much simpler.

tutor/interactive.py
kdmccormick: Right, the pause for confirmation is coming directly from [Regis's original formulation of the `-c` flag](https://github.com/overhangio/tutor/issues/967#issuecomment-1903535377). So it is good to have the prompt, but I agree with @DawoudSheraz that the current version does kinda make it look like the entire operation will be cancelled if you say No:

image

I recommend simplifying the message to just: `Remove existing Tutor environment directory? [Y/n] `

tutor/commands/k8s.py
kdmccormick: Regis makes good points, I am fine with the `launch -c` flag being removed.

CodeWithEmad: OK. I removed the launch flag.

tutor/commands/compose.py
CodeWithEmad: launch flag removed.

tutor/interactive.py
CodeWithEmad: Sure, I changed the message.

changelog.d/20240629_120748_codewithemad_cleanup_flag.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] Enhance your workflow with the new `-c` or `--clean` option for the `tutor config save` command! This feature allows you to clean your environment before each save, ensuring that all files and directories within the `env/` folder are deleted, providing you with a fresh environment each time. (by @CodeWithEmad)
```
DawoudSheraz, line 1: It's not a breaking change. The feature is only accessible behind a new argument. we can remove 💥 from the changelog.

tutor/commands/compose.py
```diff
@@ -84,10 +84,7 @@ def job_runner(self, config: Config) -> ComposeTaskRunner:
 @click.option("--skip-build", is_flag=True, help="Skip building Docker images")
 @click.pass_context
 def launch(
-    context: click.Context,
-    non_interactive: bool,
-    pullimages: bool,
-    skip_build: bool,
+    context: click.Context, non_interactive: bool, pullimages: bool, skip_build: bool
```
DawoudSheraz, line 87: this change is opposite to the change you made on line 187. It would be good if we can maintain consistency where possible.

tutor/interactive.py
kdmccormick: Thanks!

changelog.d/20240629_120748_codewithemad_cleanup_flag.md
kdmccormick: +1. Also, instead of calling on all users to enhance their workflow with this new flag, could you explain why a user would want to use it? Reading https://github.com/overhangio/tutor/issues/967, my understanding is this most users will not need to use `-c`; rather, only plugin developers or other advanced users will need it, as they are the ones who will need to clean files out of their environment.

CodeWithEmad: > It's not a breaking change.

Oops. My bad.

> could you explain why a user would want to use it?

How about this?
- [Feature] Added `-c` or `--clean` option to `tutor config save`: For plugin developers and advanced users, this option cleans the `env/` folder before saving, ensuring a fresh environment for testing and development.

tutor/commands/compose.py
CodeWithEmad: Nice catch.

DawoudSheraz (approved):

changelog.d/20240629_120748_codewithemad_cleanup_flag.md
kdmccormick: Perfect 🚀

CodeWithEmad: Done

kdmccormick:
Thanks @CodeWithEmad !!

PR #1088, opened 2024-07-01
feat: use uv as a pip alternative

MoisesGSalas:
[`uv`](https://github.com/astral-sh/uv) is a new package installer/resolver written in rust by the same people that wrote [`ruff`](https://github.com/astral-sh/ruff).

This is a small test to assess the viability of using `uv` instead of pip. It presents it self as a drop in replacement for pip, and for the most part it is, I only had to apply a [single change](https://github.com/eduNEXT/edx-platform/commit/5bc067474ce557fe57aed29174d7f81873be5cf8) due to limitations in editable VCS packages.

## Tests

Using `--target=python-requirements`:

**uv**
1. An initial build without cache (`--no-cache`): `5m08s`, `97s` on pip install edx.txt
2. Second build without cache: `4m58s`, `94s` on pip install edx.txt
3. Break cache with a manual instructions before installing edx.txt: `1s` on edx.txt

**pip**
1. An initial build without cache (`--no-cache`): _
2. Second build without cache: _
3. Break cache with a manual instructions before installing edx.txt: `220s` one edx.txt

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -9,6 +9,9 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     apt update && \
     apt install -y build-essential curl git language-pack-en
 ENV LC_ALL=en_US.UTF-8
+RUN curl -L https://github.com/astral-sh/uv/releases/download/0.2.13/uv-x86_64-unknown-linux-gnu.tar.gz | \
+    tar --gunzip --directory=/usr/local/bin --strip-components=1 --extract uv-x86_64-unknown-linux-gnu/uv
+
 {{ patch("openedx-dockerfile-minimal") }}
 
 ###### Install python with pyenv in /opt/pyenv and create virtualenv in /openedx/venv
@@ -82,20 +85,22 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 
 # Install the right version of pip/setuptools
 RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
-    pip install \
+    uv pip install \
     # https://pypi.org/project/setuptools/
     # https://pypi.org/project/pip/
     # https://pypi.org/project/wheel/
-    setuptools==69.1.1 pip==24.0 wheel==0.43.0
+    setuptools==69.1.1 wheel==0.43.0
 
 # Install base requirements
 RUN --mount=type=bind,from=edx-platform,source=/requirements/edx/base.txt,target=/openedx/edx-platform/requirements/edx/base.txt \
     --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
-    pip install -r /openedx/edx-platform/requirements/edx/base.txt
+    --mount=type=cache,target=/openedx/.cache/uv,sharing=shared \
+    uv pip install -r /openedx/edx-platform/requirements/edx/base.txt
 
 # Install extra requirements
 RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
-    pip install \
+    --mount=type=cache,target=/openedx/.cache/uv,sharing=shared \
+    uv pip install \
     # Use redis as a django cache https://pypi.org/project/django-redis/
     django-redis==5.4.0 \
     # uwsgi server https://pypi.org/project/uWSGI/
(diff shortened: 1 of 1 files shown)


regisb:
Hey Moises, these figures are getting everyone excited :) Do you think you'll be able to take this PR out of draft state? Or do you need some review?

MoisesGSalas:
Hi @regisb, sorry for leaving this abandoned. I can work on this, but first I wanted to know people's opinion about the tool.

IMO the tool is pretty awesome and slots in almost perfectly, the same applies to the other tool developed by the same team `ruff`. But considering it still has not reached v1.0.0 I wonder if some people may be worried about that.

If the overall feeling is that it's worth a shot, I can clean up the PR and have it ready for review.

regisb:
Yes, ruff is awesome, to the point that I'm considering migrating the entire code base from pylint/black to ruff. But let's talk about this another day.

DawoudSheraz:
@MoisesGSalas What are the plans for this? We can target this for nightly so that it becomes part of sumac cutoff next week.

MoisesGSalas:
@DawoudSheraz, @regisb, sorry I couldn't find the time to retake this.

I think is ready for review. I wasn't quite sure what mechanism to use to fallback to the standard `pip` command if needed so I ended up using a build argument because it seemed to be the simpler one.

tutor/templates/build/openedx/Dockerfile
```diff
@@ -51,8 +55,10 @@ RUN git config --global user.email "tutor@overhang.io" \
 {{ patch("openedx-dockerfile-git-patches-default") }}
 {%- elif EDX_PLATFORM_VERSION == "master" %}
 # Patches in nightly node
+RUN curl -fsSL https://github.com/openedx/edx-platform/commit/3642cab3ac61ddfd360e7ceb7463b52be8c4deb0.patch | git am
```
DawoudSheraz, line 58: why is this patch needed?

MoisesGSalas: [edx-proctoring-proctortrack is installed as an editable VCS dependency](https://github.com/openedx/edx-platform/blob/ecbedbcf78cd74c907f18d76b77a64e31ccf78fe/requirements/edx/github.in#L92) but uv [doesn't support that](https://github.com/astral-sh/uv/issues/5442). I have this PR https://github.com/openedx/edx-platform/pull/35654 with the same change that is used in that patch.

regisb: The fact that uv doesn't support editable dependencies is a blocking issue, I think. I suggest we wait until the [upstream issue](https://github.com/astral-sh/uv/issues/5442) is resolved.

MoisesGSalas: I think they are not going to implement it in the near future. Note that it refers to editable remote requirements, it works fine with editable requirements in a local path.

AFAIK the only use for an editable remote requirement is that it copies all the files, but you can fix that using the `MANIFEST.in`.

regisb: I understand that. That means that using `uv` means we will no longer be able to support adding editable dependencies to edx-platform. I'm not sure I'm ready to make that breaking change, as I expect that many people (Axim included) might add editable dependencies there.

Can we consider making an upstream PR to support that feature?

regisb: > Maybe we can keep standard pip as the default with the possibility to swap to uv if your fork of edx-platform doesn't contain any editable dependency.

That's already the case. People can install uv in their own Docker image and alias `pip='uv pip'`, all with a Tutor plugin.

regisb: > But that wouldn't work for the RUN commands in the Dockerfile, right?

It would. Install uv with the "openedx-dockerfile-minimal" patch, then `ln -s /usr/bin/uv /usr/bin/pip`. Maybe this will be overwritten in the python-requirements layer, in which case we would have to introduce a new patch statement just above `pip install -r /openedx/edx-platform/requirements/edx/base.txt`. Would you like to introduce such a patch?

regisb:
I'd like to revisit this PR, as I think that uv would be a great addition to Tutor. Do you think you could try to remove editable git dependencies from edx-platform? As far as I know, we would only need to change the edx-proctoring-proctortrack requirement, and install it as non-editable, which should not be a breaking change.

MoisesGSalas:
Hi @regisb, yes I will try to see if we can remove it from the requirements (it's also the only git requirement in edxapp, which is kinda annoying).

But it would still be a breaking change for people with custom requirements using editable remote dependencies right?

regisb:
> But it would still be a breaking change for people with custom requirements using editable remote dependencies right?

Yes, that's correct. But I'm not sure anymore if that would be a big deal. To be honest, I'm not sure what the use of the `-e` option in a `pip install git+https://...` command.

MoisesGSalas:
>  I'm not sure what the use of the -e option in a pip install git+https://... command.

It's only useful if you don' have proper packaging set in place and want to grab all the files (e.g. you need non .py files and you have not written the MANIFEST.in file).

regisb:
OK let's consider that this is just a fringe edge case then. We can migrate to uv in the main branch, and we'll see whether people complain. Would you like to open a PR upstream in edx-platform to remove the editable option in `-e git+https://github.com/anupdhabarde/edx-proctoring-proctortrack.git`?

regisb:
Looks like @kdmccormick beat you to it :wink: https://github.com/openedx/edx-platform/issues/36329

If edx-proctoring-proctortrack becomes deprecated, it removes the biggest hurdle in the path of this PR.

PR #1090, opened 2024-07-03
fix: add turn on mysql-native-password authentication plugin to fix broken connection for users upgrading from Tutor 15 or earlier

### LINKED ISSUE #1089: Upgrading from tutor 15 to tutor 18 breaks the MySQL connection due to the mysql_native_password authentication plugin not being found
**Bug description**

When you upgrade from tutor 15.x to tutor 18.x, the LMS cannot connect to the MySQL server during the launch step with the error:
```
ERROR 1524 (HY000): Plugin 'mysql_native_password' is not loaded
```
The same error occurs if you try to enter the SQLshell using the following methods:

- `do sqlshell` command
- Bashing into the MySQL container
- Running `./manage.py lms dbshell` in the LMS container

**How to reproduce**

1. Install tutor 15.3.9.
2. Run `tutor local launch` to launch a fresh local instance and verify that the LMS and Studio are working fine.
3. Install tutor 16.x.
4. Run `tutor config save`
5. Run `tutor local upgrade --from=olive`
6. Run `tutor local launch`
7. Follow step 3-6 for tutor 17 and tutor 18 replacing the version and release name.

And then the init task for tutor 18 crashes at the MySQL step with the above mentioned error. 

**Environment**
OS: MacOS Sonoma 14.5
Tutor versions 15.3.9 and 18.1.0

**Additional context**

This issue was initially raised in https://discuss.openedx.org/t/error-1524-hy000-plugin-mysql-native-password-is-not-loaded/13257/9

The [default](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password) authentication plugin in MySQL 8.0 and onwards has been changed to `caching_sha2_password` whereas it was `mysql_native_password` in MySQL 5.7 and earlier. If a `CREATE USER` query did not specify the authentication plugin explicitly, the default authentication plugin was used.

According to the MySQL [docs](https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/):
```
User accounts already created in existing instances are not altered as a part of an upgrade. So, all existing user accounts continue to use the same authentication plugin that they used in e.g. MySQL 5.7.
```
This means that the users that were created by Tutor in version 15 or earlier used the `mysql_native_password` authentication and  even after upgrading to MySQL 8.0 in Palm and 8.4 in Redwood, that was not changed. 

In Tutor 18.0.0, MySQL was upgraded to v8.4.0 as per #1071. In MySQL 8.4.0, `mysql_native_password` is disabled by default as per the [docs](https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html#:~:text=In%20MySQL%208.4%2C%20the%20mysql_native_password,purpose%2C%20but%20is%20not%20required.). This means that if the DB contains users using the `mysql_native_password` plugin, it will refuse to connect.

A possible solution is to start the MySQL server with the parameters `--mysql-native-password=ON`. But this might be unsafe in the long run.

Danyal-Faheem:
fixes #1089.

For tutor instances that were created with tutor 15 and earlier and then upgraded to tutor 18, the launch process would fail as the MySQL connection could not be made. 

This is because mysql-native-password was removed in MySQL 8.4.0. We turn it on temporarily to fix failing connections.

### DIFF
--- a/tutor/templates/local/docker-compose.yml
+++ b/tutor/templates/local/docker-compose.yml
@@ -44,6 +44,7 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
+      --mysql-native-password=ON
     restart: unless-stopped
     user: "999:999"
     volumes:
--- a/changelog.d/20240704_112127_danyal.faheem_fix_mysql_native_password_plugin.md
+++ b/changelog.d/20240704_112127_danyal.faheem_fix_mysql_native_password_plugin.md
@@ -1,0 +1,1 @@
+- [Bugfix] Fix `mysql-native-password is not loaded` error in MySQL 8.4 when upgrading from tutor 15 or an earlier release to tutor 18 by enabling the plugin. (by @Danyal-Faheem)
--- a/tutor/templates/k8s/deployments.yml
+++ b/tutor/templates/k8s/deployments.yml
@@ -397,6 +397,7 @@ spec:
             - "--character-set-server=utf8mb4"
             - "--collation-server=utf8mb4_unicode_ci"
             - "--binlog-expire-logs-seconds=259200"
+            - "--mysql-native-password=ON"
           env:
             - name: MYSQL_ROOT_PASSWORD
               value: "{{ MYSQL_ROOT_PASSWORD }}"


tutor/templates/local/docker-compose.yml
```diff
@@ -44,6 +44,7 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
+      --mysql-native-password=ON
```
DawoudSheraz, line 47: shouldn't this be done for k8s as well?

Danyal-Faheem: Updated. Thank you for pointing it out.

DawoudSheraz (approved):

DawoudSheraz (commented):
please add changelog entry

Danyal-Faheem:
> please add changelog entry

Updated.

PR #1079, opened 2024-06-14
feat: add do command to change MySQL charset to utf8mb4

Danyal-Faheem:
WIP.

PR #1097, opened 2024-07-18
feat: add `do` command to update the authentication plugin of MySQL users

### LINKED ISSUE #1095: Add instructions/do command to switch mysql user auth to caching_sha2_password
MySQL was upgraded to v8.4.0 in redwood (https://github.com/overhangio/tutor/pull/1071). The upgrade broke Tutor for users upgrading from Tutor 15 or earlier. MySQL 8.4 removes mysql-native-password (https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html) because the option has been deprecated. If any user plugin is authenticating via native-password, MySQL server needs to provide [--mysql-native-password=ON](https://dev.mysql.com/doc/refman/8.4/en/server-options.html#option_mysqld_mysql-native-password) as an argument.

A quick fix was added at https://github.com/overhangio/tutor/pull/1090. However, when MySQL is upgraded from 8.4, the flows will break because native-password will no longer exist. Therefore, we need to either document how to move a user from native-password to caching_sha2_password or provide a capability in tutor to do this. 

- https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html
- https://discuss.openedx.org/t/error-1524-hy000-plugin-mysql-native-password-is-not-loaded/13257

Danyal-Faheem:
closes #1095

This PR is a continuation of the issue mentioned in #1089. 

- Add a do command to update the authentication plugin of MySQL users from mysql_native_password to caching_sha2_password.
- The command requires a username and password to function. The password can be provided as a CLI option with the `--password` option or interactively. The command can also be run non-interactively with the `--non-interactive` option as it might prompt for some confirmations otherwise. 

The commands necessary to update a native installation of tutor would be:

```
tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME) --password=$(tutor config printvalue OPENEDX_MYSQL_PASSWORD) --non-interactive
tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME) --password=$(tutor config printvalue MYSQL_ROOT_PASSWORD) --non-interactive

```

**Note:** Any passwords entered here will be considered as the new password for that user as new hashes are being generated.

### DIFF
--- a/changelog.d/20240718_171945_danyal.faheem_mysql_authentication_plugin_change.md
+++ b/changelog.d/20240718_171945_danyal.faheem_mysql_authentication_plugin_change.md
@@ -1,0 +1,1 @@
+- [Improvement] Add a do command to update the authentication plugin of existing MySQL users from mysql_native_password to caching_sha2_password for compatibility with MySQL v8.4.0 and above. (by @Danyal-Faheem)
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -183,6 +183,8 @@ The password will not be required for official plugins that have database users 
 
     tutor local do update-mysql-authentication-plugin USERNAME --password=PASSWORD
 
+.. warning:: Since we are generating a new password hash, whatever password is entered here will be considered as the new password for the user. Please make similar changes to any connection strings to avoid database connection issues.
+
 To update the database users for a vanilla tutor installation::
 
     tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME)
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -168,6 +168,26 @@ By default, only the tables in the openedx database are changed. For upgrading t
 
     tutor local do convert-mysql-utf8mb4-charset --database=discovery
 
+.. _update_mysql_authentication_plugin:
+
+Updating the authentication plugin of MySQL users
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As of MySQL v8.4.0, the ``mysql_native_password`` authentication plugin has been deprecated. Users created with this authentication plugin should ideally be updated to use the latest ``caching_sha2_password`` authentication plugin.
+
+Tutor makes it easy do so with this handy command::
+
+    tutor local do update-mysql-authentication-plugin all
+
+The above command will update all the database users created by Tutor. If you only want to update the authentication plugin of specific users, you can use the ``--users`` option. This option takes comma seperated names of users to upgrade::
+
+    tutor local do update-mysql-authentication-plugin discovery ecommerce
+
+For this command, Tutor expects specific entries in the configuration for the mysql username and password of a database user. For example, if you are trying to update the user ``myuser``, the following case sensitive entries need to be present in the configuration::
+
+    MYUSER_MYSQL_USERNAME
+    MYUSER_MYSQL_PASSWORD
+
 Running arbitrary ``manage.py`` commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -177,17 +177,15 @@ As of MySQL v8.4.0, the ``mysql_native_password`` authentication plugin has been
 
 Tutor makes it easy do so with this handy command::
 
-    tutor local do update-mysql-authentication-plugin all
+    tutor local do update-mysql-authentication-plugin myuser
 
-The above command will update all the database users created by Tutor. If you only want to update the authentication plugin of specific users, you can use the ``--users`` option. This option takes comma seperated names of users to upgrade::
+The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command::
 
-    tutor local do update-mysql-authentication-plugin discovery ecommerce
+    tutor local do update-mysql-authentication-plugin myuser --password=mypassword
 
-For this command, Tutor expects specific entries in the configuration for the mysql username and password of a database user. For example, if you are trying to update the user ``myuser``, the following case sensitive entries need to be present in the configuration::
+Tutor may prompt you with some warnings if the entered password is suspected to be wrong. To avoid these prompts, use the non-interactive option::
 
-    MYUSER_MYSQL_USERNAME
-    MYUSER_MYSQL_PASSWORD
-
+    tutor local do update-mysql-authentication-plugin myuser --password=mypassword --non-interactive
 Running arbitrary ``manage.py`` commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -179,13 +179,15 @@ Tutor makes it easy do so with this handy command::
 
     tutor local do update-mysql-authentication-plugin USERNAME
 
-The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command. To update the openedx mysql user::
+The password will not be required for official plugins that have database users as tutor can infer it from the config. If the password cannot be found by tutor, you will be prompted to enter the password interactively. Alternatively, the password can also be provided as an option::
 
-    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME) --password=$(tutor config printvalue OPENEDX_MYSQL_PASSWORD)
+    tutor local do update-mysql-authentication-plugin USERNAME --password=PASSWORD
 
-Tutor may prompt you with some warnings if the entered password is suspected to be wrong. To avoid these prompts, the non-interactive option should be used. To update the root mysql user non-interactively::
+To update the database users for a vanilla tutor installation::
 
-    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME) --password=$(tutor config printvalue MYSQL_ROOT_PASSWORD) --non-interactive
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME)
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME)
+
 
 Running arbitrary ``manage.py`` commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/local.rst
+++ b/docs/local.rst
@@ -177,15 +177,16 @@ As of MySQL v8.4.0, the ``mysql_native_password`` authentication plugin has been
 
 Tutor makes it easy do so with this handy command::
 
-    tutor local do update-mysql-authentication-plugin myuser
+    tutor local do update-mysql-authentication-plugin USERNAME
 
-The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command::
+The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command. To update the openedx mysql user::
 
-    tutor local do update-mysql-authentication-plugin myuser --password=mypassword
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME) --password=$(tutor config printvalue OPENEDX_MYSQL_PASSWORD)
 
-Tutor may prompt you with some warnings if the entered password is suspected to be wrong. To avoid these prompts, use the non-interactive option::
+Tutor may prompt you with some warnings if the entered password is suspected to be wrong. To avoid these prompts, the non-interactive option should be used. To update the root mysql user non-interactively::
 
-    tutor local do update-mysql-authentication-plugin myuser --password=mypassword --non-interactive
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME) --password=$(tutor config printvalue MYSQL_ROOT_PASSWORD) --non-interactive
+
 Running arbitrary ``manage.py`` commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -222,4 +222,7 @@ The detailed steps are mentioned in `tutor-mfe <https://github.com/overhangio/tu
 
 This issue can occur when Tutor is upgraded from v15 (Olive) or earlier to v18 (Redwood) because the users created in Tutor v15 utilize the mysql_native_password authentication plugin by default. This plugin has been deprecated as of MySQL v8.4.0 which is the default MySQL server used in Tutor v18.
 
-The handy :ref:`update-mysql-authentication-plugin <update_mysql_authentication_plugin>` do command in tutor can be used to fix this issue.
+The handy :ref:`update-mysql-authentication-plugin <update_mysql_authentication_plugin>` do command in tutor can be used to fix this issue. To update the database users for a vanilla tutor installation::
+
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME) --password=$(tutor config printvalue OPENEDX_MYSQL_PASSWORD) --non-interactive
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME) --password=$(tutor config printvalue MYSQL_ROOT_PASSWORD) --non-interactive
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -216,3 +216,10 @@ NPM Dependency Conflict When overriding ``@edx/frontend-component-header`` or ``
 ----------------------------------------------------------------------------------------------------------------
 
 The detailed steps are mentioned in `tutor-mfe <https://github.com/overhangio/tutor-mfe?tab=readme-ov-file#npm-dependency-conflict-when-overriding-edxfrontend-component-header-or-edxfrontend-component-footer>`__ documentation.
+
+"Plugin 'mysql_native_password' is not loaded"
+----------------------------------------------
+
+This issue can occur when Tutor is upgraded from v15 (Olive) or earlier to v18 (Redwood) because the users created in Tutor v15 utilize the mysql_native_password authentication plugin by default. This plugin has been deprecated as of MySQL v8.4.0 which is the default MySQL server used in Tutor v18.
+
+The handy :ref:`update-mysql-authentication-plugin <update_mysql_authentication_plugin>` do command in tutor can be used to fix this issue.
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -220,9 +220,9 @@ The detailed steps are mentioned in `tutor-mfe <https://github.com/overhangio/tu
 "Plugin 'mysql_native_password' is not loaded"
 ----------------------------------------------
 
-This issue can occur when Tutor is upgraded from v15 (Olive) or earlier to v18 (Redwood) because the users created in Tutor v15 utilize the mysql_native_password authentication plugin by default. This plugin has been deprecated as of MySQL v8.4.0 which is the default MySQL server used in Tutor v18.
+This issue can occur when Tutor is upgraded from v15 (Olive) or earlier to v18 (Redwood) or later because the users created in Tutor v15 and earlier utilize the mysql_native_password authentication plugin by default. This plugin has been deprecated as of MySQL v8.4.0 which is the default MySQL server used in Tutor v18 and onwards.
 
 The handy :ref:`update-mysql-authentication-plugin <update_mysql_authentication_plugin>` do command in tutor can be used to fix this issue. To update the database users for a vanilla tutor installation::
 
-    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME) --password=$(tutor config printvalue OPENEDX_MYSQL_PASSWORD) --non-interactive
-    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME) --password=$(tutor config printvalue MYSQL_ROOT_PASSWORD) --non-interactive
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME)
+    tutor local do update-mysql-authentication-plugin $(tutor config printvalue MYSQL_ROOT_USERNAME)
--- a/tutor/commands/jobs.py
+++ b/tutor/commands/jobs.py
@@ -20,7 +20,6 @@ from tutor.commands.jobs_utils import (
     set_theme_template,
 )
 from tutor.hooks import priorities
-from tutor.types import get_typed
 
 
 class DoGroup(click.Group):
@@ -387,17 +386,14 @@ def convert_mysql_utf8mb4_charset(
 @click.option(
     "-p",
     "--password",
-    help="Specify password from the command line. If undefined, you will be prompted to input a password",
-    prompt=True,
-    hide_input=True,
+    help="Specify password from the command line.",
 )
 @click.argument(
     "user",
 )
-@click.option("-I", "--non-interactive", is_flag=True, help="Run non-interactively")
 @click.pass_obj
 def update_mysql_authentication_plugin(
-    context: Context, user: str, password: str, non_interactive: bool
+    context: Context, user: str, password: str
 ) -> t.Iterable[tuple[str, str]]:
     """
     Update the authentication plugin of MySQL users from mysql_native_password to caching_sha2_password
@@ -413,30 +409,40 @@ def update_mysql_authentication_plugin(
         )
         return
 
-    conventional_password_key = f"{user.upper()}_MYSQL_PASSWORD"
+    # Official plugins that have their own mysql user
+    known_plugins_with_mysql_keys = [
+        "credentials",
+        "discovery",
+        "ecommerce",
+        "jupyter",
+        "notes",
+        "openedx",
+        "xqueue",
+    ]
 
-    # Prompt for confirmation to move forward if password not present in config with the conventional format USER_MYSQL_PASSWORD
-    if not non_interactive and not conventional_password_key in config:
-        if not click.confirm(
-            fmt.question(
-                f"""Password for user {user} could not be verified. The entered password is: {password}
-Would you still like to continue with the upgrade process? Note: a wrong password would update the password for the user."""
-            )
-        ):
-            return
-    # Prompt for confirmation to move forward is password is present in config with the conventional format USER_MYSQL_PASSWORD
-    # but it is not the same as the value of that config variable
-    elif (
-        not non_interactive
-        and get_typed(config, conventional_password_key, str, "") != password
-    ):
-        if not click.confirm(
-            fmt.question(
-                f"""Password for user {user} is suspected to be wrong. The entered password is: {password} while the password suspected to be the correct one is {config[conventional_password_key]}
-Would you still like to continue with the upgrade process? Note: a wrong password would update the password for the user."""
-            )
-        ):
-            return
+    # Create a list of the usernames and passwords
+    known_mysql_credentials_keys = [
+        (f"{plugin.upper()}_MYSQL_USERNAME", f"{plugin.upper()}_MYSQL_PASSWORD")
+        for plugin in known_plugins_with_mysql_keys
+    ]
+    # Add the root user as it is the only one that is different from the rest
+    known_mysql_credentials_keys.append(("MYSQL_ROOT_USERNAME", "MYSQL_ROOT_PASSWORD"))
+
+    known_mysql_credentials = {}
+    # Build the dictionary of known credentials
+    for k, v in known_mysql_credentials_keys:
+        if username := config.get(k):
+            known_mysql_credentials[username] = config[v]
+
+    if not password:
+        password = known_mysql_credentials.get(user)  # type: ignore
+
+    # Prompt the user if password was not found in config
+    if not password:
+        password = click.prompt(
+            f"Please enter the password for the user {user}",
+            type=str,
+        )
 
     host = "%"
 
--- a/tutor/commands/jobs.py
+++ b/tutor/commands/jobs.py
@@ -12,13 +12,15 @@ import click
 from typing_extensions import ParamSpec
 
 from tutor import config as tutor_config
-from tutor import env, fmt, hooks, plugins
+from tutor import env, fmt, hooks
 from tutor.commands.context import Context
 from tutor.commands.jobs_utils import (
-    get_mysql_change_authentication_plugin_query,
+    create_user_template,
     get_mysql_change_charset_query,
+    set_theme_template,
 )
 from tutor.hooks import priorities
+from tutor.types import get_typed
 
 
 class DoGroup(click.Group):
@@ -110,24 +112,6 @@ def createuser(
     Password can be passed as an option or will be set interactively.
     """
     yield ("lms", create_user_template(superuser, staff, name, email, password))
-
-
-def create_user_template(
-    superuser: str, staff: bool, username: str, email: str, password: str
-) -> str:
-    opts = ""
-    if superuser:
-        opts += " --superuser"
-    if staff:
-        opts += " --staff"
-    return f"""
-./manage.py lms manage_user {opts} {username} {email}
-./manage.py lms shell -c "
-from django.contrib.auth import get_user_model
-u = get_user_model().objects.get(username='{username}')
-u.set_password('{password}')
-u.save()"
-"""
 
 
 @click.command(help="Import the demo course")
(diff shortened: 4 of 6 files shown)


tutor/commands/upgrade/common.py
```diff
@@ -43,6 +43,39 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
     )
 
 
+def get_mysql_change_authentication_plugin_query(config: Config) -> str:
+    """Helper function to generate queries depending on the loaded plugins"""
```
DawoudSheraz, line 47: the docstring can be upgraded to highlight what is done by default and what is done based on plugin's availability.

tutor/commands/upgrade/compose.py
```diff
@@ -165,6 +166,48 @@ def upgrade_from_quince(context: click.Context, config: Config) -> None:
...
 
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Quince."
+        )
+        return
+
+    # Revert the MySQL image first to build data dictionary on v8.1
```
DawoudSheraz, line 177: hmmm, if someone has already ran upgrade from quince, they would not be able to run it again successfully due to Mongo upgrade. Should we convert this into a do command as well?

tutor/commands/upgrade/k8s.py
```diff
@@ -172,6 +172,36 @@ def upgrade_from_quince(config: Config) -> None:
...
+            "nothing left to do to upgrade from Quince."
+        )
+        return
+
+    query = common_upgrade.get_mysql_change_authentication_plugin_query(config)
+
+    # We need to first revert MySQL back to v8.1 to build the data dictionary on it
+    # And also to update the authentication plugin as it is disabled on v8.4
+    message = f"""Automatic release upgrade is unsupported in Kubernetes. If you are upgrading from Olive or an earlier release to Redwood, you
+    should upgrade the authentication plugin of your users. To upgrade, run something similar to:
```
DawoudSheraz, line 188: Instead of saying run something similar, we should add the exact commands the user should run.

tutor/commands/upgrade/common.py
Danyal-Faheem: Updated.

tutor/commands/upgrade/k8s.py
Danyal-Faheem: These are the exact commands that the user should run. I've updated the wording in the message.

tutor/commands/upgrade/compose.py
Danyal-Faheem: Hmm, I didn't think of that. Maybe we can add a small check if Mongo is already updated to v7.0.7 and we should not upgrade it then?

DawoudSheraz: Hmmm, I think that having to run this as part of upgrade is going to be weird, considering it is mid release. 
We can add do command and maybe package this in upgrade for sumac?

Danyal-Faheem: Alright, sure. This change shouldn't take too much time. I'll convert this into a do command and update the PR.

changelog.d/20240718_171945_danyal.faheem_mysql_authentication_plugin_change.md
```diff
@@ -0,0 +1,4 @@
+- [Bugfix] Manually update the authentication plugin of MySQL users from mysql_native_password to caching_sha2_password when upgrading from quince. (by @Danyal-Faheem)
+   This process should be automatic for most users. However, if you are running a third-party MySQL (i.e., RUN_MYSQL=false), you are expected to perform this process yourself. Please refer to the third-party provider's documentation for detailed instructions. Ensuring that your MySQL version is up-to-date is crucial for maintaining compatibility and security.
+- [Bugfix] Do not directly upgrade MySQL from v5.7 to v8.4 when upgrading from quince as MySQL does not allow that. First, upgrade to v8.1 and then to v8.4. (by @Danyal-Faheem)
+  This process should be automatic for most users. However, if you are running a third-party MySQL (i.e., RUN_MYSQL=false), you are expected to perform this process yourself. Please refer to the third-party provider's documentation for detailed instructions. Ensuring that your MySQL version is up-to-date is crucial for maintaining compatibility and security.
```
DawoudSheraz, line 4: nit: please remove this changelong entry bullet point.

tutor/commands/upgrade/common.py
```diff
@@ -43,6 +43,46 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
     )
 
 
+def get_mysql_change_authentication_plugin_query(config: Config) -> str:
```
DawoudSheraz, line 46: Wondering if we should have an override/additional param provided by user to only upgrade a particular plugin's database. Lets say they have upgraded the default already and do not have plugins enabled. Once plugins are enabled, they will be re-doing the migration for default.

tutor/commands/upgrade/common.py
```diff
@@ -43,6 +43,46 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
...
+    These users are also upgraded
+    """
+    loaded_plugins = list(plugins.iter_loaded())
+
+    def generate_mysql_authentication_plugin_update_query(
+        username: ConfigValue, password: ConfigValue, host: str
+    ) -> str:
+        return f"ALTER USER '{username}'@'{host}' IDENTIFIED with caching_sha2_password BY '{password}';"
+
+    host = "%"
```
DawoudSheraz, line 62: why is the host set to %?

tutor/commands/upgrade/compose.py
DawoudSheraz: I suspect this is to be removed?

tutor/commands/upgrade/compose.py
```diff
@@ -165,6 +166,52 @@ def upgrade_from_quince(context: click.Context, config: Config) -> None:
...
+    fmt.echo_info("Waiting for MySQL to boot...")
+    sleep(30)
+
+    query = common_upgrade.get_mysql_change_authentication_plugin_query(config)
+
+    # Update the authentication plugin in v8.1 as it is disabled in v8.4
+    context.invoke(
+        compose.execute,
+        args=[
+            "mysql",
```
DawoudSheraz, line 195: do we need this, as we are handling this in https://github.com/overhangio/tutor/pull/1102?

tutor/commands/upgrade/common.py
Danyal-Faheem: The host for [users created](https://github.com/overhangio/tutor/blob/a97a7b0abdb1ae73df2f0769a2609e6767b40075/tutor/templates/jobs/init/mysql.sh#L20) by Tutor is always set to '%'. The users created by MySQL itself uses the 'localhost' host.

tutor/commands/upgrade/common.py
Danyal-Faheem: Yes, I was also thinking that we could add a `--users` option with comma separated name of users to the do command. This way, users can also choose to upgrade their choice of databases.

changelog.d/20240718_171945_danyal.faheem_mysql_authentication_plugin_change.md
Danyal-Faheem: Removed.

tutor/commands/upgrade/common.py
Danyal-Faheem: I've added a `--users` option as well.

tutor/templates/local/docker-compose.yml
```diff
@@ -44,7 +44,7 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
-      --mysql-native-password=ON
+      {% if DOCKER_IMAGE_MYSQL >= "docker.io/mysql:8.4.0" -%}--mysql-native-password=ON{%- endif %}
```
hinakhadim, line 47: @Danyal-Faheem I'm assuming that this condition says If MySQL has a version greater than `8.4.0` then add `--mysql-native-password=ON`.

This condition fails for the following cases on my system:
1. `tutor config save --set DOCKER_IMAGE_MYSQL=docker.io/mysql:8.14.0`
2. `tutor config save --set DOCKER_IMAGE_MYSQL=docker.io/mysql:10.0.0`

**Reason**: The >= operator is typically used for numerical or lexical (alphabetical) comparisons. Since Docker image versions are strings, this condition is comparing the strings lexicographically, not numerically.

Danyal-Faheem: Great catch, I didn't think of the edge case. Thank you for picking it out. Let me update this to a number based comparison.

tutor/templates/local/docker-compose.yml
```diff
@@ -44,7 +44,11 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
+      # We only require this option for MySQL 8.4 and above
+      # Breaks MySQL for previous versions as this option does not exist on versions earlier than 8.4
+      {% if DOCKER_IMAGE_MYSQL.split(':')[-1] == "latest" or (DOCKER_IMAGE_MYSQL.split(':')[-1].split('.') | map('int') | list >= '8.4.0'.split('.') | map('int') | list) -%}
```
DawoudSheraz, line 49: nit: we can set DOCKER_IMAGE_MYSQL.split(':')[-1] to a variable and use that in comparison. It will make the the code readable.

docs/local.rst
```diff
@@ -141,6 +141,26 @@ The default Open edX theme is rather bland, so Tutor makes it easy to switch to
...
+
+Tutor makes it easy do so with this handy command::
+
+    tutor local do update_mysql_authentication_plugin
+
+If you only want to update the authentication plugin of specific users, you can use the ``--users`` option. This option takes comma seperated names of users to upgrade::
+
+    tutor local do update_mysql_authentication_plugin --users=discovery,ecommerce
+
+Do note that if you are updating a specific user, there should be corresponding entries in the configuration for the mysql username and password for that user. For example, if you are trying to update the user ``myuser``, the following case sensitive entries need to be present in the configuration::
```
DawoudSheraz, line 159: this paragraph can be reworded to focus that for any particular user, the command expects the configuration settings in a certain format. Then, the myuser examples can be added afterwards/

tutor/templates/local/docker-compose.yml
```diff
@@ -44,7 +44,12 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
+      # We only require this option for MySQL 8.4.0 and above
+      # Breaks MySQL for previous versions as this option does not exist on versions earlier than 8.4.0
+      {% set mysql_version = DOCKER_IMAGE_MYSQL.split(':')[-1] -%}
```
DawoudSheraz, line 49: nit: please do the same in k8s file as well.

DawoudSheraz (commented):
While testing, I came across the following error
```
Error: Username or Password for User mfe not found in config. Please make sure that the following entries are present in the configuration:
MFE_MYSQL_USERNAME
MFE_MYSQL_PASSWORD
```
The current code raises the exception if config is not found. It is iterating on all the plugins. Not all plugins would have DB users. Instead of raising, we can log warning and continue.

docs/local.rst
Danyal-Faheem: I've updated this paragraph.

tutor/templates/local/docker-compose.yml
Danyal-Faheem: Done.

DawoudSheraz (approved):

Faraz32123:
Hello @regisb, can u have a look at it for it's final iteration?

tutor/fmt.py
```diff
@@ -46,6 +46,14 @@ def alert(text: str) -> str:
     return click.style("⚠️  " + text, fg="yellow", bold=True)
 
 
+def warning(text: str) -> str:
```
regisb, line 49: What's the purpose of `warning` when we already have `alert`?

tutor/templates/local/docker-compose.yml
```diff
@@ -44,7 +44,12 @@ services:
       --character-set-server=utf8mb4
       --collation-server=utf8mb4_unicode_ci
       --binlog-expire-logs-seconds=259200
+      # We only require this option for MySQL 8.4.0 and above
```
regisb, line 47: I don't think we should include this check. If we are running MySQL, then it's not our job to remain compatible with all possible versions of MySQL. We just assume that we are running the version that is in config.yml.

Let's remove this change, along with the one in k8s/deployments.yml.

tutor/commands/jobs.py
```diff
@@ -315,6 +317,49 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
     yield ("lms", command)
 
 
+@click.command(context_settings={"ignore_unknown_options": True})
+@click.option(
```
regisb, line 321: This should not be an option, but an argument, because it's mandatory. The command signature should not be:

     tutor local do update_mysql_authentication_plugin --users=regis,danyal

but:

    tutor local do update_mysql_authentication_plugin regis danyal

Moreover, we should be able to update all users with users=all:

    tutor local do update_mysql_authentication_plugin all

tutor/commands/jobs.py
```diff
@@ -315,6 +317,49 @@ def sqlshell(args: list[str]) -> t.Iterable[tuple[str, str]]:
...
+        )
+        return
+
+    users_to_update = users.split(",") if users else list(plugins.iter_loaded())
+
+    query = get_mysql_change_authentication_plugin_query(
+        config, users_to_update, not users
+    )
+
+    # In case there is no user to update the authentication plugin of
```
regisb, line 351: Let's return earlier if there is no user:

    if not users:
        return

tutor/commands/jobs_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    host = "%"
+    query = ""
+
+    def generate_mysql_authentication_plugin_update_query(
+        username: ConfigValue, password: ConfigValue, host: str
+    ) -> str:
+        fmt.echo_info(
+            f"Authentication plugin of user {username} will be updated to caching_sha2_password"
+        )
+        return f"ALTER USER '{username}'@'{host}' IDENTIFIED with caching_sha2_password BY '{password}';"
```
regisb, line 47: This function is extremely confusing, and there are many ways in which it can fail. In order to improve it, I'd like to understand if it's possible to automatically detect all mysql users that were created with the legacy authentication plugin?

regisb (changes requested):

tutor/commands/jobs_utils.py
Danyal-Faheem: The only way to do that would be to get the output of a MySQL query inside tutor codebase. I am not aware of any method that we can use to get the output of a command run inside a container inside the tutor codebase.

tutor/fmt.py
Danyal-Faheem: Alert prints to standard error and this seemed like a situation where we just wanted to display a warning rather than an error.

tutor/templates/local/docker-compose.yml
Danyal-Faheem: This was added for compatibility with #1102. When we downgrade MySQL to v8.1, it will crash if this option exists as the option does not exist in MySQL v8.1. We can make this check a lot simpler, with something like this:

```
{% if DOCKER_IMAGE_MYSQL >= "docker.io/mysql:8.4.0" -%}- "--mysql-native-password=ON"{%- endif %}
```
However, that does not handle any edge cases.

tutor/commands/jobs_utils.py
regisb: Assuming we could get the output of a mysql query in tutor, what would be that query?

Danyal-Faheem: Something similar to this:

```
select user from mysql.user where plugin='mysql_native_password' and host='%';
```

We need to limit the host to '%' as Tutor uses that host value by default.

Danyal-Faheem: Hey Regis, any thoughts on how we should proceed with this?

tutor/fmt.py
regisb: Let's keep an alert. We don't want to write to stdout because it might mangle output from other commands, such as `tutor config printvalue/printroot`. And if we write to stderr than we are better off just using alert.

tutor/commands/jobs_utils.py
regisb: I'm not sure. basically, we need to perform all tasks inside a single templated bash script -- including the ` mysql --user...` part. This bash script can be templated, such that we don't have to pass config values to the script-generating function. I'm sorry I don't have time to think about this more...

Danyal-Faheem: So maybe something similar to a stored procedure? Something akin to what we did in #1079?

We could get the names of all the users and update them inside that procedure. Although, I'd have to look into how we would manage passwords then.

tutor/commands/jobs_utils.py
```diff
@@ -131,3 +137,71 @@ def get_mysql_change_charset_query(
...
+        )
+        return f"ALTER USER IF EXISTS '{username}'@'{host}' IDENTIFIED with caching_sha2_password BY '{password}';"
+
+    def generate_user_queries(users: Sequence[str]) -> str:
+        query = ""
+        for user in users:
+            user_uppercase = user.upper()
+            if not (
+                f"{user_uppercase}_MYSQL_USERNAME" in config
+                and f"{user_uppercase}_MYSQL_PASSWORD" in config
```
regisb, line 183: This looks very fragile. We can't rely on the arbitrary naming of some configuration settings to auto-detect user names.

Danyal-Faheem: We [found](https://github.com/search?q=path%3A**%2Fplugin.py+MYSQL_PASSWORD%22%3A&type=code&ref=advsearch&p=1) that this was a convention used across the plugins, but I can understand your position here. However, we still need the exact username and password to perform this operation. Therefore, I propose that we can provide an option to provide the exact username and password. Something similar to:

```
tutor local do update-mysql-authentication-plugin discovery ecommerce --usernames discovery-username,ecommerce-username --passwords discovery-password,ecommerce-password
```
This isn't very intuitive but since this would be a one time operation, maybe it is fine? Another con I see here is that users will have to have passwords out in the open for headless scripts unless they utilize environment variables.

I am open to any suggestions you might have.

Danyal-Faheem: Hi @regisb, pinging you again for your thoughts on this.

regisb: Let's go with a mandatory `username` CLI argument. When the `--password=...` option is not provided, it should be interactively set. (see how the `createuser` command handles password definition) It's not a big deal if passwords are printed in stdout.

Then, we should instruct users to migrate their accounts with:

    tutor local do update-mysql-authentication-plugin \
        --password="$(tutor config printvalueOPENEDX_MYSQL_PASSWORD)"\
        $(tutor config printvalue OPENEDX_MYSQL_USERNAME)

Danyal-Faheem: This looks good. However, I have one concern. This basically limits us to updating one user at a time. If a user has 5-6 plugins enabled, this would become quite hectic as the MySQL root user and Open edX mysql user would require 2 separate commands by themselves along, not to mention for each plugin as well.

regisb: I agree, but I think it's OK, because this command will only have to be run once for every mysql user.

Danyal-Faheem: I've updated the command to reflect the changes as you mentioned.

I've also moved some code to the newly created jobs_utils file as the jobs.py file was getting too bloated.

docs/local.rst
```diff
@@ -168,6 +168,24 @@ By default, only the tables in the openedx database are changed. For upgrading t
...
+
+    tutor local do update-mysql-authentication-plugin myuser
+
+The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command::
+
+    tutor local do update-mysql-authentication-plugin myuser --password=mypassword
+
+Tutor may prompt you with some warnings if the entered password is suspected to be wrong. To avoid these prompts, use the non-interactive option::
+
+    tutor local do update-mysql-authentication-plugin myuser --password=mypassword --non-interactive
```
regisb, line 188: This line is causing the docs build to fail. Add one empty line.

docs/local.rst
```diff
@@ -168,6 +168,24 @@ By default, only the tables in the openedx database are changed. For upgrading t
...
+.. _update_mysql_authentication_plugin:
+
+Updating the authentication plugin of MySQL users
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As of MySQL v8.4.0, the ``mysql_native_password`` authentication plugin has been deprecated. Users created with this authentication plugin should ideally be updated to use the latest ``caching_sha2_password`` authentication plugin.
+
+Tutor makes it easy do so with this handy command::
+
+    tutor local do update-mysql-authentication-plugin myuser
```
regisb, line 180: replace "myuser" by "USERNAME" to make it clearer that this part needs to be edited by the user.

docs/local.rst
```diff
@@ -168,6 +168,24 @@ By default, only the tables in the openedx database are changed. For upgrading t
...
+
+The password will be required to be entered interactively. Optionally, the password can also be provided as part of the command::
+
+    tutor local do update-mysql-authentication-plugin myuser --password=mypassword
```
regisb, line 184: Please add an example with `--password=$(tutor config printvalue ...)`

docs/troubleshooting.rst
```diff
@@ -216,3 +216,10 @@ NPM Dependency Conflict When overriding ``@edx/frontend-component-header`` or ``
 ----------------------------------------------------------------------------------------------------------------
 
 The detailed steps are mentioned in `tutor-mfe <https://github.com/overhangio/tutor-mfe?tab=readme-ov-file#npm-dependency-conflict-when-overriding-edxfrontend-component-header-or-edxfrontend-component-footer>`__ documentation.
+
+"Plugin 'mysql_native_password' is not loaded"
+----------------------------------------------
+
+This issue can occur when Tutor is upgraded from v15 (Olive) or earlier to v18 (Redwood) because the users created in Tutor v15 utilize the mysql_native_password authentication plugin by default. This plugin has been deprecated as of MySQL v8.4.0 which is the default MySQL server used in Tutor v18.
+
+The handy :ref:`update-mysql-authentication-plugin <update_mysql_authentication_plugin>` do command in tutor can be used to fix this issue.
```
regisb, line 225: What is the command that should be run for a vanilla installation?

tutor/commands/jobs.py
```diff
@@ -428,6 +378,79 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+@click.option(
+    "-p",
+    "--password",
+    help="Specify password from the command line. If undefined, you will be prompted to input a password",
+    prompt=True,
+    hide_input=True,
+)
+@click.argument(
+    "user",
+    nargs=1,
```
regisb, line 396: `nargs=1` can be removed, unless I'm mistaken.

tutor/commands/jobs.py
```diff
@@ -428,6 +378,79 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+    config = tutor_config.load(context.root)
+
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=False). It is your "
+            "responsibility to update the authentication plugin of mysql users."
+        )
+        return
+
+    conventional_password_key = f"{user.upper()}_MYSQL_PASSWORD"
```
regisb, line 417: This part about automatically loading password from config should be removed. In particular, with the current implementation, a warning will be printed even after user inputs a password.

tutor/commands/jobs.py
```diff
@@ -428,6 +378,79 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+        ):
+            return
+
+    host = "%"
+
+    query = f"ALTER USER IF EXISTS '{user}'@'{host}' IDENTIFIED with caching_sha2_password BY '{password}';"
+
+    mysql_command = (
+        "mysql --user={{ MYSQL_ROOT_USERNAME }} --password={{ MYSQL_ROOT_PASSWORD }} --host={{ MYSQL_HOST }} --port={{ MYSQL_PORT }} --database={{ OPENEDX_MYSQL_DATABASE }} --show-warnings "
+        + shlex.join(["-e", query])
```
regisb, line 448: I think the following will be more robust:

```
 yield ("lms", shlex.join([
    "mysql",
    "--user={{ MYSQL_ROOT_USERNAME }}",
    "--password={{ MYSQL_ROOT_PASSWORD }}",
    "--host={{ MYSQL_HOST }}",
    "--port={{ MYSQL_PORT }}",
    "--database={{ OPENEDX_MYSQL_DATABASE }}",
    "--show-warnings",
    "-e",
    query,
]))
```

tutor/commands/jobs_utils.py
```diff
@@ -2,7 +2,29 @@
 This module provides utility methods for tutor `do` commands
 
 Methods:
+- `create_user_template`: Generates the necessary commands to create a user in openedx.
 - `get_mysql_change_charset_query`: Generates MySQL queries to upgrade the charset and collation of columns, tables, and databases.
+- `set_theme_template`: Generates the necessary commands to set a theme on a specific domain in openedx.
```
regisb, line 7: Let's not adopt the habit of documenting the functions at the top of the file. Can you please move the function short descriptions to the function docstrings? (in triple quotes)

regisb (changes requested):

docs/local.rst
Danyal-Faheem: Updated.

docs/local.rst
Danyal-Faheem: I've added an example

docs/troubleshooting.rst
Danyal-Faheem: I've added the two commands necessary for a vanilla installation in the troubleshooting section as well

tutor/commands/jobs.py
Danyal-Faheem: You're right! I though `nargs=1` was necessary to specify that an argument is not optional but I guess that's why options exist...

tutor/commands/jobs.py
Danyal-Faheem: Regarding this,

The warnings will only be printed if the password entered by a user does not match the conventional format that tutor expects or if the value in config of the conventional key does not match the one entered by the user. This is just done to make sure no one accidentally updates the password of their mysql users and is left wondering why their platform doesn't work now.

There is also an option of using the command non-interactively.

Do you still think we should remove this? I understand its not intuitive at all but I added it just in case someone doesn't accidentally update their passwords.

tutor/commands/jobs.py
Danyal-Faheem: This certainly looks better. I've updated the original one to this.

tutor/commands/jobs_utils.py
Danyal-Faheem: I did it as @DawoudSheraz requested me to add file docstrings. But, I guess we're not following that convention for now. So, I've removed the file docstring in favor of function docstrings.

tutor/commands/jobs.py
regisb: Yeah, I think we should remove this, as it will cause more confusion than resolve anything. If the password is incorrect, the command will fail, right? (which is good)

Danyal-Faheem: > If the password is incorrect, the command will fail, right? (which is good)

Unfortunately, it wont fail. Instead, mysql will just update the password of the user to the incorrectly entered password and then the platform will be unable to connect because the password has been updated. This was my main motivation for adding these checks.

regisb: OK, I understand. Here's what I suggest:

1. Get rid of the interactive option, remove the corresponding warning.
2. Remove `prompt=True` in the password option definition.
3. In the command definition, build a dict of known username/passwords:

```
known_mysql_credentials = {}
for k, v in [("OPENEDX_MYSQL_USERNAME", "OPENEDX_MYSQL_PASSWORD"), ("NOTES_MYSQL_USERNAME", "NOTES_MYSQL_PASSWORD"), ...]:
    if username := config.get(k):
        known_mysql_credentials[username] = config[v]
if not password:
    password = known_mysql_credentials.get(username)
if not password:
    # prompt for password
```

In other words, the password is automatically detected if it matches a known username. Optionally, we can build `known_mysql_credentials` from a new filter.

4. Change the documentation to tell users to run:
 
    tutor local do update-mysql-authentication-plugin $(tutor config printvalue OPENEDX_MYSQL_USERNAME)
    tutor local do update-mysql-authentication-plugin $(tutor config printvalue NOTES_MYSQL_USERNAME)
    ...

Would that work?

Sorry about the back and forth on this feature. I'm aware that this PR has been open for 6+ months, and I'm eager to merge it.

Danyal-Faheem: I like this idea as well.

> Optionally, we can build known_mysql_credentials from a new filter.

Since this is a one time command, lets not go this route. We can hardcode the values and if someone is using a custom plugin, they can just use the username password option.

I've updated the rest of the code and docs as per the method you mentioned here.

tutor/commands/jobs.py
```diff
@@ -428,6 +377,95 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
     fmt.echo_info("MySQL charset and collation successfully upgraded")
 
 
+@click.command(
+    short_help="Update the authentication plugin of a mysql user to caching_sha2_password.",
+    help=(
+        "Update the authentication plugin of a mysql user to caching_sha2_password from mysql_native_password. You can specify either specific users to update or all to update all users."
```
DawoudSheraz, line 383: please update the help text as the original text is not up-to-date.

tutor/commands/jobs.py
```diff
@@ -428,6 +377,95 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+            "You are not running MySQL (RUN_MYSQL=False). It is your "
+            "responsibility to update the authentication plugin of mysql users."
+        )
+        return
+
+    # Official plugins that have their own mysql user
+    known_plugins_with_mysql_keys = [
+        "credentials",
+        "discovery",
+        "ecommerce",
```
DawoudSheraz, line 416: ecommerce has been deprecated, let's not add it here.

tutor/commands/jobs.py
```diff
@@ -428,6 +377,95 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+        "jupyter",
+        "notes",
+        "openedx",
```
DawoudSheraz, line 419: openedx user is not part of plugin but rather core, it does not belong in the list. We can rename the variable to known_mysql_users (or something along these lines) and then clarify via comments which is core and which is from plugins.

tutor/commands/jobs.py
```diff
@@ -428,6 +377,95 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+        "xqueue",
+    ]
+
+    # Create a list of the usernames and passwords
```
DawoudSheraz, line 423: nit: Create a list of the username and password config variables/keys

DawoudSheraz (approved):

tutor/commands/jobs.py
Danyal-Faheem: Updated

tutor/commands/jobs.py
```diff
@@ -428,6 +377,95 @@ def generate_query_to_append(tables: list[str], exclude: bool = False) -> str:
...
+    known_mysql_credentials_keys = [
+        (f"{plugin.upper()}_MYSQL_USERNAME", f"{plugin.upper()}_MYSQL_PASSWORD")
+        for plugin in known_plugins_with_mysql_keys
+    ]
+    # Add the root user as it is the only one that is different from the rest
+    known_mysql_credentials_keys.append(("MYSQL_ROOT_USERNAME", "MYSQL_ROOT_PASSWORD"))
+
+    known_mysql_credentials = {}
+    # Build the dictionary of known credentials
```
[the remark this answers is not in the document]
Danyal-Faheem: Updated

tutor/commands/jobs.py
Danyal-Faheem: I've updated the variable name and added a small comment to signify the difference between plugin users and core user.

tutor/commands/jobs.py
Danyal-Faheem: You are right, thanks for this catch. 

> Or should we, for backward compatibility purposes?

Since this PR will be merged in release and will be launched in a version of Tutor v19.0.1+, it does not make sense to keep ecommerce. If someone is running ecommerce, they can still use the command and provide the password as a command line option.

regisb (approved):
Looks awesome to me :) thanks for the hard work!

PR #1099, opened 2024-07-24
feat: adds `tutor config edit`

tecoholic:
Quickly launch the default YAML editor for editing the `config.yml` file.

### Background

As a developer working with multiple projects, I am constantly changing the tutor config and re-running the `tutor dev launch` to update the devstack. While it is as simple as running `vim $(tutor config printroot)/config.yml`, I often found myself wanting to simply express this better using tutor config edit. So this is an attempt at cross-platform solution doing the same. It might not be the best solution, but something to start with if someone else finds this useful and improves their DevEx.

*Caveat:* I have only tested this on Linux (both open and `xdg-open` works in mine, so it could be said Unix).

### DIFF
--- a/changelog.d/20241114_115534_arunmozhi_add_config_edit_command.md
+++ b/changelog.d/20241114_115534_arunmozhi_add_config_edit_command.md
@@ -1,0 +1,1 @@
+- [Feature] Adds `tutor config edit`. This opens the active tutor environment's config.yaml in an editor for manual editing. (by @tecoholic)
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -1,15 +1,19 @@
 from __future__ import annotations
 
 import json
+import os.path
+import subprocess
 import typing as t
 
 import click
 import click.shell_completion
 
+from shutil import which
+
 from tutor import config as tutor_config
 from tutor import env, exceptions, fmt, hooks
 from tutor import interactive as interactive_config
-from tutor import serialize
+from tutor import serialize, utils
 from tutor.commands.context import Context
 from tutor.commands.params import ConfigLoaderParam
 from tutor.types import Config, ConfigValue
@@ -255,9 +259,37 @@ def patches_show(context: Context, name: str) -> None:
         print(rendered)
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
+@click.pass_obj
+def edit(context: Context, editor: str, save: bool) -> None:
+    config_file = tutor_config.config_path(context.root)
+
+    if not os.path.isfile(config_file):
+        raise exceptions.TutorError(
+            f"Missing config file at {config_file}. Have you run 'tutor local launch' yet?"
+        )
+
+    open_cmd = []
+    if which("open"):  # MacOS & linux distributions that ship `open`. eg., Ubuntu
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        # Calling "start" on a regular file opens it with the default editor.
+        # The second argument "" just means "don't give the window a custom title".
+        open_cmd = ["start", '""', config_file]
+    else:
+        raise exceptions.TutorError(
+            f"Failed to find utility to launch an editor. Edit {config_file} with the editor of your choice."
+        )
+
+    utils.execute(*open_cmd)
+
+
 config_command.add_command(save)
 config_command.add_command(printroot)
 config_command.add_command(printvalue)
 patches_command.add_command(patches_list)
 patches_command.add_command(patches_show)
 config_command.add_command(patches_command)
+config_command.add_command(edit)
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -261,7 +261,7 @@ def patches_show(context: Context, name: str) -> None:
 
 @click.command(name="edit", help="Edit config.yml of the current environment")
 @click.pass_obj
-def edit(context: Context, editor: str, save: bool) -> None:
+def edit(context: Context) -> None:
     config_file = tutor_config.config_path(context.root)
 
     if not os.path.isfile(config_file):
--- a/tutor/commands/config.py
+++ b/tutor/commands/config.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 import os.path
-import subprocess
 import typing as t
 
 import click


tutor/commands/config.py
```diff
@@ -220,8 +224,32 @@ def patches_list(context: Context) -> None:
     renderer.print_patches_locations()
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
+@click.pass_obj
+def edit(context: Context) -> None:
+    config_file = os.path.join(context.root, "config.yml")
```
regisb, line 230: Use `config.config_path(context.root)` instead.

tutor/commands/config.py
```diff
@@ -220,8 +224,32 @@ def patches_list(context: Context) -> None:
...
+    open_cmd = None
+
+    if which("open"):  # MacOS
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        open_cmd = ["start", '""', config_file]
+    else:
+        click.echo(
```
regisb, line 241: This command should fail if neither open commands are available. You could for instance `raise exceptions.TutorError`.

tutor/commands/config.py
```diff
@@ -220,8 +224,32 @@ def patches_list(context: Context) -> None:
...
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
+@click.pass_obj
+def edit(context: Context) -> None:
+    config_file = os.path.join(context.root, "config.yml")
+
+    open_cmd = None
+
+    if which("open"):  # MacOS
```
regisb, line 234: Let's amend the comment. `open` is also available on Ubuntu, for instance.

I have another, more important problem with this command. According to my testing, when I run `open file.txt` on a headless server (for instance: a remote production server), it opens a read-only editor. How can we fix this command to open a write-capable editor? (`xdg-open` is not available on servers without a Desktop) Note that `$EDITOR` is set to "vim" on this server.

tutor/commands/config.py
```diff
@@ -220,8 +224,32 @@ def patches_list(context: Context) -> None:
...
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        open_cmd = ["start", '""', config_file]
+    else:
+        click.echo(
+            "Cannot find a way to open the editor automatically. Kindly open the file manually: "
```
regisb, line 242: Let's rephrase that along the lines of: "Failed to detect an adequate text file editor". (mode concise, no need for "kindly")

tutor/commands/config.py
```diff
@@ -220,8 +224,32 @@ def patches_list(context: Context) -> None:
...
+            + config_file
+        )
+        return
+
+    subprocess.call(open_cmd)
```
regisb, line 247: This command will do something funky if the file or its directory does not exist. I suggest the command should fail if the config file does not exist.

regisb (changes requested):

tecoholic:
@regisb Hi, Thanks for the review comments. I hoped to have addressed them by now. Somehow haven't found the time. I will handle the comments this week.

kdmccormick:
I have nothing to add beyond Regis's review--just want to say that this is a great idea and thank you for the contribution 👍🏻

tutor/commands/config.py
tecoholic: @regisb I wanted to test this scenario to evaluate the options. However, the server image I used (ubuntu based) didn't have a `open` at all. And my desktop with Ubuntu seems to just ship `xdg-open` as `open`. 

```
❯ which open
/usr/bin/open

❯ open --version
xdg-open 1.1.3
```

> when I run `open file.txt` on a headless server (for instance: a remote production server), it opens a read-only editor.

Which version of `open` is this, and what's the editor that it opens?

tutor/commands/config.py
tecoholic: Great catch! The `open`s I tested all launched the browser with the argument as the URL.

tutor/commands/config.py
tecoholic: On the topic of a "writable editor", MacOS's `open` implements `-a` flag, which allows specifying the application to open the file in. I am wondering if something similar can be added to this command which can force a specific editor to be launched. `tutor config edit -a vim` maybe? What do think about this?

tutor/commands/config.py
```diff
@@ -220,8 +224,31 @@ def patches_list(context: Context) -> None:
...
+    open_cmd = None
+
+    if which("open"):  # MacOS & linux distributions that ship `open`. eg., Ubuntu
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        open_cmd = ["start", '""', config_file]
+    else:
+        raise exceptions.TutorError(f"Failed to find utility to launch an editor.")
```
DawoudSheraz, line 244: In the error message, we can specify which utilities are used by the command internally so that user can install them on the system if needed (not all the users of Tutor are dev, some are operators who might not want to delve into code to see what is happening under the hood).

tutor/commands/config.py
DawoudSheraz: I think this can be a good option, this gives control to the user. For instance, if someone wants to open the file in Sublime/VScode, they can override the default utils the command is checking.

kdmccormick:
Hi @tecoholic . I'm looking forward to this change! Looks like there are just a couple minor change requests left, and then we're good to merge this.

tecoholic:
@kdmccormick Hi, thanks for the ping. I was waiting on @regisb for more details about [read only editor](https://github.com/overhangio/tutor/pull/1099#discussion_r1713097296). Can we skip that scenario for now?

regisb:
@kdmccormick I am less interested in this feature than you seem to be -- mostly because (IMHO) it doesn't bring anything more than a .bashrc alias. E.g: `alias tutor-config-edit='vim "$(tutor config printroot)/env.yml"'`. So I'd rather defer to you for review.

tutor/commands/config.py
```diff
@@ -230,8 +234,31 @@ def patches_list(context: Context) -> None:
...
+        raise exceptions.TutorError(f"Missing config file at {config_file}")
+
+    open_cmd = None
+
+    if which("open"):  # MacOS & linux distributions that ship `open`. eg., Ubuntu
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        open_cmd = ["start", '""', config_file]
```
kdmccormick, line 252:
```suggestion
        # Calling "start" on a regular file opens it with the default editor.
        # The second argument "" just means "don't give the window a custom title".
        open_cmd = ["start", '""', config_file]
```

tutor/commands/config.py
```diff
@@ -230,8 +234,31 @@ def patches_list(context: Context) -> None:
     renderer.print_patches_locations()
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
+@click.pass_obj
+def edit(context: Context) -> None:
+    config_file = tutor_config.config_path(context.root)
+
+    if not os.path.isfile(config_file):
+        raise exceptions.TutorError(f"Missing config file at {config_file}")
```
kdmccormick, line 243:
```suggestion
        raise exceptions.TutorError(f"Missing config file at {config_file}. Have you run 'tutor local launch' yet?")
```
In most cases, I believe that the file will be missing if and only if the Tutor has never launched Tutor. What do you think about adding a hint for users here?

tutor/commands/config.py
kdmccormick: After some research, it seems to me that Windows will always have `start`, macOS will always have `open`, and Linux desktop environments will always have `xdg-open`. That leaves only non-desktop Linux users and perhaps BSD users. I think those users will know their systems better then we will, and they will be capable of figuring out how to open a yaml file as long as we tell them where it is 😄  @tecoholic , rather than letting them know what to install, could you just enhance this error message a bit just to let users know that they can edit configuration at `{config_file}`?

tutor/commands/config.py
```diff
@@ -230,8 +234,31 @@ def patches_list(context: Context) -> None:
...
+        open_cmd = ["open", config_file]
+    elif which("xdg-open"):  # Linux
+        open_cmd = ["xdg-open", config_file]
+    elif which("start"):  # Windows
+        open_cmd = ["start", '""', config_file]
+    else:
+        raise exceptions.TutorError(f"Failed to find utility to launch an editor.")
+
+    subprocess.call(open_cmd)
+
```
kdmccormick, line 257: What do you think about adding a message at the end here along the lines of `Remember to run 'tutor config save' in order to apply configuration edits to your environment.`

kdmccormick (commented):
Thanks for the patience @tecoholic . If you can rebase and address these comments, I'll be happy to merge this.

tecoholic:
@kdmccormick Thank you for your comments and your patience with this PR. I have addressed your comments to my best. Kindly take another look when you can.

DawoudSheraz (commented):
Please add a changelog entry for this change.

tutor/commands/config.py
```diff
@@ -255,9 +259,41 @@ def patches_show(context: Context, name: str) -> None:
         print(rendered)
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
```
kdmccormick, line 262: Optional: What do you think about adding a `-s/--save` flag here? If supplied, then instead of reminding the user to run `tutor config save`, it would just save config automatically.

tutor/commands/config.py
```diff
@@ -255,9 +259,41 @@ def patches_show(context: Context, name: str) -> None:
...
+    elif which("start"):  # Windows
+        # Calling "start" on a regular file opens it with the default editor.
+        # The second argument "" just means "don't give the window a custom title".
+        open_cmd = ["start", '""', config_file]
+    else:
+        raise exceptions.TutorError(
+            f"Failed to find utility to launch an editor. Edit {config_file} with the editor of your choice."
+        )
+
+    subprocess.call(open_cmd)
```
kdmccormick, line 287: For CLI UX consistency, please use the `execute` function from `tutor.utils`. It is the same as this line, except it prints out the underlying command in purple for the user's benefit, and it formats any errors in red instead of showing a stack trace.

kdmccormick (changes requested):
I tested locally, works great on Ubuntu! I also commented out the `which("open")` block in order to confirm that `xdg-open` also works.

In addition to Dawoud's request to add a changelog entry, I have just one more change request and one optional suggestion ⬇️

tutor/commands/config.py
tecoholic: This was something I thought of adding later based on user feedback. Now that you have pointed it out, I have taken it as the feeback and added 2 features I think are QOL improvments.

1. `-s` automatically updating the environment after a successful save. The caveat here is "successful save". More on them below.
2. `-e` to specify an editor of choice. The annoying thing about desktop systems is, installing new software somehow messes up file associations. I discovered after installing KDE that my YAML files now open in Kate Editor. Now I do `tutor config edit -e (vim|emacs|..)` to open the file in a specific editor.

**Save caveats:**
* Sometimes, when I do `:wq` in Neovim (Lazyvim config), it returns a non-zero error code. This prevents the "save" from happening.
* When the file gets opened in a GUI editor like Kate, it opens it as an independent process, instead of subprocess. So the `utils.execute()` returns as soon as the editor is launched instead of return after the editor is closed. So, the environment is updated with config before editing.

tecoholic:
@kdmccormick Thank you for the feedback. I added a changelog entry, new options `-s` and `-e` [explanation above](https://github.com/overhangio/tutor/pull/1099#discussion_r1841486466), made the messages a bit more verbose. Kindly take a look.

changelog.d/20241114_115534_arunmozhi_add_config_edit_command.md
```diff
@@ -0,0 +1,12 @@
...
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+
+<!-- - 💥[Feature] Foobarize the blorginator. This breaks plugins by renaming the `FOO_DO` filter to `BAR_DO`. (by @regisb) -->
+<!-- - [Improvement] This is a non-breaking change. Life is good. (by @billgates) -->
```
DawoudSheraz, line 11: please remove the default text from changelog.

changelog.d/20241114_115534_arunmozhi_add_config_edit_command.md
```diff
@@ -0,0 +1,12 @@
...
+[Feature] Adds `tutor config edit`. This opens the active tutor environment's config.yaml in an editor for manual editing. (by @tecoholic)
```
DawoudSheraz, line 12: nit: Please add `-` at the start so that it renders as a list item.

tecoholic:
@DawoudSheraz I have addressed your comments about the docs and updated the PR.

DawoudSheraz (approved):

tutor/commands/config.py
kdmccormick: Thank you for spelling out those caveats.

> Sometimes, when I do :wq in Neovim (Lazyvim config), it returns a non-zero error code. This prevents the "save" from happening.

Could you catch this error, and if `--save` was provided, warn the user that the editor exited non-zero and thus their config changes were not applied?

> When the file gets opened in a GUI editor like Kate, it opens it as an independent process, instead of subprocess. So the utils.execute() returns as soon as the editor is launched instead of return after the editor is closed. So, the environment is updated with config before editing.

Ah, this one is a bit worrying. Visual editors often will accept a flag that blocks the command until the editor is exited, e.g. `subl --wait` for Sublime Text. That would be useful for this particular scenario, e.g.,:

    tutor config edit -s -e 'subl --wait'

This will fail with your current code, but could be fixed by running `shlex.split` on the `-e` argument. 

All that said, with all details and nuances that these new flags make us consider, I feel that we are delaying the `tutor config edit` feature more than we need to. It would be nice to have the base command Sumac. @tecoholic  how would you feel about omitting `-s` and `-e` for now, and opening a follow-up PR to add them?

kdmccormick (changes requested):

tutor/commands/config.py
tecoholic: @kdmccormick Thanks for taking another look. I agree with your assessment about the flags being a bit sub-optimal. That's the reason, I initially only considered them as future improvements.

So, I will roll back those bits and leave the bare command alone in this PR.

tecoholic:
@kdmccormick Updated the PR with simpler version.

P.S: I left a reply to the comment in the review. But, it doesn't show up here. So, adding a note here, just in case.

tutor/commands/config.py
kdmccormick: Sounds good!

tutor/commands/config.py
```diff
@@ -255,9 +259,37 @@ def patches_show(context: Context, name: str) -> None:
         print(rendered)
 
 
+@click.command(name="edit", help="Edit config.yml of the current environment")
+@click.pass_obj
+def edit(context: Context, editor: str, save: bool) -> None:
```
kdmccormick, line 264:
```suggestion
def edit(context: Context) -> None:
```
Leftover from `-s` and `-e

kdmccormick (changes requested):
Just [one issue](https://github.com/overhangio/tutor/pull/1099/files#r1869713944), looks and works great otherwise. Thanks for proposing and working through this!

kdmccormick (changes requested):
Sorry @tecoholic , one last thing: could you change the PR's base to `release` and rebase? In order to get improvements out to the biggest set of users, the Tutor convention is to merge to `release` if there is nothing breaking, risky, or specific to Open edX's master branches.

tecoholic:
@kdmccormick Hi, thanks for letting me correct the branch. I couldn't move all the commits and the history with it. So, I just squashed everything and based it on the `release` branch. I have applied the fix you pointed out as well.

kdmccormick (approved):
Thank you @tecoholic !

PR #1102, opened 2024-07-26
fix: upgrade MySQL from 5.7 to 8.1 first and then to 8.4

Danyal-Faheem:
When a user attempts to directly upgrade their tutor instance from tutor v15 or earlier to tutor v18 or later, i.e upgrade from MySQL v5.7 or earlier to MySQL v8.4 or later, MySQL will crash with the container getting stuck in a boot loop. This is because upgrading directly from MySQL v5.7 -> v8.4 is [not allowed](https://dev.mysql.com/doc/refman/8.4/en/upgrade-paths.html). We therefore first jump to v8.1, build the data dictionary and then go to v8.4.

Note that this issue does not occur if one attempts to upgrade to tutor v16 or v17 first and then to tutor v18 because that upgrade path is acceptable for MySQL.

### DIFF
--- a/tutor/commands/upgrade/common.py
+++ b/tutor/commands/upgrade/common.py
@@ -83,7 +83,6 @@ def verify_tutor_version_for_mysql_upgrade(config: Config) -> tuple[bool, str]:
 
     new_mysql_docker_image = str(config["DOCKER_IMAGE_MYSQL"])
 
-    # Do not perform manual upgrade if running v16 or v17, only for tutor v18 or later
     if (
         new_mysql_docker_image == "docker.io/mysql:8.0.33"
         or new_mysql_docker_image == "docker.io/mysql:8.1.0"
--- a/tutor/commands/upgrade/common.py
+++ b/tutor/commands/upgrade/common.py
@@ -60,6 +60,39 @@ def get_mongo_upgrade_parameters(
     return mongo_version, admin_command
 
 
+def verify_tutor_version_for_mysql_upgrade(config: Config) -> tuple[bool, str]:
+    """
+    Checks if a MySQL upgrade is needed based on the Tutor version and MySQL setup.
+
+    This method ensures that MySQL is running and determines if the upgrade
+    process should proceed based on the Tutor version. It is intended for upgrades
+    from Tutor version 15 to version 18 or later. Manual upgrade steps are not
+    required for versions 16 or 17.
+
+    Returns:
+        tuple: A boolean indicating whether the upgrade process should proceed and
+               a string representing the Docker image of MySQL if applicable.
+    """
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Olive."
+        )
+        return False, ""
+
+    new_mysql_docker_image = str(config["DOCKER_IMAGE_MYSQL"])
+
+    # Do not perform manual upgrade if running v16 or v17, only for tutor v18 or later
+    if (
+        new_mysql_docker_image == "docker.io/mysql:8.0.33"
+        or new_mysql_docker_image == "docker.io/mysql:8.1.0"
+    ):
+        return False, ""
+
+    return True, new_mysql_docker_image
+
+
 PALM_RENAME_ORA2_FOLDER_COMMAND = """
 if stat '/openedx/data/ora2/SET-ME-PLEASE (ex. bucket-name)' 2> /dev/null; then
     echo "Renaming ora2 folder..."
--- a/tutor/commands/upgrade/common.py
+++ b/tutor/commands/upgrade/common.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+from typing import Optional
+
 import click
+from packaging import version
 
 from tutor import config as tutor_config
 from tutor import fmt, plugins
-from tutor.types import Config
+from tutor.types import Config, get_typed
 
 
 def upgrade_from_lilac(config: Config) -> None:
@@ -60,7 +63,7 @@ def get_mongo_upgrade_parameters(
     return mongo_version, admin_command
 
 
-def verify_tutor_version_for_mysql_upgrade(config: Config) -> tuple[bool, str]:
+def get_intermediate_mysql_upgrade(config: Config) -> Optional[str]:
     """
     Checks if a MySQL upgrade is needed based on the Tutor version and MySQL setup.
 
@@ -70,26 +73,22 @@ def verify_tutor_version_for_mysql_upgrade(config: Config) -> tuple[bool, str]:
     required for versions 16 or 17.
 
     Returns:
-        tuple: A boolean indicating whether the upgrade process should proceed and
-               a string representing the Docker image of MySQL if applicable.
+        Optional[str]: The docker image of MySQL to upgrade to or None if not applicable
     """
-    if not config["RUN_MYSQL"]:
+    if not get_typed(config, "RUN_MYSQL", bool):
         fmt.echo_info(
             "You are not running MySQL (RUN_MYSQL=false). It is your "
             "responsibility to upgrade your MySQL instance to v8.4. There is "
             "nothing left to do to upgrade from Olive."
         )
-        return False, ""
-
-    new_mysql_docker_image = str(config["DOCKER_IMAGE_MYSQL"])
-
-    if (
-        new_mysql_docker_image == "docker.io/mysql:8.0.33"
-        or new_mysql_docker_image == "docker.io/mysql:8.1.0"
-    ):
-        return False, ""
-
-    return True, new_mysql_docker_image
+        return None
+    image_tag = get_typed(config, "DOCKER_IMAGE_MYSQL", str).split(":")[-1]
+    # If latest image, we assign a constant value to invalidate the condition
+    # as we know that the latest image will always be greater than 8.1.0
+    target_version = (
+        version.Version("8.1.1") if image_tag == "latest" else version.parse(image_tag)
+    )
+    return "docker.io/mysql:8.1.0" if target_version > version.parse("8.1.0") else None
 
 
 PALM_RENAME_ORA2_FOLDER_COMMAND = """
--- a/tutor/commands/upgrade/compose.py
+++ b/tutor/commands/upgrade/compose.py
@@ -172,18 +172,20 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
     # Note that the DOCKER_IMAGE_MYSQL value is never saved, because we only save the
     # environment, not the configuration.
     tutor_env.save(context.obj.root, config)
-    context.invoke(compose.start, detach=True, services=["mysql"])
-    fmt.echo_info("Waiting for MySQL to boot...")
-    sleep(30)
+    if not is_mysql_service_ready(
+        context, config, old_mysql_docker_image.split(":")[1]
+    ):
+        return
 
     # Upgrade back to v8.4
     config["DOCKER_IMAGE_MYSQL"] = new_mysql_docker_image
     # Note that the DOCKER_IMAGE_MYSQL value is never saved, because we only save the
     # environment, not the configuration.
     tutor_env.save(context.obj.root, config)
-    context.invoke(compose.start, detach=True, services=["mysql"])
-    fmt.echo_info("Waiting for MySQL to boot...")
-    sleep(30)
+    if not is_mysql_service_ready(
+        context, config, new_mysql_docker_image.split(":")[1]
+    ):
+        return
     context.invoke(compose.stop)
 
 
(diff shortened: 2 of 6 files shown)


tutor/commands/upgrade/compose.py
```diff
@@ -165,6 +165,36 @@ def upgrade_from_quince(context: click.Context, config: Config) -> None:
...
 
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Quince."
+        )
+        return
+
+    # Revert the MySQL image first to build data dictionary on v8.1
```
regisb, line 176: Let's move this section to the v15 -> v16 upgrade. It's unnecessary to go through this step for users running quince or palm.

Danyal-Faheem: This error happens in a particular scenario when a user upgrades their tutor instance directly from v15 -> v18. This is why I had added it in the redwood upgrade. Adding it to the v15 -> v16 would also be unnecessary, as that works just fine.

DawoudSheraz: > when a user upgrades their tutor instance directly from v15 -> v18

Don't they need to run upgrade for each version even if they installed v18 when they had v15?

Danyal-Faheem: I've added it to the olive upgrade process now and added a small check to only run the code if someone is upgrading to v18 or above.

tutor/commands/upgrade/compose.py
```diff
@@ -158,6 +158,44 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Olive."
+        )
+        return
+
+    # Revert the MySQL image first to build data dictionary on v8.1
```
DawoudSheraz, line 178: All this can be moved to a util/inner function. Add docstrings to explain the need for that function.

tutor/commands/upgrade/k8s.py
```diff
@@ -164,6 +164,36 @@ def upgrade_from_olive(context: Context, config: Config) -> None:
...
+    ):
+        return
+    
+    if not config["RUN_MYSQL"]:
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Quince."
+        )
+        return
```
DawoudSheraz, line 183: nit: This check can be moved above. If MySQL is running, only then we do the MySQL related checks and then the commands/print commands that do the necessary operations.

tutor/commands/upgrade/compose.py
Danyal-Faheem: Updated.

DawoudSheraz:
@regisb FYI

tutor/commands/upgrade/common.py
```diff
@@ -60,6 +60,39 @@ def get_mongo_upgrade_parameters(
...
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Olive."
+        )
+        return False, ""
+
+    new_mysql_docker_image = str(config["DOCKER_IMAGE_MYSQL"])
+
+    # Do not perform manual upgrade if running v16 or v17, only for tutor v18 or later
```
DawoudSheraz, line 86: no need for this comment as it has been addressed in docstring.

tutor/commands/upgrade/k8s.py
```diff
@@ -165,6 +165,26 @@ def upgrade_from_olive(context: Context, config: Config) -> None:
...
 
+    upgrade_to_redwood_onwards, _ = (
+        common_upgrade.verify_tutor_version_for_mysql_upgrade(config)
+    )
+
+    if not upgrade_to_redwood_onwards:
+        return
+
+    # We need to first revert MySQL back to v8.1 to build the data dictionary on it
+    # And also to update the authentication plugin as it is disabled on v8.4
```
DawoudSheraz, line 176: nit: we can remove this comment as the message on the next lines conveys the same thing.

tutor/commands/upgrade/common.py
Danyal-Faheem: Removed the comment.

tutor/commands/upgrade/k8s.py
Danyal-Faheem: Removed the comments.

tutor/commands/upgrade/compose.py
```diff
@@ -158,6 +158,34 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+    old_mysql_docker_image = "docker.io/mysql:8.1.0"
+    click.echo(fmt.title(f"Upgrading MySQL to v{new_mysql_docker_image.split(':')[1]}"))
+    config["DOCKER_IMAGE_MYSQL"] = old_mysql_docker_image
+    # Note that the DOCKER_IMAGE_MYSQL value is never saved, because we only save the
+    # environment, not the configuration.
+    tutor_env.save(context.obj.root, config)
+    context.invoke(compose.start, detach=True, services=["mysql"])
+    fmt.echo_info("Waiting for MySQL to boot...")
+    sleep(30)
```
regisb, line 177: Is there any way to avoid that sleep statement? It's bound to be either too long or too short for many people. We could avoid that if we could run some sort of `mysql --upgrade` one-off command, instead of starting the full mysql service.

Danyal-Faheem: From these [docs](https://dev.mysql.com/doc/refman/8.0/en/mysql-upgrade.html):

```
As of MySQL 8.0.16, the MySQL server performs the upgrade tasks previously handled by mysql_upgrade.
Consequently, mysql_upgrade is unneeded and is deprecated as of that version; expect it to be removed in a future
version of MySQL. Because mysql_upgrade no longer performs upgrade tasks, it exits with status 0 unconditionally.
```

And this [mysql-8-0-16-mysql_upgrade-is-going-away](https://dev.mysql.com/blog-archive/mysql-8-0-16-mysql_upgrade-is-going-away/) as well, it is not possible to run a one off `mysql upgrade` command. MySQL handles the upgrade process itself whenever a new MySQL server is started and it detects that an upgrade process is required.

The official [guide](https://dev.mysql.com/doc/refman/8.0/en/docker-mysql-getting-started.html#docker-upgrading) on how to upgrade docker based mysql servers also states the same as above.

However, what we can do is healthcheck polling to wait until the server is ready for connections before returning similar to what is being done in the [mysql init job](https://github.com/overhangio/tutor/blob/master/tutor/templates/jobs/init/mysql.sh#L4-L15).

tutor/commands/upgrade/compose.py
```diff
@@ -158,6 +158,34 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+
+    if not upgrade_to_redwood_onwards:
+        return
+
+    # Revert the MySQL image first to build data dictionary on v8.1
+    old_mysql_docker_image = "docker.io/mysql:8.1.0"
+    click.echo(fmt.title(f"Upgrading MySQL to v{new_mysql_docker_image.split(':')[1]}"))
+    config["DOCKER_IMAGE_MYSQL"] = old_mysql_docker_image
+    # Note that the DOCKER_IMAGE_MYSQL value is never saved, because we only save the
+    # environment, not the configuration.
```
DawoudSheraz, line 173: curious: So this only keep the change in that terminal environment and not to file, correct?

tutor/commands/upgrade/compose.py
DawoudSheraz: poll check with a timeout might be better suited here. In case the service does not start after 30 seconds, there would be no point in executing the remaining operations.

DawoudSheraz (approved):

tutor/commands/upgrade/compose.py
Danyal-Faheem: Correct.

Similar to how it is being done in the MongoDB upgrade.

https://github.com/overhangio/tutor/blob/3143335d69a7f9b012551d14838ea81101101551/tutor/commands/upgrade/compose.py#L187-L190

Faraz32123:
Hello @regisb, can u please have a look at this PR.

tutor/commands/upgrade/common.py
```diff
@@ -60,6 +60,38 @@ def get_mongo_upgrade_parameters(
...
+    This method ensures that MySQL is running and determines if the upgrade
+    process should proceed based on the Tutor version. It is intended for upgrades
+    from Tutor version 15 to version 18 or later. Manual upgrade steps are not
+    required for versions 16 or 17.
+
+    Returns:
+        tuple: A boolean indicating whether the upgrade process should proceed and
+               a string representing the Docker image of MySQL if applicable.
+    """
+    if not config["RUN_MYSQL"]:
```
regisb, line 76: `get_typed(config, "RUN_MYSQL", bool)`

tutor/commands/upgrade/common.py
```diff
@@ -60,6 +60,38 @@ def get_mongo_upgrade_parameters(
...
+        fmt.echo_info(
+            "You are not running MySQL (RUN_MYSQL=false). It is your "
+            "responsibility to upgrade your MySQL instance to v8.4. There is "
+            "nothing left to do to upgrade from Olive."
+        )
+        return False, ""
+
+    new_mysql_docker_image = str(config["DOCKER_IMAGE_MYSQL"])
```
regisb, line 84: `get_typed(config, "DOCKER_IMAGE_MYSQL", str)`

tutor/commands/upgrade/common.py
```diff
@@ -60,6 +60,38 @@ def get_mongo_upgrade_parameters(
...
+
+    if (
+        new_mysql_docker_image == "docker.io/mysql:8.0.33"
+        or new_mysql_docker_image == "docker.io/mysql:8.1.0"
+    ):
+        return False, ""
+
+    return True, new_mysql_docker_image
```
regisb, line 92: Instead of returning a complex tuple of values, let's return a single value: `new_mysql_docker_image` when an upgrade is required, None when it isn't.

tutor/commands/upgrade/common.py
```diff
@@ -60,6 +60,38 @@ def get_mongo_upgrade_parameters(
     return mongo_version, admin_command
 
 
+def verify_tutor_version_for_mysql_upgrade(config: Config) -> tuple[bool, str]:
```
regisb, line 63: This function is needlessly complicated and got me really confused. What we want to do is to go through the 8.1 upgrade before the 8.4 upgrade, right? Then we only need to write:

```
def get_intermediate_mysql_upgrade(config) -> str:
    if not get_typed(config, "RUN_MYSQL", bool):
         return None
    return "docker.io/mysql:8.1.0" if get_typed(config, "DOCKER_IMAGE_MYSQL", str).startswith("docker.io/mysql:8.4") else None
```

tutor/commands/upgrade/compose.py
```diff
@@ -158,6 +158,36 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
     upgrade_mongodb(context, config, "4.2.17", "4.2")
     upgrade_mongodb(context, config, "4.4.22", "4.4")
 
+    upgrade_to_redwood_onwards, new_mysql_docker_image = (
```
regisb, line 161: This part is just too confusing. I suggest the following:

```
intermediate_mysql_docker_image = get_intermediate_mysql_upgrade(config)
if not intermediate_mysql_docker_image:
    return

click.echo(fmt.title(f"Upgrading MySQL to {intermediate_mysql_docker_image}"))
mysql_docker_image = config["DOCKER_IMAGE_MYSQL"]
config["DOCKER_IMAGE_MYSQL"] = intermediate_mysql_docker_image
context.invoke(do.commands["init"], limit="mysql")
```

Note that the mysql init task does the exact same command as `is_mysql_service_ready`, which means that we can get rid of this function.

regisb (changes requested):

tutor/commands/upgrade/common.py
Danyal-Faheem: The condition was checking against two specific MySQL versions was in the case someone jumps multiple tutor versions, e.g, tutor 17 -> 20 and the new MySQL image is on 9.0 for example, this upgrade step wouldn't run and the upgrade process would crash again. But I think that would become too complex and this solution is more elegant. 

Should we not echo the info that MySQL is not running and it is up to the user to upgrade it?

tutor/commands/upgrade/compose.py
Danyal-Faheem: Huh, I didn't know we could call init commands directly in the tutor codebase. Let me try this out.

tutor/commands/upgrade/common.py
regisb: > The condition was checking against two specific MySQL versions was in the case someone jumps multiple tutor versions, e.g, tutor 17 -> 20 and the new MySQL image is on 9.0 for example

In that case, the condition should be: 

    import packaging
    target_version = packaging.version(get_typed(config, "DOCKER_IMAGE_MYSQL", str))
    return "docker.io/mysql:8.1.0" if target_version > packaging.version("8.1.0") else None

Danyal-Faheem: > target_version = packaging.version(get_typed(config, "DOCKER_IMAGE_MYSQL", str))

This will fail if the image tag is set to latest. [docs reference](https://packaging.pypa.io/en/stable/version.html#packaging.version.InvalidVersion)

regisb: Obviously. So what should we do in that case?

Danyal-Faheem: I've added a default check for latest as it would always be greater than 8.1 for our case.

tutor/commands/upgrade/compose.py
Danyal-Faheem: This worked out perfectly, thanks for the heads up.

regisb (approved):
lgtm! thanks for bearing with my many comments.

PR #1103, opened 2024-07-26
fix: warning message indentation in upgrade

regisb:
The warning message printed during `tutor local upgrade` was incorrectly indented.

I decided not to write a changelog entry, given how minor this change is.


DawoudSheraz (approved):
nit: while it is not docs, either docs or chore would be a better commit type here (when merging).

PR #1105, opened 2024-08-02
feat: Implement kill command for local and development environments

CodeWithEmad:
In a development setting, gracefully stopping containers can be time-consuming and unnecessary. This introduces a command to kill containers directly, improving efficiency.

### DIFF
--- a/changelog.d/20240725_212933_codewithemad_kill.md
+++ b/changelog.d/20240725_212933_codewithemad_kill.md
@@ -1,0 +1,1 @@
+- [Improvement] New `kill` command for local and development environments. (by @CodeWithEmad)
--- a/tests/commands/test_images.py
+++ b/tests/commands/test_images.py
@@ -1,7 +1,7 @@
 from unittest.mock import Mock, patch
 
 from tests.helpers import PluginsTestCase, temporary_root
-from tutor import images, plugins, utils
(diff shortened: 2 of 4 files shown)


changelog.d/20240725_212933_codewithemad_kill.md
```diff
@@ -0,0 +1 @@
+- [Improvement] New `kill` command for local and development environments. (by @CodeWithEmad)
```
DawoudSheraz, line 1: shouldn't this be better off in documentation/troubleshooting instead of a new command? While I understand the need, I don't feel like the tutor codebase would be the right place.

regisb:
I see little benefit of this over `tutor local/dev dc kill`. It's not a command that we want to encourage people to run, so I think it's better to keep it "hidden".

changelog.d/20240725_212933_codewithemad_kill.md
CodeWithEmad: Well yes. It can be in docs too, but the point is to have a quick command to stop the environment without wasting time. Each stop can take up to 10 - 15 seconds and goes up depending on your plugins and machine.

CodeWithEmad:
You're right @regisb. I close this one.

PR #1108, opened 2024-08-09
fix: work around outline generation bug

arbrandes:
This works around an upstream bug where the course outline is not
generated when a course is imported, resulting in a 500 error when
trying to access the imported course in the LMS.

### DIFF
--- a/tutor/commands/jobs.py
+++ b/tutor/commands/jobs.py
@@ -173,6 +173,9 @@ echo "INFO: Will import course data at: $course_root" && echo
 # Import into CMS
 python ./manage.py cms import ../data "$course_root"
 
+# Work around bug where outline is not generated
+python ./manage.py cms update_course_outline course-v1:OpenedX+DemoX+DemoCourse
+
 # Re-index courses
 ./manage.py cms reindex_course --all --setup"""
     yield ("cms", template)


tutor/commands/jobs.py
```diff
@@ -173,6 +173,9 @@ def importdemocourse(
 # Import into CMS
 python ./manage.py cms import ../data "$course_root"
 
+# Work around bug where outline is not generated
+python ./manage.py cms update_course_outline course-v1:OpenedX+DemoX+DemoCourse
```
DawoudSheraz, line 177: Is it impacting all the courses or only under certain circumstances? If the outline is not generated in initial import, does it not generate when a user navigates to LMS  or edits something on CMS?

DawoudSheraz: @arbrandes Hi, catching up on ☝🏽

DawoudSheraz: @arbrandes Hi, pinging for ☝🏽 . Thanks

arbrandes: That's a good question.  I would guess this impacts all courses that are imported using the same command, yes, but I haven't actually tested it.

DawoudSheraz: @arbrandes Were you able to test ☝🏽 ? Thanks

DawoudSheraz: If it happens for every course, how does this fix the larger problem?

regisb, line 177: There's another issue with the proposed fix, which is that it does not work when importing courses other than course-v1:OpenedX+DemoX+DemoCourse.

AFAIU this is not a bug in Tutor, but in edx-platform, and it should be fixed there. We can cherry-pick the change here if necessary.

arbrandes: It doesn't. :)

DawoudSheraz: Right. Do we really need this PR then? As Régis mentioned below, if this is an upstream issue, we should aim to fix that. Thanks

brian-smith-tcril:
Is there an upstream issue on the `edx-platform` repo for this? I found some things that seem *similar* when searching there https://github.com/openedx/edx-platform/pulls?q=is%3Apr+is%3Aopen+course+outline but nothing that matched 1:1

ormsbee:
I believe this is actually a Tutor issue. Specifically, Tutor is calling the `import` management command [with production settings instead of dev settings](https://github.com/openedx/edx-platform/issues/36170#issuecomment-2744282360).

DawoudSheraz:
> I believe this is actually a Tutor issue. Specifically, `tutor dev do importdemocourse` is calling the `import` management command [with production settings instead of dev settings](https://github.com/openedx/edx-platform/issues/36170#issuecomment-2744282360).

Nice. It seems this is true for all the dev do jobs, not just this. I am testing and verifying this. Once done, I will create an appropriate fix PR.

DawoudSheraz:
Created https://github.com/overhangio/tutor/pull/1213 for fixup

PR #1112, opened 2024-08-16
feat: add `patches show ...` command

regisb:
I found myself in a situation where I wasn't sure if a patch was correctly added by the Indigo plugin. To troubleshoot that issue, I wanted to print that rendered patch. I thought it would make a nice addition to the CLI.


regisb:
I didn't manage to write a test with an actual patch, so I added a test for an empty patch... (mocks are hard)

Issue #1114, opened 2024-08-20
Ubuntu 22.04 upgrade


DawoudSheraz:
- Ensure CI uses the latest ubuntu
- Ensure dockerfile is updated to use ubuntu ~~24.04~~ 22.04
- Ensure any documentation/references are updated 
- Ensure the change works as expected post-upstream release

See https://github.com/overhangio/tutor/issues/1113 for context

DawoudSheraz:
Draft PR https://github.com/overhangio/tutor/pull/1117

DawoudSheraz:
While the PR is not complete yet, adding the details of the changes so far:

- The user with ID 1000 is now shipped with ubuntu and we cannot use 1000 uid in Dockerfile. That id has been updated to 1001.
  -  https://askubuntu.com/questions/1513927/ubuntu-24-04-docker-images-now-includes-user-ubuntu-with-uid-gid-1000
  - https://bugs.launchpad.net/cloud-images/+bug/2005129
- python-openssl has been updated to python3-openssl

DawoudSheraz:
There are issues with xmlsec import as part of SAML3. The following error is raised when building the image. It is unclear of the exact cause, though various online resources mention uninstalling and re-installing specific versions of xmlsec and its dependencies.

- https://github.com/SAML-Toolkits/python3-saml/issues/411
- https://github.com/xmlsec/python-xmlsec/issues/320
- https://github.com/SAML-Toolkits/python3-saml/issues/374
- https://github.com/xmlsec/python-xmlsec/issues/254
- 

```=> ERROR [production 19/37] RUN ./manage.py lms --settings=tutor.i18n pull_plugin_translations --verbose --repository='openedx/openedx-translation  11.7s
------
 > [production 19/37] RUN ./manage.py lms --settings=tutor.i18n pull_plugin_translations --verbose --repository='openedx/openedx-translations' --revision='main':
10.74 Traceback (most recent call last):
10.74   File "/openedx/edx-platform/./manage.py", line 103, in <module>
10.74     startup.run()
10.74   File "/openedx/edx-platform/lms/startup.py", line 20, in run
10.74     django.setup()
10.74   File "/openedx/venv/lib/python3.11/site-packages/django/__init__.py", line 24, in setup
10.75     apps.populate(settings.INSTALLED_APPS)
10.75   File "/openedx/venv/lib/python3.11/site-packages/django/apps/registry.py", line 116, in populate
10.75     app_config.import_models()
[...]

PR #1117, opened 2024-09-03
feat: update ubuntu

DawoudSheraz:
### Description

Upgrade Open edX image to use Ubuntu ~~24.04~~ 22.04 (See https://github.com/overhangio/tutor/issues/1114#issuecomment-2429082353 for context on 24.04 -> 22.04 switch).

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -1,12 +1,9 @@
 # syntax=docker/dockerfile:1
 ###### Minimal image with base system requirements for most stages
-FROM docker.io/ubuntu:24.04 AS minimal
+FROM docker.io/ubuntu:22.04 AS minimal
 LABEL maintainer="Overhang.io <contact@overhang.io>"
 
 ENV DEBIAN_FRONTEND=noninteractive
-
-# Delete default UID=1000 `ubuntu` user to ensure we can use id 1000 for app user
-RUN userdel -r ubuntu
 
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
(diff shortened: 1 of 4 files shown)


tutor/templates/build/openedx/Dockerfile
```diff
@@ -138,9 +141,10 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     apt update \
     && apt install -y gettext gfortran graphviz graphviz-dev libffi-dev libfreetype6-dev libgeos-dev libjpeg8-dev liblapack-dev libmysqlclient-dev libpng-dev libsqlite3-dev libxmlsec1-dev lynx mysql-client ntp pkg-config rdfind
 
+
 # From then on, run as unprivileged "app" user
 # Note that this must always be different from root (APP_USER_ID=0)
-ARG APP_USER_ID=1000
+ARG APP_USER_ID=1001
```
regisb, line 147: This has pretty big implications. For one, the UID on the host is very often 1000, so changing the internal UID is likely to have more implications in terms of ownership, especially for mounted volumes.  Thus, I think we should avoid running as a different UID. Instead, can we try out this suggestion to delete the existing ubuntu user? https://askubuntu.com/a/1515958

DawoudSheraz: Sure. I initially thought of doing that but it looked weird. Some online suggestions were to use the next available uid which was 1001 in this case.

DawoudSheraz: I have updated the PR, please take another look.

DawoudSheraz: @regisb Hi, a reminder on this. Thanks

kdmccormick:
2U found an issue with 24.04 (which I don't yet understand) that led them to pin edx-platform tests back to 22.04: https://github.com/openedx/edx-platform/pull/35635

I'm going to ask them about it tomorrow morning at Maintenance WG. Perhaps we hold off until edx-platform is back to testing on 24.04?

DawoudSheraz:
> 2U found an issue with 24.04 (which I don't yet understand) that led them to pin edx-platform tests back to 22.04: [openedx/edx-platform#35635](https://github.com/openedx/edx-platform/pull/35635)
> 
> I'm going to ask them about it tomorrow morning at Maintenance WG. Perhaps we hold off until edx-platform is back to testing on 24.04?

Sure. @regisb FYI

DawoudSheraz:
@kdmccormick Hi, I have updated the PR to use Ubuntu 22.04. Please take another look. Thanks

kdmccormick (approved):
Thanks @DawoudSheraz . The diff looks good to me.

I'm going to manually test the build later today when my internet connection is better, but if you have already manually tested it and feel solid about the upgrade, then feel free to merge first.

PR #1119, opened 2024-09-04
Build OCI-compatible image cache

### LINKED ISSUE #1118: 400 Bad request when building images with registry cache to ECR
**Bug description**
When building images with registry cache, and when the cache is in AWS' ECR, building results in 400 Bad request error.

```
------
 > importing cache manifest from 2xxxxxxx4.dkr.ecr.us-east-1.amazonaws.com/myrepo/openedx:202409032122-cache:
------
------
 > exporting cache to registry:
------
error: failed to solve: rpc error: code = Unknown desc = error writing manifest blob: failed commit on ref "sha256:dbddfa9dc32dd219457de9c9b1534f667024da6e36e5a085f7f52540fc9ef9af": unexpected status from PUT request to https://2xxxxxxx4.dkr.ecr.us-east-1.amazonaws.com/v2/myrepo/openedx/manifests/202409032122-cache: 400 Bad Request
Error: Command failed with status 1: docker buildx build --tag=2xxxxxxx4.dkr.ecr.us-east-1.amazonaws.com/myrepo/openedx:202409032122 --output=type=docker --builder=container --cache-from=type=registry,ref=2xxxxxxx4.dkr.ecr.us-east-1.amazonaws.com/myrepo/openedx:202409032122-cache --cache-to=type=registry,mode=max,ref=2xxxxxxx4.dkr.ecr.us-east-1.amazonaws.com/myrepo/openedx:202409032122-cache /home/myuser/mytutorroot/env/build/openedx
Error building openedx

```
**How to reproduce**

- Create an ECR registry in AWS
- Declare the registry url in the DOCKER_REGISTRY variable
- Log in to ECR
- Run `tutor images build openedx --cache-to-registry`

**Environment**

Tested with Tutor 17.0.6

**Additional context**

As per [this document](https://aws.amazon.com/es/blogs/containers/announcing-remote-cache-support-in-amazon-ecr-for-buildkit-clients/), adding `image-manifest=true` context to the docker buildx command fixed the issue. 

> The new context key introduced in Buildkit 0.12 here is image-manifest. Setting this key’s value to true lets you now store an OCI-compatible version of a remote cache in the registry.

If it doesn't break the official repos (which I cannot test), I would like to add this key to the [image build args](https://github.com/overhangio/tutor/blob/46b401690c812448c17b2cd91f90cf765a7a9678/tutor/commands/images.py#L230).

angonz:
Fixes #1118 

Add image-manifest=true to build context to support OCI-compatible image cache.
This allows using the registry cache in other registries, like AWS ECR.

### DIFF
--- a/tutor/commands/images.py
+++ b/tutor/commands/images.py
@@ -227,7 +227,7 @@ def build(
                 image_build_args.append(f"--cache-from=type=registry,ref={tag}-cache")
             if cache_to_registry:
                 image_build_args.append(
-                    f"--cache-to=type=registry,mode=max,ref={tag}-cache"
+                    f"--cache-to=image-manifest=true,type=registry,mode=max,ref={tag}-cache"
                 )
 
             # Build contexts


DawoudSheraz (approved):

regisb:
:thinking: According to my tests this change should not break our CI. But I'm not sure what's the impact, to be honest. I fail to understand this part of the docs: https://docs.docker.com/build/cache/backends/

> By default, the OCI media type generates an image index for the cache image. Some OCI registries, such as Amazon ECR, don't support the image index media type: application/vnd.oci.image.index.v1+json. If you export cache images to ECR, or any other registry that doesn't support image indices, set the image-manifest parameter to true to generate a single image manifest instead of an image index for the cache image.

I am tempted not to merge this change, for the following reasons:

1. I don't really have a clue what this does.
2. It makes the build command more complicated.
3. You can achieve the same result by implementing the DOCKER_BUILD_COMMAND filter https://docs.tutor.edly.io/reference/api/hooks/catalog.html#tutor.hooks.Filters.DOCKER_BUILD_COMMAND

angonz:
Hi Régis!
To be honest, I stopped using the registry cache because it only slows the image building process. Registry cache seems to be tied to the image tag. As we create a new tag for every build, cache is always regenerated. Btw, I didn't know of the DOCKER_BUILD_COMMAND setting. That might be a better alternative.

DawoudSheraz:
Given the conversations above, are we good to close this PR @angonz?

angonz:
Yes.

fghaas:
I opened PR #1161 not having checked for closed issues (sorry!), and @regisb asked me to look at this one and comment, so here goes:

> 2. It makes the build command more complicated.
> 3. You can achieve the same result by implementing the DOCKER_BUILD_COMMAND filter https://docs.tutor.edly.io/reference/api/hooks/catalog.html#tutor.hooks.Filters.DOCKER_BUILD_COMMAND

This is pushing complexity to the downstream user, which I'd argue is always a bad idea. "We won't merge a one-liner because you can just as well write a plugin" is not a compelling proposition in my book. :)

However, to illustrate why this should go in (not as a feature addition but rather a bug fix), let's address this question:

> 1. I don't really have a clue what this does.

Buckle up, because it's about to get ugly. :)

BuildKit had a long-standing issue (https://github.com/moby/buildkit/issues/2251) which meant that the cache image index it was generating was essentially proprietary and not compliant with the relevant spec.

In the comments on that issue, it was proposed that OCI-compliant manifests could be enabled with `oci-manifest=true`. However, when the feature landed in https://github.com/moby/buildkit/pull/3724, the option name ended up being `image-manifest`. 

So, if (and only if) we specify `image-manifest=true`, BuildKit does the correct (OCI-compliant) thing and creates an OCI-compliant rather than a proprietary manifest format. And that's the proper way to support cached image builds across multiple registry providers.

Also, on this one:

> To be honest, I stopped using the registry cache because it only slows the image building process. Registry cache seems to be tied to the image tag. As we create a new tag for every build, cache is always regenerated.

I cannot corroborate the "slows the image building process" part. And yes, we do have to build from scratch every time we create a new tag. However, it is very helpful to be able to use cached builds for multiple builds of the same image tag, which tends to happen during stages of code review.

regisb:
Thanks to you both @angonz and @fghaas for bearing with me. I agree that we should merge this change. 

@angonz your PR did not include a changelog entry; @fghaas your PR has a changelog entry, but no `"by @fghaas"` signature. It's by no means mandatory to include a signature, but if you do, then I think it would make sense that both your names should be mentioned.

fghaas:
>Your explanations are very clear and helpful, thanks a lot.

I hesitate to take credit for the explanation, as it was @kcollasarundell who very helpfully explained this to *me,* all credit should go to him, and any misinterpretations of his explanation are mine. :)

> @angonz your PR did not include a changelog entry; @fghaas your PR has a changelog entry, but no "by @fghaas" signature. It's by no means mandatory to include a signature, but if you do, then I think it would make sense that both your names should be mentioned.

Sure. I have updated #1161 with an extended changelog entry, and have also added `Co-authored-by` and `Fixes` lines to the commit message for that patch.

angonz:
Thank you all! I'm very happy to see that multiple visions on one problem end up in something much better.

PR #1124, opened 2024-09-26
feat: create volume for private LMS/CMS media assets

ormsbee:
Content Libraries will store asset files like images, transcripts, and PDFs using Learning Core–which in turn uses django-storages and expects some backing file store. These files should NOT be directly accessible via the browser, both because of access policies and the fact that the filenames will not be meaningful by themselves and must be translated by app logic. For details see:

* https://github.com/openedx/openedx-learning/blob/main/docs/decisions/0015-serving-static-assets.rst
* https://github.com/openedx/openedx-learning/blob/main/openedx_learning/apps/authoring/contents/models.py
* https://github.com/openedx/openedx-learning/blob/main/openedx_learning/apps/authoring/components/models.py

The existing /data/openedx-media -> /openedx/media mount is publicly accessible by browser, and so is not appropriate for this purpose. This commit creates a parallel /data/openedx-media-private -> /openedx/media-private volume mapping instead.

This commit also creates configuration needed for openedx-learning to point to this new directory.

This commit does NOT currently add support for this in k8s, as that will require changes to the minio plugin.


tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -246,5 +246,14 @@
     "user": None,
 }
 
+LEARNING_CORE = {
```
ormsbee, line 249: Actually, I should probably make this `OPENEDX_LEARNING`, to reduce confusion...

bradenmacdonald:
✅ I tested this using [these instructions](https://github.com/openedx/edx-platform/pull/35557#issuecomment-2381052162) and it's working well.

changelog.d/20240926_151223_dave_private_media.md
```diff
@@ -0,0 +1 @@
+[Feature] Create a new /data/openedx-media-private volume to store media files that are not publicly accessible via the browser, and create configuration that Learning Core (openedx-learning) will use to access this storage space. Learning Core will use a sub-directory of /data/openedx-media-private to store Content Library file uploads, but other apps can create their own sub-directories for their own use. (by @ormsbee)
```
kdmccormick, line 1: changelog is a bulleted list
```suggestion
- [Feature] Create a new /data/openedx-media-private volume to store media files that are not publicly accessible via the browser, and create configuration that Learning Core (openedx-learning) will use to access this storage space. Learning Core will use a sub-directory of /data/openedx-media-private to store Content Library file uploads, but other apps can create their own sub-directories for their own use. (by @ormsbee)
```

kdmccormick (approved):

regisb:
Here's the corresponding MinIO issue: https://github.com/overhangio/tutor-minio/issues/48

PR #1128, opened 2024-10-01
feat: do not run celery workers in development

regisb:
In development, edx-platform runs with CELERY_ALWAYS_EAGER=True, which
means that lms-worker and cms-worker never catch celery tasks!

This change was heavily inspired by: https://github.com/overhangio/tutor/pull/1041


arbrandes (approved):
Looks great to me! 👍🏼

DawoudSheraz (approved):

Issue #1130, opened 2024-10-03
Add support for running multiple Celery queues


Ian2012:
**Is your feature request related to a problem? Please describe.**

Open edX already has implemented solutions for running multiple celery queues such as the old known `high` and `high_mem` [default queues](https://github.com/openedx/edx-platform/blob/9ff0dac55cb30c21f7973d503c40f2755671f2de/lms/envs/common.py#L2894-L2905) however, tutor operates by using a single worker deployment with a default pool (`prefork`) and this is not always as performant as desired as that pool is designed for CPU intensive tasks such as the ones related to grading, while I/O bound tasks such as the ones used in Aspects would benefit from having a `gevent` pool which used green threads aka virtual threads to accomplish high levels of concurrency. This has already been tested and implemented in [tutor-contrib-celery](https://github.com/eduNEXT/tutor-contrib-celery) and the benefits have been notorious as the resources are better handled and I/O bound tasks are resolved pretty fast.

**Describe the solution you'd like**

Allow tutor users to configure multiple Celery deployments with specialized settings for each queue's tasks. With defaults matching what's currently expected to run on each queue.

**Describe alternatives you've considered**

Developing a separate plugin for Celery: [tutor-contrib-celery](https://github.com/eduNEXT/tutor-contrib-celery) however we think the community would benefit from having a standard way to set Celery and also don't need to build a custom openedx images with those requirements: https://github.com/openedx/edx-platform/pull/35591

**Additional context**

https://celery.school/celery-worker-pools

regisb:
Hey @Ian2012! I'm glad you are opening this issue today as there are multiple conversations going on right now about Celery:

https://github.com/overhangio/tutor/issues/1126
https://github.com/overhangio/tutor/pull/1010
https://github.com/overhangio/tutor/pull/1128

As usual, I'd like to resolve these issues by:

1. Improving the default settings for end users.
2. Making it possible to easily customize these default settings.

Let's address these two items in reverse order:

## Customization

Let's assume that we provide a mechanism in Tutor core that makes it possible to customize the celery command exactly as you need. You are then also able to run any celery worker by adding containers to the docker-compose/k8s override files. I don't have a perfect solution to this problem just yet, but I'm working on it. Assuming this solution exists, it should enable you to customize the celery workers as you see fit, right?

[...]

PR #1131, opened 2024-10-04
POC feat: add support for celery multiqueue

Ian2012:
### Description

This PR allows to deploy multiple Celery deployments with different settings and queues for each one:

### DIFF
--- a/tutor/hooks/catalog.py
+++ b/tutor/hooks/catalog.py
@@ -514,6 +514,9 @@ class Filters:
     #: :param str file_path: The path to the file being checked.
     IS_FILE_RENDERED: Filter[bool, [str]] = Filter()
 
+    #: List of workers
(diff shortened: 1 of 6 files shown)


tutor/hooks/catalog.py
```diff
@@ -514,6 +514,9 @@ def your_filter_callback(some_data):
     #: :param str file_path: The path to the file being checked.
     IS_FILE_RENDERED: Filter[bool, [str]] = Filter()
 
+    #: List of workers
+    CELERY_WORKERS_CONFIG: Filter[dict[str, dict[str, Any]], []] = Filter()
```
regisb, line 518: I'm not in favor of a filter that would be used to create extra containers. There is already another mechanisms to do that, which is patching the docker-compose/deployments.yml files. Let's avoid creating another one.

Let's continue this conversation in the corresponding issue (#1130).

PR #1134, opened 2024-10-08
feat: add a filter to define celery workers command

Ian2012:
### Description

This PR implements two new similar filters to define the command for the celery workers.

### DIFF
--- a/tutor/plugins/openedx.py
+++ b/tutor/plugins/openedx.py
@@ -142,3 +142,45 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
 hooks.Filters.ENV_TEMPLATE_VARIABLES.add_item(
     ("iter_mounted_directories", iter_mounted_directories)
 )
+
+
+hooks.Filters.LMS_WORKER_COMMAND.add_items(
+    [
+        "celery",
+        "--app=lms.celery",
+        "worker",
+        "--loglevel=info",
+        "--hostname=edx.lms.core.default.%h",
+        "--queues=edx.lms.core.default,edx.lms.core.high,edx.lms.core.high_mem",
+        "--max-tasks-per-child=100",
+    ]
+)
+
+
+hooks.Filters.CMS_WORKER_COMMAND.add_items(
+    [
+        "celery",
+        "--app=cms.celery",
+        "worker",
+        "--loglevel=info",
+        "--hostname=edx.cms.core.default.%h",
+        "--queues=edx.cms.core.default,edx.cms.core.high,edx.cms.core.low",
+        "--max-tasks-per-child=100",
+    ]
+)
+
+
+def iter_cms_celery_parameters() -> t.Iterator[str]:
+    yield from hooks.Filters.CMS_WORKER_COMMAND.iterate()
+
+
+def iter_lms_celery_parameters() -> t.Iterator[str]:
+    yield from hooks.Filters.LMS_WORKER_COMMAND.iterate()
+
+
+hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
+    [
+        ("iter_cms_celery_parameters", iter_cms_celery_parameters),
+        ("iter_lms_celery_parameters", iter_lms_celery_parameters),
+    ]
+)
--- a/tutor/templates/local/docker-compose.prod.yml
+++ b/tutor/templates/local/docker-compose.prod.yml
@@ -31,7 +31,8 @@ services:
     environment:
       SERVICE_VARIANT: lms
       DJANGO_SETTINGS_MODULE: lms.envs.tutor.production
-    command: celery --app=lms.celery worker --loglevel=info --hostname=edx.lms.core.default.%%h --max-tasks-per-child=100 --exclude-queues=edx.cms.core.default
+    command: {% for value in iter_lms_celery_parameters() %}
+      - "{{value}}"{% endfor %}
     restart: unless-stopped
     volumes:
       - ../apps/openedx/settings/lms:/openedx/edx-platform/lms/envs/tutor:ro
@@ -50,7 +51,8 @@ services:
     environment:
       SERVICE_VARIANT: cms
       DJANGO_SETTINGS_MODULE: cms.envs.tutor.production
-    command: celery --app=cms.celery worker --loglevel=info --hostname=edx.cms.core.default.%%h --max-tasks-per-child 100 --exclude-queues=edx.lms.core.default
+    command: {% for value in iter_lms_celery_parameters() %}
+      - "{{value}}"{% endfor %}
     restart: unless-stopped
     volumes:
       - ../apps/openedx/settings/lms:/openedx/edx-platform/lms/envs/tutor:ro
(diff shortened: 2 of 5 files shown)


tutor/templates/build/openedx/Dockerfile
```diff
@@ -99,7 +99,9 @@ RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
     # Use redis as a django cache https://pypi.org/project/django-redis/
     django-redis==5.4.0 \
     # uwsgi server https://pypi.org/project/uWSGI/
-    uwsgi==2.0.24
+    uwsgi==2.0.24 \
+    # gevent for celery
+    gevent==24.2.1
```
jfavellar90, line 104: @Ian2012 please remove this change as discussed in the large instances meeting

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,71 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
 hooks.Filters.ENV_TEMPLATE_VARIABLES.add_item(
     ("iter_mounted_directories", iter_mounted_directories)
 )
+
+
+@hooks.Filters.LMS_WORKER_COMMAND.add(priority=hooks.priorities.HIGH)
+def _add_default_lms_worker_parameters(
+    worker_configs: dict[str, str]
```
regisb, line 149: Let's make this a `list[str]`, such that:

1. arguments are properly ordered
2. we can support value-less options, such as `--without-gossip`.

Also, `worker_configs` should be renamed to `command`.

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,71 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
...
+) -> dict[str, str]:
+    worker_configs = {
```
regisb, line 151:
```
worker_configs.append([
    "celery",
    "worker",
    "--app=lms.celery",
    "--loglevel=info",
    "--hostname=edx.lms.core.default.%%h",
    ...
])
return workers_config
```

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,71 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
...
+    }
+    return worker_configs
+
+
+@hooks.lru_cache
+def get_cms_celery_parameters() -> dict[str, str]:
+    """
+    This function is cached for performance.
+    """
+    return hooks.Filters.CMS_WORKER_COMMAND.apply({})
```
regisb, line 178: We don't need both `iter_*` and `get_*` functions. Instead, let's just have:

```
@hooks.lru_cache
def iter_cms_celery_parameters():
    yield from hooks.Filters.CMS_WORKER_COMMAND.iterate()

hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
    [
        ("iter_cms_celery_parameters", iter_cms_celery_parameters),
        ("iter_lms_celery_parameters", iter_lms_celery_parameters),
    ]
)
```

regisb (changes requested):

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,63 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
...
+@hooks.Filters.LMS_WORKER_COMMAND.add(priority=hooks.priorities.HIGH)
+def _add_default_lms_worker_parameters(
+    command: list[str]
+) -> list[str]:
+    command = [
+        "celery",
+        "worker",
+        "--app=lms.celery",
+        "--loglevel=info",
+        "--hostname=edx.lms.core.default.%%h",
```
regisb, line 156: Actually, this should be "%h". The extra "%" is a [very old artefact](https://github.com/overhangio/tutor/commit/e357e763d703a4c8ca0b3f7a79a7be675b296297) from the time I copied the command string from ansible templates. As a result, celery workers have an incorrect name, such as "celery@edx.lms.core.default.%b92a7a891007". Notice the extra "%".

This is the right time to fix this 7-year-old bug :)

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,63 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
 hooks.Filters.ENV_TEMPLATE_VARIABLES.add_item(
     ("iter_mounted_directories", iter_mounted_directories)
 )
+
+
+@hooks.Filters.LMS_WORKER_COMMAND.add(priority=hooks.priorities.HIGH)
+def _add_default_lms_worker_parameters(
+    command: list[str]
+) -> list[str]:
+    command = [
```
regisb, line 151: We should use `append`, because we don't want to override any option set before. Actually, we don't need to add a callback function, now that we have a list command arguments:

```
hooks.Filters.LMS_WORKER_COMMAND.add_items([
   "celery",
    "worker",
    "--app=lms.celery",
    ...
])
```

tutor/plugins/openedx.py
```diff
@@ -142,3 +142,63 @@ def is_directory_mounted(image_name: str, dirname: str) -> bool:
...
+        "--hostname=edx.cms.core.default.%%h",
+        "--queues=edx.cms.core.default,edx.cms.core.high,edx.cms.core.low",
+        "--max-tasks-per-child=100",
+    ]
+    return command
+
+
+@hooks.lru_cache
+def iter_cms_celery_parameters() -> t.Iterator[str]:
+    """
```
regisb, line 181: These docstrings are useless because they duplicate the mypy signature, let's remove them.

tutor/templates/local/docker-compose.prod.yml
```diff
@@ -31,7 +31,7 @@ services:
     environment:
       SERVICE_VARIANT: lms
       DJANGO_SETTINGS_MODULE: lms.envs.tutor.production
-    command: celery --app=lms.celery worker --loglevel=info --hostname=edx.lms.core.default.%%h --max-tasks-per-child=100 --exclude-queues=edx.cms.core.default
+    command:{% for value in iter_lms_celery_parameters() %} {{value}}{% endfor %}
```
regisb, line 34: Let's take the opportunity to convert this to a list, just like in k8s/deployments.yml.

regisb (changes requested):

regisb:
Please also add a changelog entry.

tutor/templates/local/docker-compose.prod.yml
```diff
@@ -50,7 +51,7 @@ services:
     environment:
       SERVICE_VARIANT: cms
       DJANGO_SETTINGS_MODULE: cms.envs.tutor.production
-    command: celery --app=cms.celery worker --loglevel=info --hostname=edx.cms.core.default.%%h --max-tasks-per-child 100 --exclude-queues=edx.lms.core.default
+    command:{% for value in iter_cms_celery_parameters() %} {{value}}{% endfor %}
```
regisb, line 54: I think you forgot to update this one as well.

regisb (approved):
Awesome!

Ian2012:
@regisb can this be backported to sumac?

regisb:
It will be, because this PR was merged in the "master" (now "release") branch.

PR #1140, opened 2024-10-17
fix: run MySQL 8.1 as a separate container during upgrade from Olive to Redwood

Danyal-Faheem:
This PR is a fix for the issues brought forward in #1102.

We do this because MySQL 8.1 does not have the `--mysql-native-password` option and when upgrading from tutor v15 -> v18, this container would crash as the option does not exist.

We have this option turned on for backwards compatibility from tutor v18.1.1 onwards as users who have upgraded all the way from MySQL v5.7 -> v8.4 have their original users created using the mysql_native_password authentication plugin. See #1090 for more details.

After this PR is merged in nightly, we should make a backport to master as well coupled with the changes already made in #1102.

### DIFF
--- a/tutor/commands/upgrade/compose.py
+++ b/tutor/commands/upgrade/compose.py
@@ -171,7 +171,8 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
     # the upgrade order of 5.7 -> 8.1 -> 8.4
     # Use the mysql-8.1 context so that we can clear these filters later on
     with hooks.Contexts.app("mysql-8.1").enter():
-        hooks.Filters.ENV_PATCHES.add_item(
+        hooks.Filters.ENV_PATCHES.add_items(
+            [
             (
                 "local-docker-compose-services",
                 """
@@ -184,9 +185,7 @@ mysql-8.1:
     --collation-server=utf8mb3_general_ci
     --binlog-expire-logs-seconds=259200
     """,
-            )
-        )
-        hooks.Filters.ENV_PATCHES.add_item(
+            ),
             (
                 "local-docker-compose-jobs-services",
                 """
@@ -195,7 +194,7 @@ mysql-8.1-job:
   depends_on: {{ [("mysql-8.1", RUN_MYSQL)]|list_if }}
         """,
             )
-        )
+            ])
         hooks.Filters.CONFIG_DEFAULTS.add_item(("MYSQL_HOST", "mysql-8.1"))
 
         hooks.Filters.CLI_DO_INIT_TASKS.add_item(
--- a/tutor/commands/upgrade/compose.py
+++ b/tutor/commands/upgrade/compose.py
@@ -166,13 +166,15 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
         return
 
     click.echo(fmt.title(f"Upgrading MySQL to {intermediate_mysql_docker_image}"))
-    
+
     # We start up a mysql-8.1 container to build data dictionary to preserve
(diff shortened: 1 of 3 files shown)


tutor/commands/upgrade/compose.py
```diff
@@ -166,16 +167,53 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+            (
+                "local-docker-compose-services",
+                """
+mysql-8.1:
+  extends: mysql
+  image: docker.io/mysql:8.1.0
+  command: >
+    mysqld
+    --character-set-server=utf8mb3
+    --collation-server=utf8mb3_general_ci
```
DawoudSheraz, line 184: curious: What happens here if someone has already changed the connection strings on their Mysql57?

tutor/commands/upgrade/compose.py
```diff
@@ -166,16 +167,53 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+            (
+                "local-docker-compose-jobs-services",
+                """
+mysql-8.1-job:
+  image: docker.io/mysql:8.1.0
+  depends_on: {{ [("mysql-8.1", RUN_MYSQL)]|list_if }}
+        """,
+            )
+        )
+        hooks.Filters.CONFIG_DEFAULTS.add_item(("MYSQL_HOST", "mysql-8.1"))
```
DawoudSheraz, line 199: nit: Can't add_items be used for hooks.Filters.ENV_PATCHES here?

tutor/commands/upgrade/compose.py
```diff
@@ -166,16 +167,53 @@ def upgrade_from_olive(context: click.Context, config: Config) -> None:
...
+    context.invoke(compose.stop, services=["mysql-8.1"])
 
-    # Change the image back to v8.4
-    config["DOCKER_IMAGE_MYSQL"] = mysql_docker_image
+    # Clear the filters added for mysql-8.1 as we don't need them anymore
+    hooks.clear_all(context="app:mysql-8.1")
+
+    # Save environment and run init for mysql 8.4 to make sure MySQL is ready
     tutor_env.save(context.obj.root, config)
+    context.invoke(jobs.initialise, limit="mysql")
```
DawoudSheraz, line 216: Where is mysql version set to 8.4? Or is it 8.4 already before the code enters MySQL 8.1 context?

tutor/commands/upgrade/compose.py
Danyal-Faheem: I am not sure as I have not tested this out, however, we would want this container to be as similar to the one already being launched in tutor v17. I'm not sure we want to cater to every modification a user might have made.

tutor/commands/upgrade/compose.py
Danyal-Faheem: It is already set prior to the context when the user installs tutor v18 onwards.

tutor/commands/upgrade/compose.py
DawoudSheraz: Right.

regisb (commented):
This looks great! Can you add a changelog entry?

regisb (approved):

Danyal-Faheem:
Updated with the changelog entry as well.

tutor/commands/upgrade/compose.py
DawoudSheraz: @Danyal-Faheem

DawoudSheraz (approved):

PR #1141, opened 2024-10-17
feat: replace Elasticsearch by Meilisearch

regisb:
With this change, we get rid of Elasticsearch across all of Tutor. Instead, we run Meilisearch, which is much more lightweight in terms of memory usage. Obviously, this is a (very) breaking change. Indexing commands will be run during init, such that search should work as before.

After the edx-search PR is merged and the dependency is upgraded in edx-platform, we should remove the manual `RUN pip install ...` command.


changelog.d/20241017_125209_regis_meilisearch.md
```diff
@@ -0,0 +1 @@
+- 💥[Feature] Replace Elasticsearch by Meilisearch. Elasticsearch was both a source of complexity and high resource usage. With this change, we no longer run Elasticsearch to perform common search queries across Open edX. This includes: course discovery, courseware search and studio search. Instead, we index all these documents in a Meilisearch instance, which is much more lightweight in terms of memory consumption. (by @regisb)
```
DawoudSheraz, line 1: this wont work out of box with discovery, right? course-discovery uses ES heavily in various places. Merging this can break certain things on tutor-discovery.

tutor/env.py
```diff
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
+from copy import deepcopy
```
DawoudSheraz, line 3: why was this moved above? It was correct before in terms of import order.

tutor/templates/apps/openedx/settings/partials/common_lms.py
```diff
@@ -37,6 +37,10 @@
     "LOCATION": "staticfiles_lms",
 }
 
+# Enable search features
+FEATURES["ENABLE_COURSE_DISCOVERY"] = True
```
DawoudSheraz, line 41: Why is course-discovery enabled by default in here?

tutor/templates/dev/docker-compose.yml
```diff
@@ -32,19 +32,18 @@ services:
     ports:
         - "8001:8000"
 
+  meilisearch:
```
DawoudSheraz, line 35: Why is this not behind an IF condition?

changelog.d/20241017_125209_regis_meilisearch.md
pomegranited: Confirmed -- trying to run this with `tutor-discovery` enabled throws an Elasticsearch connection error on `./manage.py update_index --disable-change-limit`

So maybe we move the elasticsearch dependency and setup to tutor-discovery?

tutor/templates/apps/openedx/settings/partials/common_lms.py
pomegranited: Agreed -- the course search features need these feature flags, not `ENABLE_COURSE_DISCOVERY` ([ref](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/edx_search.html#lms)):
* `ENABLE_COURSEWARE_SEARCH`
* `ENABLE_DASHBOARD_SEARCH`

tutor/templates/jobs/init/cms.sh
```diff
@@ -16,3 +16,7 @@ fi
 # Create waffle switches to enable some features, if they have not been explicitly defined before
 # Copy-paste of units in Studio (highly requested new feature, but defaults to off in Quince)
 (./manage.py cms waffle_flag --list | grep contentstore.enable_copy_paste_units) || ./manage.py lms waffle_flag --create contentstore.enable_copy_paste_units --everyone
+
+# Re-index studio and courseware content
+./manage.py cms reindex_studio --experimental
+./manage.py cms reindex_course --active
```
pomegranited, line 22: Should gate these too (though not sure about which variable to use to do this?)

```suggestion
{% if ACTIVATE_MEILISEARCH %}
./manage.py cms reindex_studio --experimental
./manage.py cms reindex_course --active
{% endif %}
```

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -34,12 +34,14 @@
...
-# Elasticsearch connection parameters
-ELASTIC_SEARCH_CONFIG = [{
-  {% if ELASTICSEARCH_SCHEME == "https" %}"use_ssl": True,{% endif %}
-  "host": "{{ ELASTICSEARCH_HOST }}",
-  "port": {{ ELASTICSEARCH_PORT }},
-}]
+# Meilisearch connection parameters
+MEILISEARCH_ENABLED = True
+MEILISEARCH_URL = "{{ MEILISEARCH_URL }}"
+MEILISEARCH_PUBLIC_URL = "{{ MEILISEARCH_PUBLIC_URL }}"
```
pomegranited, line 40: This variable overrides the `MEILISEARCH_PUBLIC_URL` set above in [development.py](https://github.com/overhangio/tutor/pull/1141/files#diff-74ef659bea171686911a1025aa77a9a83fe9f602a3f3fcf201a19c37d8fb2380R18), so tutor dev meilisearch's public URL is missing the `:7700` port.

So I think it should be removed from here, and explicitly set in the lms + cms `production.py` like it is in `development.py`.

```suggestion
```

pomegranited (changes requested):
@regisb This is looking great -- just a couple of minor issues when I used this in my tutor dev env.

Am [deploying a sandbox](https://github.com/openedx/edx-platform/pull/35689) to test the k8s setup, will report back with any issues there.

changelog.d/20241017_125209_regis_meilisearch.md
DawoudSheraz: Created https://github.com/overhangio/tutor-discovery/issues/88 to add ES to tutor-discovery and unblock this PR.

regisb: Yes, this change will break discovery. But since we agreed that we tutor-discovery wouldn't be supported in Sumac, I think we're fine.

tutor/config.py
```diff
@@ -235,7 +235,6 @@ def upgrade_obsolete(config: Config) -> None:
     for name in [
         "ACTIVATE_LMS",
         "ACTIVATE_CMS",
-        "ACTIVATE_ELASTICSEARCH",
```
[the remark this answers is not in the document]
regisb: The role of this function is to get rid of obsolete settings. In the past `ACTIVATE_ELASTICSEARCH` was renamed to `RUN_ELASTICSEARCH`. We don't have a `ACTIVATE_MEILISEARCH` setting, so we don't need to replace it.

tutor/templates/apps/openedx/settings/partials/common_all.py
regisb: Are you sure? The development.py settings are supposed to override the settings from common_all.py:

```
$ tutor dev run --no-deps lms ./manage.py lms shell -c "from django.conf import settings; print(settings.MEILISEARCH_PUBLIC_URL)"
http://meilisearch.local.openedx.io:7700
```

tutor/templates/dev/docker-compose.yml
regisb: Yes, great point.

tutor/templates/jobs/init/cms.sh
regisb: No, I don't think that's right. We assume that meilisearch is running, somewhere. Even if RUN_MEILISEARCH is false, it just means that meilisearch is running on a remote host. So we still need to run the migration commands.

tutor/templates/k8s/volumes.yml
```diff
@@ -78,4 +78,4 @@ spec:
     requests:
       storage: 1Gi
```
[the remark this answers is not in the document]
regisb: > The original meilisearch plugin [requests 5Gi storage](https://github.com/open-craft/tutor-contrib-meilisearch/blob/main/tutor_meili/patches/k8s-volumes#L14) -- is it ok to use that here too?

Sure.

> We also need a caddy alias added, I think in tutor/templates/local/docker-compose.prod.yml?

I'm not sure I understand why. Such aliases exist for internal communication between containers using the public URL. But we have a private URL setting for meilisearch, so we shouldn't need that alias. Perhaps the author of [this commit](https://github.com/open-craft/tutor-contrib-meilisearch/commit/a043ed04357574a6de49cf6e3c8d2f74ce578259) could enlighten us. @bradenmacdonald I summon thee \o/

tutor/templates/apps/openedx/settings/partials/common_all.py
pomegranited: Ah.. you are correct, the lms/development.py settings are fine, because they [import common_lms.py *before* setting MEILISEARCH_PUBLIC_URL](https://github.com/overhangio/tutor/pull/1141/files#diff-74ef659bea171686911a1025aa77a9a83fe9f602a3f3fcf201a19c37d8fb2380R5).

But the cms/development.py [imports common_cms.py *after* setting `MEILISEARCH_PUBLIC_URL`](https://github.com/overhangio/tutor/pull/1141/files#diff-d0396341985a7618fe5e749e2e8f7da3fadea6bb9aec6aefb3e456ac1ff99f7eR19), and so `MEILISEARCH_PUBLIC_URL` gets overwritten.

Can we move the common_cms.py import above?

tutor/templates/build/openedx/Dockerfile
```diff
@@ -265,6 +265,8 @@ RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
 # https://pypi.org/project/ipython (>=Python 3.10 started with 8.20)
 RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
     pip install ipdb==0.13.13 ipython==8.24.0
+# Re-install forked edx-search which is otherwise overridden by development.txt
+RUN pip install git+https://github.com/regisb/edx-search@regisb/meilisearch#edx-search
```
pomegranited, line 269: nit: this can be removed now that [edx-search==4.1.0](https://github.com/openedx/edx-search/tree/v4.1.0) is now installed in [`master`](https://github.com/openedx/edx-platform/blob/master/requirements/edx/base.txt#L516-L517) and [`open-release/sumac.master`](https://github.com/openedx/edx-platform/blob/open-release/sumac.master/requirements/edx/base.txt#L516-L517).

tutor/templates/k8s/volumes.yml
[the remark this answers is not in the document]
pomegranited: I was able to deploy a k8s sandbox without adding a caddy alias for meilisearch (using https://github.com/overhangio/tutor/pull/1143), so I withdraw this request. @bradenmacdonald  please disregard :)

tutor/templates/jobs/init/cms.sh
pomegranited: Oh yes, you're correct!

There's an issue with `reindex_course` referencing elasticsearch directly (https://github.com/openedx/edx-platform/pull/35743), but that's not Tutor's problem :)

pomegranited (approved):
👍 once comments [1](https://github.com/overhangio/tutor/pull/1141/files/3b5ab266207fad0b2d77e3c4574fe0d03c999d00#r1821673075) & [2](https://github.com/overhangio/tutor/pull/1141/files/3b5ab266207fad0b2d77e3c4574fe0d03c999d00#r1821723496) are addressed.

I tested this by deploying and testing a k8s sandbox configured with https://github.com/openedx/edx-platform/pull/35689 . The sandbox demonstrates a functional meilisearch deployment, with published course content search indexed for students, and draft content for courses + libraries v2 indexed in Authoring MFE for Content Authors. Note that this deployment required fixes from https://github.com/overhangio/tutor-mfe/pull/231 + https://github.com/openedx/edx-platform/pull/35743.

However, I'm unable to create new posts using tutor-forum@v2 (https://github.com/overhangio/tutor-forum/pull/48), so I wasn't able to test this with the new edx-search Meilisearch engine. Happy to leave the sandbox running to test updated versions of tutor-forum@v2 and https://github.com/edly-io/edx-platform.git@edly/forumv.

- [x] I tested this using the sandbox configured by https://github.com/openedx/edx-platform/pull/35689 
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> -- upgrade documentation will be done as part of a separate PR.
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

tutor/templates/build/openedx/Dockerfile
regisb: :+1:

tutor/templates/apps/openedx/settings/partials/common_all.py
regisb: Very good catch. I have no idea how the import found its way so far down. I'll move it back up.

tutor/templates/jobs/init/cms.sh
regisb: I suggest we run ` ./manage.py lms shell -c "import search.meilisearch; search.meilisearch.create_indexes()"` in init/lms.sh. What do you think?

regisb:
Again, thanks for the quick and thorough review Jill. Meilisearch is not supposed to work with forum v2 just yet, this would be pending some work on my side. I'd like to complete this PR before I move on to the integration of meilisearch in forum v2.

tutor/templates/k8s/volumes.yml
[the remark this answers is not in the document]
bradenmacdonald: For Studio search, the frontend makes queries to Meilisearch directly, so it must have a public URL _and_ a private URL. But if you're saying that inter-container communication is always using the private URL, not the public URL, so we don't need the alias - that makes sense to me.

tutor/templates/jobs/init/cms.sh
bradenmacdonald: Also keep in mind that until we resolve https://github.com/openedx/modular-learning/issues/235 (which is in progress), we should avoid running `./manage.py cms reindex_studio --experimental` on any instances with thousands of courses or more.

tutor/templates/k8s/volumes.yml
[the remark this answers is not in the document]
regisb: > But if you're saying that inter-container communication is always using the private URL, not the public URL

That's what I'm hoping. I know that edx-platform is making some weird internal calls to discovery, ecommerce and credential services with the public URL, but that's really a bad pattern (IMHO).

tutor/templates/jobs/init/cms.sh
regisb: OK I'll comment out this command then.

regisb: @bradenmacdonald Now that I think of it, it does seem overkill to re-index the studio but also courses in general for every call to `tutor local launch`. @DawoudSheraz also noticed that we re-indexed all courses during course import: https://github.com/overhangio/tutor/blob/e2786afa58106a9a006ee5aa224cc806ed1bb917/tutor/commands/jobs.py#L176

Is there any way we can be smarter about which courses to re-index, and when? For instance, it would be reasonable to automatically re-index a course after it has been created. 

In a perfect world, I would like to be able to get rid of all calls to `./manage.py cms reindex_studio` and `./manage.py cms reindex_course` in Tutor.  Do you think that's possible within a reasonable time frame?

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -34,12 +34,14 @@
 # Behave like memcache when it comes to connection errors
 DJANGO_REDIS_IGNORE_EXCEPTIONS = True
 
-# Elasticsearch connection parameters
-ELASTIC_SEARCH_CONFIG = [{
-  {% if ELASTICSEARCH_SCHEME == "https" %}"use_ssl": True,{% endif %}
-  "host": "{{ ELASTICSEARCH_HOST }}",
-  "port": {{ ELASTICSEARCH_PORT }},
-}]
+# Meilisearch connection parameters
```
DawoudSheraz, line 37: nit: I don't think this comment is needed here, the code ahead is self-explanatory.

tutor/templates/jobs/init/cms.sh
bradenmacdonald: @regisb We only need to run `reindex_studio` on the initial launch of a new instance (because it creates the index and sets the required index parameters, much like a database migration), if the index configuration has changed (only affects tutor nightly / master branch of edx-platform), and presumably when the user upgrades to a new major version.

Running it on the initial launch will be extremely fast, assuming no courses/libraries exist.

So I would recommend putting it only in the upgrade / provisioning workflow and not in the general `launch` flow. Except maybe in development mode it should be in the `launch` as well, assuming people won't have too much data on their dev instances.

bradenmacdonald: As for re-indexing courses, I haven't looked into it but it does sound like we're doing it more than is necessary.

regisb: Is there a function that would only create and configure the indexes (without deleting them first), such that we could call this function at every launch? Pretty much like a `./manage.py migrate` command.

bradenmacdonald: It doesn't exist yet, but we could add that. A `--create-only` flag would be trivial.

The edge case is when upgrading to a newer commit/release and the index parameters have changed; in that case, we need to replace the index and the new index will be empty unless we also re-index all the content. This could be done asynchronously though; it also doesn't replace the existing index until the reindexing is complete, so there's no downtime.

regisb: I think we do need a lightweight, idempotent index creation command that works in just all cases. This can take different forms: a single python function, a management command, or a set of migrations. If we can get this in Sumac then that's great, if not then we should probably think about it for later releases.

tutor/templates/apps/openedx/settings/partials/common_all.py
regisb: Sure, it's obvious, but I'd like to keep this comment for proper separation of the different sections.

DawoudSheraz (approved):

tutor/templates/jobs/init/cms.sh
bradenmacdonald: I added a proposal to https://github.com/openedx/modular-learning/issues/235 for a new `--init` flag that will achieve this.

PR #1143, opened 2024-10-22
Fixes some missing config for meilisearch in Tutor core

pomegranited:


regisb:
Great catch. This should not be necessary after the latest round of changes in my PR: https://github.com/overhangio/tutor/pull/1141

PR #1148, opened 2024-10-28
fix: wrong YouTube URL

CodeWithEmad:
This will fix the old YouTube URL returning 404.


DawoudSheraz (approved):
The commit type should be docs as the change is document-related only.

PR #1152, opened 2024-11-05
fix: attach container stdin on tutor dev start <service>

Danyal-Faheem:
Edx-platform developers regularly utilize debuggers inside their codebase and the `tutor dev start` command does not allow attaching to containers on certain OSes such as MacOS. This command will allow users to attach a single running container using the `docker compose attach` command.

Caveats:

- Can only attach one service in one running terminal

### DIFF
--- a/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
+++ b/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
@@ -1,1 +1,1 @@
-- [Improvement] Attach container stdin when running tutor dev start <service> for debugging. (by @Danyal-Faheem)
+- [Bugfix] Fix breakpoint debugging by attaching container stdin when running tutor dev start <service> for a single <service>. (by @Danyal-Faheem)
--- a/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
+++ b/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
@@ -1,1 +1,1 @@
-- [Bugfix] Fix breakpoint debugging by attaching container stdin when running tutor dev start <service> for a single <service>. (by @Danyal-Faheem)
+- [Bugfix] Fix breakpoint debugging by attaching container stdin when running tutor dev start <service> for a single service. (by @Danyal-Faheem)
--- a/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
+++ b/changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
@@ -1,0 +1,1 @@
+- [Improvement] Attach container stdin when running tutor dev start <service> for debugging. (by @Danyal-Faheem)
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -245,23 +245,25 @@ def start(
     services: list[str],
 ) -> None:
     command = ["up", "--remove-orphans"]
-    attach = len(services) == 1
+    attach = len(services) == 1 and not detach
     if build:
         command.append("--build")
     # We have to run the container in detached mode first to attach to it
     if detach or attach:
         command.append("-d")
-
+    else:
+        fmt.echo_info("ℹ️  To exit logs without stopping the containers, use ctrl+z")
+    
     # Start services
     config = tutor_config.load(context.root)
     context.job_runner(config).docker_compose(*command, *services)
-    if attach and not detach:
+    
+    if attach:
         fmt.echo_info(
             f"""Attaching to service {services[0]}
 ℹ️  To detach without stopping the service, use ctrl+p followed by ctrl+q"""
         )
         context.job_runner(config).docker_compose("attach", *services)
-
 
 @click.command(help="Stop a running platform")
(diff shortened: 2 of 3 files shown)


kdmccormick:
Thanks for the PR Danyal! This is a problem I have noticed on Ubuntu as well. I'm able to attach to an individual container's stdout by doing, for example, `tutor dev start cms`, [which is what the docs recommend](https://docs.tutor.edly.io/tutorials/edx-platform.html#debugging-with-breakpoints). But I'm not able to run any debugger commands, so it seems that stdin is detached.

I'm not super keen to add a new command, but breakpoint debugging is definitely something we all need. I'll test this out and leave a review soon.

regisb:
I was hoping I was the only one who observed that issue... The problem is that, in my case, `docker compose attach` does not work either. I need to run `docker attach <containerID>`. Is it the case for you as well?

Danyal-Faheem:
@regisb, I get the same error as you do which prompted me to make this PR.

I also initially thought that `docker compose attach <container>` wouldn't work and I would need to do `docker attach <container>` instead, but to my surprise, it works perfectly.

kdmccormick (changes requested):
This is working great for me. Can you update those dev docs that I linked in my comment above?

Danyal-Faheem:
Hey @kdmccormick, I've updated the docs as well. Can you take another look and let me know what you think?

regisb:
Based on my experience, it seems that `docker compose up` no longer attaches stdin to the specified containers. I'm wondering why the right fix should be. Is it to add a new command? Or should we automatically run `docker compose attach` when we run `tutor dev start <a single container>`?

kdmccormick:
@regisb , I really like that idea, especially since I cannot imagine any situation where I run `tutor dev start lms` _without_  wanting stdin to be attached.

There is only one minor problem:
* To escape from `docker compose up` without stopping containers, the sequence is `Ctrl-z`
* To escape from `docker compose attach` without stopping containers, the sequence is `Ctrl-p Ctrl-q`

If we went with your proposal, that would mean that `tutor ... start` would have different escape sequences depending on whether you provided zero/multiple argument versus 1 argument. Perhaps we could mitigate the confusion with good docs and by printing a helpful message, like: `ℹ️ To detach without stopping the platform, type: ....`.

What do you think @regisb , go forward with modifying `tutor ... start`, or stick with `tutor dev attach`?

regisb:
> Perhaps we could mitigate the confusion with good docs and by printing a helpful message, like: ℹ️ To detach without stopping the platform, type: .....

Yes, I think we should do that. I honestly didn't know about escape commands, as I just ctrl+c all the time. In the process I just discovered that we also need to ctrl+p+q on `tutor dev run`.

So my vote goes to changing the "start" command. Especially since it's so easy to run `tutor dev dc attach`, I don't really see a point of adding another command.

kdmccormick:
@Danyal-Faheem could you incorporate [Régis's suggestion](https://github.com/overhangio/tutor/pull/1152#issuecomment-2473406307) as well as a help message for detachment?

Danyal-Faheem:
Hey @kdmccormick, sorry about that. I've been pretty busy the past few weeks. I'll try to make the requested changes this week.

changelog.d/20241205_211837_danyal.faheem_add_dev_attach_command.md
```diff
@@ -0,0 +1 @@
+- [Improvement] Attach container stdin when running tutor dev start <service> for debugging. (by @Danyal-Faheem)
```
kdmccormick, line 1:
```suggestion
- [Bugfix] Fix breakpoint debugging by attaching container stdin when running tutor dev start <service> for a single <service>. (by @Danyal-Faheem)
```

I'd consider this a bugfix, since the docs have incorrectly stated that this worked for a while.

tutor/commands/compose.py
```diff
@@ -245,14 +245,22 @@ def start(
...
+    # We have to run the container in detached mode first to attach to it
+    if detach or attach:
         command.append("-d")
 
     # Start services
     config = tutor_config.load(context.root)
     context.job_runner(config).docker_compose(*command, *services)
+    if attach and not detach:
+        fmt.echo_info(
+            f"""Attaching to service {services[0]}
```
kdmccormick, line 260: In the case that we are running multiple services without `-d`, can you print that `ctrl-z` will exit logs without stopping the services?

kdmccormick (changes requested):
A couple small requests, otherwise this looks great and works great.

tutor/commands/compose.py
```diff
@@ -245,14 +245,22 @@ def start(
     services: list[str],
 ) -> None:
     command = ["up", "--remove-orphans"]
+    attach = len(services) == 1
```
regisb, line 248: Nit: we can simplify the code slightly by writing:

 `attach = len(services) == 1 and not detach`

Then below:

```if detach: 
    command.append("-d")
...
if attach:
    ...
```

Danyal-Faheem: That's a great suggestion, I've made the change.

kdmccormick (approved):
Looks good and works great. This is an important bugfix, thank you @Danyal-Faheem !

PR #1153, opened 2024-11-07
feat!: Rename branches master->release, nightly->main

kdmccormick:
TEP: https://discuss.openedx.org/t/tep-rename-nightly-next-and-master-release/12994

GitHub issue: https://github.com/overhangio/tutor/issues/1096

We also incorporate the `IS_NIGHTLY` suggestion from [this rejected bugfix PR](https://github.com/overhangio/tutor/pull/1044), under the name `TUTOR_BRANCH_IS_MAIN`, as a reliable way of determining whether to patch for Tutor Release or Tutor Main.

To Do before merging:

* Queue up PRs for the other core plugins
* Make sure Edly is ready to update the GitLib pipelines
* Make sure Regis is ready to rename the branch in GitHub
* Maybe: _Wait until after the Sumac release on Dec 9?_

Immediately after merging:

* Merge this same commit into Main (the auto-sync may be temporarily broken at this point)
* Rebase this PR and merge it into Main: https://github.com/overhangio/tutor/pull/1154
* Merge the core plugin PRs
* Merge: https://github.com/overhangio/tpi/pull/6
* Edly: Merge GitLab pipline updates
* Regis: Rename branches in GitHub
* Put an announcement in the forums, x-posted to Slack, explaining how Nightly devs can migrate to Main:
  ```bash
  tutor dev stop
  OLD_ROOT="$(tutor config printroot)"
  OLD_PLUGINS_ROOT="$(tutor plugins printroot)"
  git branch -m nightly main
  git switch main
  git pull
  NEW_ROOT=$(tutor config printroot)"
  NEW_PLUGINS_ROOT=$(tutor plugins printroot)"
  mv "$OLD_ROOT" "$NEW_ROOT"`
  mv "$OLD_PLUGINS_ROOT" "$NEW_PLUGINS_ROOT"`
  tutor config save
  ```

PR #1154, opened 2024-11-07
Change version suffix to 'main' and load plugins from their main branches

kdmccormick:
This PR contains the https://github.com/overhangio/tutor/pull/1153, plus two Nightly-specific (aka Main-specific) commits:

* **Change version suffix from nightly to main.** As indicated in the linked PR's changelog entry, this is a breaking change, as it will rename the default TUTOR_ROOT and TUTOR_PLUGINS_ROOT.
* **Load plugins from main branches rather than nightly branches**. This is a safe changes as long as the plugins have already renamed their branches. If they haven't, then pop this commit off and merge it in later.


regisb (approved):
gogogo!

PR #1155, opened 2024-11-11
feat: new patch for mysql options

kaustavb12:
## Description

Currently no charset is set if extrenal MySQL instance is used with Tutor. Also, there is no option to set `charset` as required.

This PR introduces a patch called `mysql-options` which will allow tutor user to set `charset` and other options as required.

## Other information

Private Ref : [FAL-3945](https://tasks.opencraft.com/browse/FAL-3945)


bradenmacdonald:
Shouldn't we just make `charset: "utf8mb4"` the default in general, rather than relying on users to patch it? Is it not the case because some people use an old version of MySQL that doesn't support it, or have old table formats, or what?

bradenmacdonald:
We'll need some input here from a Tutor maintainer.

tutor/templates/apps/openedx/config/partials/auth.yml
```diff
@@ -20,5 +20,6 @@ DATABASES:
       {%- if RUN_MYSQL %}
       charset: "utf8mb4"
       {%- endif %}
+      {{ patch("mysql-options") }}
```
regisb, line 23: Let's not add anything to this file, which I would very much like to see disappear. MySQL settings can already be overridden from the edx-platform common settings. Let's not add useless patches.

Issue #937, opened 2023-11-14
We must replace uwsgi by something else


regisb:
uwsgi is now in maintenance mode: https://uwsgi-docs.readthedocs.io/en/latest/

> The project is in maintenance mode (only bugfixes and updates for new languages apis). Do not expect quick answers on github issues and/or pull requests (sorry for that) A big thanks to all of the users and contributors since 2009. 

So we should take the opportunity that we are releasing Quince to remove uwsgi from the code base.

DawoudSheraz:
Before jumping into the context, the following 3 terms are very different despite having similar names:

- WSGI (Web Server Gateway Interface) - it is a specification that defines the interfacing between Web server and an application. It is used in Open edX backend apps, via wsgi.py file.
- uWSGI - a utility/package that allows building hosting services compliant with WSGI standards (this is uwsgi this issue is talking about). It offers many things like HTTP server, proxies, process managers, monitoring, etc.
- uwsgi - a binary protocol used by uWSGI to talk to other web servers (like nginx)

Without going deeper into various utils, the top results that could be alternatives to uWSGI include:

-  [gunicorn](https://gunicorn.org/)
-  [mod_wsgi](https://modwsgi.readthedocs.io/en/master/index.html)
- [CherryPy](https://docs.cherrypy.dev/en/latest/advanced.html#wsgi-support)
- [waitress](https://docs.pylonsproject.org/projects/waitress/en/stable/index.html#extended-documentation)

Upon looking in tutor, I found out gunicorn was used before uWSGI. The migration to uWSGI was done when upgrading to Koa (https://github.com/overhangio/tutor/commit/728ef966dc572d37cf76f08c112277a5111b0d6b). Not sure if we can still revert to gunicorn in current state. Compared with uWSGI, gunicorn is slow (some comparison is on https://emptyhammock.com/projects/info/pyweb/gunicorn.html, however it might not be exactly true).

mod_wsgi is mainly Apache module. There were some references of nginx mod_wsgi but I was not able to find any latest information on that. CherryPy is both a Web server and a minimalistic Web framework. Waitress, on the other hand, is WSGI only server running on PyPy on Unix (python 3.7+).

I will be looking into each item and further explore any other alternatives as well. 

### References

1. https://stackoverflow.com/questions/38601440/what-is-the-point-of-uwsgi
2. https://www.ultravioletsoftware.com/single-post/2017/03/23/An-introduction-into-the-WSGI-ecosystem
[...]

PR #1157, opened 2024-11-12
fix: Don't build uwsgi with xml support.

feanil:
When uwsgi runs with xml support, it throws the following error on startup with the latest version of edx-platform:

```
xmlsec.InternalError: (-1, 'lxml & xmlsec libxml2 library version mismatch')
```

See https://github.com/xmlsec/python-xmlsec/issues/320 for more details on this.

Essentially, the uWSGI wheel is built against a different version of libxml2 and it causes this version mismach error when trying to load xmlsec.  The xml support in uWSGI is only needed for runing with xml configuration files, which tutor does not do.  Following the guidance of the above issue, I updated the `openedx` Dockerfile to no longer compile with any xml support as a part of the build. The time to build uWSGI was only slightly more than building from cache so I'm not concerned about major slowdowns in the build time for un-cached builds.

### DIFF
--- a/changelog.d/20241113_131448_feanil_fix_uwsgi_xml_issues_master.md
+++ b/changelog.d/20241113_131448_feanil_fix_uwsgi_xml_issues_master.md
@@ -1,13 +1,1 @@
-<!--
-Create a changelog entry for every new user-facing change. Please respect the following instructions:
-- Indicate breaking changes by prepending an explosion 💥 character.
-- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
-- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
-  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
-  the release notes for every release.
--->
-
-<!-- - 💥[Feature] Foobarize the blorginator. This breaks plugins by renaming the `FOO_DO` filter to `BAR_DO`. (by @regisb) -->
-<!-- - [Improvement] This is a non-breaking change. Life is good. (by @billgates) -->
-
-[Bugfix] Don't build uwsgi with XML support (by @feanil)
+- [Bugfix] Don't build uwsgi with XML support (by @feanil)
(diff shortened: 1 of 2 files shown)


regisb (commented):
This seems to be one of the first ripple effects of the pending [uwsgi deprecation](https://github.com/overhangio/tutor/issues/937)... :fearful: 

Since this is fixing a bug, can you please add a changelog entry? https://docs.tutor.edly.io/tutor.html#contributing

DawoudSheraz:
Since ubuntu PR was merged in nightly, this PR should be merged to nightly as well. Apart from changelog, you might want to rebase the branch on nightly.

feanil:
Added a changelog entry, let me know if you need anything else.

changelog.d/20241113_131448_feanil_fix_uwsgi_xml_issues_master.md
```diff
@@ -0,0 +1,13 @@
...
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+
+<!-- - 💥[Feature] Foobarize the blorginator. This breaks plugins by renaming the `FOO_DO` filter to `BAR_DO`. (by @regisb) -->
+<!-- - [Improvement] This is a non-breaking change. Life is good. (by @billgates) -->
+
```
DawoudSheraz, line 12: please remove this default text from changelog.

changelog.d/20241113_131448_feanil_fix_uwsgi_xml_issues_master.md
```diff
@@ -0,0 +1,13 @@
...
+[Bugfix] Don't build uwsgi with XML support (by @feanil)
```
DawoudSheraz, line 13:
```suggestion
- [Bugfix] Don't build uwsgi with XML support (by @feanil)
```

DawoudSheraz (approved):

DawoudSheraz:
@regisb Can you take another look at this, please? Thanks

regisb (approved):
lgtm! we should now split the django-redis and uwsgi installations, because they use different installation instructions, but we can do this in a separate PR.

PR #1158, opened 2024-11-14
fix: disable MySQL backend in forum on upgrade

regisb:
In Sumac, it was decided that we would not automatically migrating data from mongodb to mysql. Instead, we encourage users to do so manually, in the tutor-forum README.


tutor/commands/upgrade/common.py
```diff
@@ -46,6 +46,22 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
...
+    if plugins.is_loaded("forum"):
+        fmt.echo_alert(
+            "Your platform is going to be configured to store forum data in MongoDB. "
+            "You are STRONGLY ENCOURAGED to migrate your forum data to MySQL as soon as possible. "
+            "To do so, refer to the tutor-forum plugin documentation: https://github.com/overhangio/tutor-forum/#installation"
+        )
+        context.obj.job_runner(config).run_task(
+            "lms",
+            """
+(./manage.py lms waffle_flag --list | grep forum_v2.enable_mysql_backend) || ./manage.py lms waffle_flag --create --deactivate forum_v2.enable_mysql_backend
```
DawoudSheraz, line 60: curious: OR here is to not run flag creation/disable if the flag is not set in the first place, right?

regisb: Yes, exactly. Maybe the platform administrator has decided to migrate (`forum_v2.enable_mysql_backend = true`) or to explicitly stick to mongodb (`forum_v2.enable_mysql_backend = false`).

PR #1159, opened 2024-11-15
chore: split uwsgi/django-redis installation for efficiency

regisb:
I decided not to add a changelog entry because there is no impact on end users.


regisb:
FYI @feanil

PR #1161, opened 2024-11-19
fix: When pushing images to registry cache, use image-manifest=true

fghaas:
Without this change, when building images with `--cache-to-registry`, BuildKit uses a proprietary cache artifact format, which breaks when using third-party registries such as [Harbor](https://goharbor.io/) or [ECR](https://aws.amazon.com/ecr/).
    
By adding the `image-manifest=true` option, BuildKit uses an OCI-compliant cache artifact format that should be compatible with all registries. This option requires BuildKit 0.12 or later (check with `docker buildx ls`).
    
See https://github.com/goharbor/harbor/issues/18941 and https://github.com/moby/buildkit/issues/2251 for background information.
    
Co-authored-by: @angonz

### DIFF
--- a/changelog.d/20241119_111602_fghaas_image_manifest.md
+++ b/changelog.d/20241119_111602_fghaas_image_manifest.md
@@ -1,0 +1,8 @@
+- [Improvement] When building images with
+  `tutor images build --cache-to-registry`, use an OCI-compliant cache
+  artifact format that should be universally compatible with all
+  registries. This enables the use of that option when working with
+  third-party registries such as [Harbor](https://goharbor.io/) or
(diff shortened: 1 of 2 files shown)


fghaas:
This change should not pose problems for users who have previously used the default image artifact format, since the cache is only *populated* on a fresh image build. That is to say, if this change is applied to an existing Tutor installation, and the user rebuilds an image (bumping the image tag, of course), then only the new image will use the compatible cache artifact format.

regisb:
Hi Florian! Can you please have a look at [this comment](https://github.com/overhangio/tutor/pull/1119#issuecomment-2413964465) and tell me what you think?

regisb (approved):
looks awesome to me!

PR #1162, opened 2024-11-21
feat: get rid of `is_docker_rootless` template filter

regisb:
This filter was used only by Elasticsearch, so we no longer need to include it in Tutor.


jfavellar90:
The last commit on this [PR](https://github.com/overhangio/tutor/pull/1162/commits/53b778b69e68e5a20ca9dbfa7f1208ccd782b521) LGTM, however, there are still conflicts to merge

DawoudSheraz (approved):

Issue #956, opened 2023-12-06
`make test-types` fails for plugins when tutor is installed in editable mode


regisb:
**Bug description**

Type tests fail when tutor is installed in editable mode. This is inconvenient when testing plugins locally.

**How to reproduce**

Install tutor in editable mode:

    cd tutor/
    pip install -e .[dev]

Run type tests for a plugin:

    cd tutorcairn/
    make test-types

```
mypy --exclude=templates --ignore-missing-imports --implicit-reexport --strict ./tutorcairn
tutorcairn/plugin.py:122: error: Untyped decorator makes function "_print_superset_host" untyped  [misc]
Found 1 error in 1 file (checked 3 source files)
make: *** [Makefile:16: test-types] Error 1
```

That error is caused by the fact that mypy does not find the tutor module, has the following attests:

```
$ mypy -m tutor
mypy: can't find module 'tutor'
```

**Additional context**

This is a known issue. See the upstream tickets in mypy and setuptools:

https://github.com/python/mypy/issues/13392
https://github.com/pypa/setuptools/issues/3518

The issue can be bypassed by installing tutor in [legacy mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior):

    pip install -e . --config-settings editable_mode=compat

Or by installing in non-editable mode:

    pip install .

Abdul-Muqadim-Arbisoft:
The issue encountered involved a type checking error with` mypy`, a static type checker for Python, due to missing` type stubs` for `pkg_resources`, a module in the `setuptools library`. `Setuptools` is used for building and distributing Python packages but doesn't include its own type annotations. This problem was resolved by installing the `types-setuptools package`, which provides the necessary type stubs for `setuptools`. With `types-setuptools` installed,` mypy` could successfully analyze the code for type consistency, eliminating the above encountered error.

regisb:
Here's a Dockerfile to reproduce the issue:

```
# syntax=docker/dockerfile:1
FROM python:3.10

RUN mkdir -p /test

ADD --keep-git-dir=true https://github.com/overhangio/tutor.git#master /test/tutor
ADD --keep-git-dir=true https://github.com/overhangio/tutor-cairn.git#master /test/tutor-cairn

RUN python -m venv /test/venv
ENV PATH /test/venv/bin:${PATH}
RUN pip install --upgrade pip setuptools
RUN pip install -e '/test/tutor[dev]'

WORKDIR /test/tutor-cairn
[...]

PR #1163, opened 2024-11-22
feat: migrate from setup.py/setuptools to pyproject.toml/hatch

regisb:
While investigating issue #956, I realized that the issue might be resolved if we switched from setuptools to hatch as package builder. But to achieve that, we also need to migrate from setup.py to pyproject.toml. This was a long overdue chore, as setup.py is no longer recommended (see [PEP 621](https://peps.python.org/pep-0621/)).

I'm not sure whether we should be merging this change now, later (or ever). I _think_ this should be safe... So I would lean towards merging it now (and eventually migrate plugins and cookiecutter to pyproject.toml as well). But I'm curious what ya'll think @kdmccormick @DawoudSheraz.

Updated twine and other dependencies to fix some build issues https://github.com/pypi/warehouse/issues/15611


Makefile
```diff
@@ -82,12 +78,8 @@ coverage-browse-report: coverage-html ## Open the HTML report in the browser
 bundle: ## Bundle the tutor package in a single "dist/tutor" executable
 	pyinstaller tutor.spec
 
-bootstrap-dev: ## Install dev requirements
-	pip install .
-	pip install -r requirements/dev.txt
-
-bootstrap-dev-plugins: bootstrap-dev ## Install dev requirements and all supported plugins
```
DawoudSheraz, line 89: any particular reasons for removing this?

Makefile
```diff
@@ -9,19 +9,15 @@ docs: ## Build HTML documentation
 	$(MAKE) -C docs
 
 compile-requirements: ## Compile requirements files
-	pip-compile requirements/base.in
-	pip-compile requirements/dev.in
-	pip-compile requirements/docs.in
+	pip-compile ${COMPILE_OPTS} requirements/base.in
+	pip-compile ${COMPILE_OPTS} requirements/dev.in
+	pip-compile ${COMPILE_OPTS} requirements/docs.in
```
DawoudSheraz, line 14: What's the reason for having COMPILE_OPTS? What values can it have?

regisb: This is a technique that I also used in edx-platform to de-duplicate the list of `*.in` files in the Makefile.

Makefile
regisb: The `make` are not supposed to be part of the Tutor API, so I think it's OK to make breaking changes here. The goal of this change is to get rid of an unused command.

pyproject.toml
```diff
@@ -1,2 +1,60 @@
+# https://packaging.python.org/en/latest/tutorials/packaging-projects/
+# https://hatch.pypa.io/latest/config/build/
+
+[project]
+name = "tutor"
+license = {file = "LICENSE.txt"}
```
Danyal-Faheem, line 6: As we are now using the entire license file as the license, running `pip show tutor` displays the entire license file against the license field making it a cluttered mess if someone wants to read the other attributes of the package.

Instead I propose changing this to a text:

```
license = {text = "AGPL-3.0-only"}
```
The license names are available [here](https://spdx.org/licenses/).

CC: @DawoudSheraz

DawoudSheraz: Thanks for highlighting this. Using text should be good enough, the entire license content will certainly make pip show difficult to navigate. Also, I don't think we need to have dict with text key, we can add the license name directly
```
license = "AGPL-3.0-only"
```

Danyal-Faheem: > Also, I don't think we need to have dict with text key, we can add the license name directly
Unfortunately, it doesn't work for me without the dict and text key, it shows empty otherwise 😕.

DawoudSheraz: Right, I am also able to reproduce it.

DawoudSheraz: There are inconsistencies in metadata translations w.r.t build tools. Even though hatch's official docs mention using license="str" to populate license, many users have reported the metadata not generating/rendering as expected https://stackoverflow.com/questions/75570780/pip-show-does-not-show-summary-home-page-author-and-license.

Danyal-Faheem (approved):

DawoudSheraz:
@regisb Some updates were made to this pull request. Please take a look. Thanks

regisb (commented):
Looks great to me!

Issue #1126, opened 2024-09-27
Celery lms/cms-worker consumes too much RAM


igobranco:
**Bug description**

The normal execution of the celery workers, lms-worker and cms-worker services, consumes too much memory by default.
Tested on my tutor local, but I'm assuming that with tutor dev is the same.
By default the celery start 1 process by each CPU. If the deployed laptop or server has multiple CPU architecture, then the celery will launch multiple OS processes.
Each celery OS process consumes memory.
This makes more CPUs you have, more memory RAM the celery worker will consume.
The problem that I see is that on a deployed server, the operator probably don't want this magic, and want to control how much celery OS process each container/pod will start and consequently how much memory each container/pods will consume.

**How to reproduce**

Run a tutor local environment and see how much RAM your celery workers is consuming. 
On 16 CPU server/laptop an idle environment each container lms/cms-worker consumes >2GB RAM per container.

If you add `--concurrency=1` to the lms/cms-worker command an idle tutor local env uses <300MB per container.

**Environment**

tutor, version 14.2.3
But this applies also to newer version.

**Solution (my opinion)**
1. Add a way to parameterize the concurrency on each worker, ex: `--concurrency={{ LMS_WORKER_CELERY_CONCURRENCY }}`
2. Each celery worker by default should always consume the minimum memory, so it should defaults to just `1`.
3. In my opinion there should be an easy way to customize both celery commands, lms-worker and cms-worker. Ex. adding the `--prefetch-multiplier=1 --without-gossip --without-mingle`

regisb:
Hi Ivo, I agree that we need to provide a way to override the number of concurrent workers. And we also need to provide good defaults for that value.

In that context, I'm curious to have your take on [this comment](https://github.com/overhangio/tutor/pull/1010/files#r1516533054) by @dkaliberda, where he makes the case that we should default to concurrency=1 and instead scale the number of replicas.

I'm not sure I agree with that comment, because in my experience there is quite a bit of overhead incurred by scaling celery workers horizontally (by increasing the number of replicas) as opposed to vertically (by increasing the number of workers). For instance, here are the figures for memory usage on my laptop, in idle mode:

- 1 replica x 1 worker: 319 MB
- 1 replica x 4 workers: 746 MB
- 4 replicas x 1 worker: 1280 MB

In your current use case, what would be the ideal number of replicas/workers? (both for LMS and CMS)

igobranco:
[...]

PR #1165, opened 2024-11-22
fix: improve celery performance

Ian2012:
See https://github.com/eduNEXT/tutor/pull/3 for details

### DIFF
--- a/tutor/plugins/openedx.py
+++ b/tutor/plugins/openedx.py
@@ -153,6 +153,9 @@ hooks.Filters.LMS_WORKER_COMMAND.add_items(
         "--hostname=edx.lms.core.default.%h",
         "--queues=edx.lms.core.default,edx.lms.core.high,edx.lms.core.high_mem",
         "--max-tasks-per-child=100",
+        "--without-gossip", # Disable worker communication. All messages should be captured from the broker.
+        "--without-mingle", # Disable worker synchronization.
+        "--heartbeat-interval=60", # Reduce the heartbeat interval to 1 min.
     ]
 )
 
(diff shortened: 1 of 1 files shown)


regisb:
I agree that we should add `--without-mingle` and `--without-gossip`. But it seems that there is a misunderstanding about the heartbeat. If I understand [this article](https://ankurdhuriya.medium.com/understanding-celery-workers-concurrency-prefetching-and-heartbeats-85707f28c506) correctly (which has more information than the official docs...), the heartbeat is not to detect whether the broker is down, but for the broker to detect when a worker is down. When a worker is down, then its tasks are reassigned to another worker. Surely, we don't want to wait 1 minute until tasks are reassigned. I'm not sure whether we should increase from 2s. If we do, we shouldn't go beyond 5-10s (IMHO).

Concerning `-O fair`: @igobranco was suggesting [here](https://github.com/overhangio/tutor/issues/1126) that we add `--prefetch-multiplier=1`. This would resolve the issue without adding `-O fair`, right?

Ian2012:
If there is a concern about the heartbeat, we can use the default for now. 

Yes, setting the `prefetch-multiplier` to 1 should resolve that issue too, this setting is what improved the performance on a large instance and helped to not lose tasks when workers are OOM Killed or pods are deleted forcefully, this with `ACKS_LATE` enabled, however, this would require celery tasks to be idempotent, such as course indexing, grade calculation or reports, however I think that's a large discussion that's out of scope for this PR. See the [docs](https://docs.celeryq.dev/en/latest/userguide/optimizing.html#reserve-one-task-at-a-time) for more information. 

I will update the PR with the new defaults.

DawoudSheraz (approved):

regisb (changes requested):
Can you please add a changelog entry? https://docs.tutor.edly.io/tutor.html#contributing

regisb (approved):
Alright, let's merge this!

PR #1170, opened 2024-12-03
fix: use cms celery parameters for cms-worker

Danyal-Faheem:
This fixes a bug introduced in #1134 where opening a course would throw a 500 error because the Course Outline could not be found. This is because the cms-worker was being run with the lms-worker parameters and would never pick up the Course Outline Generation tasks.

### DIFF
--- a/tutor/templates/local/docker-compose.prod.yml
+++ b/tutor/templates/local/docker-compose.prod.yml
@@ -51,7 +51,7 @@ services:
     environment:
       SERVICE_VARIANT: cms
       DJANGO_SETTINGS_MODULE: cms.envs.tutor.production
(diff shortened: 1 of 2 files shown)


tutor/templates/local/docker-compose.prod.yml
```diff
@@ -51,7 +51,7 @@ services:
     environment:
       SERVICE_VARIANT: cms
       DJANGO_SETTINGS_MODULE: cms.envs.tutor.production
-    command: {% for value in iter_lms_celery_parameters() %}
+    command: {% for value in iter_cms_celery_parameters() %}
```
regisb, line 54: I can't believe we didn't see this one...

regisb (approved):
Great catch!

Issue #1172, opened 2024-12-03
Tutor is using a deprecated GitHub macOS 12 runner in releases


regisb:
**Bug description**

Publishing a new release fails on GitHub: https://github.com/overhangio/tutor/actions/runs/12144657784

> The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see https://github.com/actions/runner-images/issues/10721

We need to upgrade to macOS 13. Once we do, we need to upload the binary to the 18.2.0 release: https://github.com/overhangio/tutor/releases/tag/v18.2.0

**How to reproduce**

Attempt to publish a new release on GitHub.

PR #1175, opened 2024-12-09
build: update macos and python version in release yaml

DawoudSheraz:
Resolve https://github.com/overhangio/tutor/issues/1172.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
           - os: ubuntu-20.04
             locale: C.UTF-8
           # https://endoflife.date/macos
-          - os: macos-12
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         # https://github.com/actions/setup-python
         uses: actions/setup-python@v3
         with:
-          python-version: 3.9
+          python-version: 3.12
```
regisb, line 41: No, we should keep the oldest possible version here. See comment line 24.

DawoudSheraz: Reverted.

regisb (approved):
gogogo

PR #1176, opened 2024-12-09
feat: Force-enable navigation sidebar in upgrade to Sumac

arbrandes:
When upgrading to Sumac, forcefully enable the Learning MFE's navigation
sidebar.


tutor/commands/upgrade/common.py
```diff
@@ -47,6 +47,12 @@ def upgrade_from_nutmeg(context: click.Context, config: Config) -> None:
 
 
 def upgrade_from_redwood(context: click.Context, config: Config) -> None:
+    # Forcefully enable the learning MFE's navigation sidebar.
+    if plugins.is_loaded("mfe"):
+        context.obj.job_runner(config).run_task(
+            "lms", "./manage.py lms waffle_flag --create --everyone courseware.enable_navigation_sidebar"
```
regisb, line 53: :thinking: I believe that if the flag already exists, then this command will crash -- which is the reason why we do this little grep-toggle-sing-and-dance for other flags. Did you check? ideally this upgrade command would succeed even if the flag already exists.

arbrandes: I just tested it: it doesn't crash.  It does what we'd hope it does: if the flag doesn't exist, it creates it.  If it _does_ exist, it updates it.

arbrandes: (The reason for the grep dance is to persist the operator's past choices, I think.)

regisb: In that case :ok_hand: !

regisb (approved):

PR #1178, opened 2024-12-12
build: Explicitly Install assets.txt in openedx image

kdmccormick:
From the upstream Paver DEPR [1], this accounts for a change in the requirements files:

> Starting in Sumac, these dependencies will be removed from
> requirements/edx/base.txt. Instead, operators will need to install:
> * requirements/edx/assets.txt to build static assets

In the future, we could optimize the openedx image build by installing assets.txt in a separate, rarely-invalidated cache layer. libsass in particular takes 60+ seconds to install, so this is promising. However, for now, in order to prepare for [1], we make just the simplest possible change, which is to install assets.txt along with base.txt.

We also take this opportunity to bump the nodeenv version from 1.8.0 to 1.9.1, matching edx-platform's. This is another area where we could make use of assets.txt in a future Dockerfile refactoring.

[1] https://github.com/openedx/edx-platform/issues/34467

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -92,10 +92,11 @@ RUN --mount=type=cache,target=/openedx/.cache/pip,sharing=shared \
     # https://pypi.org/project/wheel/
     setuptools==69.1.1 pip==24.0 wheel==0.43.0
 
-# Install base requirements
+# Install base requirements and asset-building requirements
 RUN --mount=type=bind,from=edx-platform,source=/requirements/edx/base.txt,target=/openedx/edx-platform/requirements/edx/base.txt \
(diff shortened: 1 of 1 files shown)


kdmccormick:
@regisb I'm sure Sumac is your priority right now, but when you get a chance to look at it, this PR is the final barrier to us [removing paver from edx-platform entirely](https://github.com/openedx/edx-platform/pull/34832).

DawoudSheraz (approved):

regisb (approved):
Yup, this looks good to me. But shouldn't we wait until Sumac is released? Such that we can rebase sumac on top of main if necessary. The paver removal is not compatible with Sumac, right?

kdmccormick:
> But shouldn't we wait until Sumac is released? Such that we can rebase sumac on top of main if necessary. The paver removal is not compatible with Sumac, right?

Sure, sounds good. I'll merge right after Sumac.

PR #1179, opened 2024-12-15
feat: Add special characters to `random_string` util

### LINKED ISSUE #1177: Add util for random_string with special characters
https://github.com/overhangio/tutor/blob/0b934aff639fdd5c63ad859075c59ebea3b6a255/tutor/utils.py#L69

Add new util `random_string_special` that includes string.punctuation(https://docs.python.org/3/library/string.html#string.punctuation)

This is needed when using random_string to create Clickhouse passwords

vahid75:
If you pass argument: special=True, special characters will be added to the generated random string.
Close #1177

### DIFF
--- a/tutor/utils.py
+++ b/tutor/utils.py
@@ -66,9 +66,12 @@ def ensure_directory_exists(path: str) -> None:
         os.makedirs(path)
 
 
-def random_string(length: int) -> str:
+def random_string(length: int, special: bool=False) -> str:
+    characters = string.ascii_letters + string.digits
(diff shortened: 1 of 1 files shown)


regisb:
Hi! Thanks for your PR. I think that the solution proposed in issue #1177 (implement changes in a plugin) is sufficient and we don't need to make changes in Tutor core. If we did, we would have to figure out how to escape special characters in many different contexts, such as yaml, python, json, etc. And we don't have a great solution for that.

PR #1180, opened 2024-12-17
chore: add main version suffix, update plugins.txt to install main branches

DawoudSheraz:
Update version_suffix and plugins.txt on main branch post sumac release

PR #1181, opened 2024-12-18
chore: remove ecommerce plugin as it is deprecated

Faraz32123:
- remove ecommerce plugin from nightly as it is deprecated.
- check tutor-ecommerce deprecation issue for more info: overhangio/tutor-ecommerce#83


Faraz32123:
tutor-ecommerce was already removed from main [here](https://github.com/overhangio/tutor/pull/1180/commits/fbfef07339c0cf5b01902f953ef7ab91feab1075).

Made this PR to remove ecommerce completely from tutor main(nighlty).
@regisb @DawoudSheraz need your review on this.

tutor/plugins/v0.py
```diff
@@ -282,7 +282,6 @@ class OfficialPlugin(BasePlugin):
     NAMES = [
         "android",
         "discovery",
-        "ecommerce",
```
DawoudSheraz, line 285: not sure if this should be removed from v0 as it can serve some backwards compatibility purposes.

Faraz32123: yes, this needs to be discussed...

regisb: This class is only meant for the discovery of plugins in the official github-built binary (see `make bundle`). We don't want the ecommerce plugin there, at least not in the main branch, so we should remove it from this list.

DawoudSheraz (approved):

regisb (approved):
lgtm!

PR #1184, opened 2025-01-01
fix: add meilisearch as a dependency on lms-job

Danyal-Faheem:
Running the Meilisearch init job when the platform is stopped would crash as the meilisearch container wouldn't be running and the endpoint could not be hit.

We add a dependency in the lms-job on meilisearch as the meilisearch init job is run inside the lms container.

### DIFF
--- a/tutor/templates/local/docker-compose.jobs.yml
+++ b/tutor/templates/local/docker-compose.jobs.yml
@@ -29,7 +29,7 @@ services:
         {%- for mount in iter_mounts(MOUNTS, "openedx", "lms-job") %}
         - {{ mount }}
         {%- endfor %}
-      depends_on: {{ [("mysql", RUN_MYSQL), ("mongodb", RUN_MONGODB)]|list_if }}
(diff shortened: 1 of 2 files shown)


DawoudSheraz (approved):

tutor/templates/local/docker-compose.jobs.yml
```diff
@@ -29,7 +29,7 @@ services:
         {%- for mount in iter_mounts(MOUNTS, "openedx", "lms-job") %}
         - {{ mount }}
         {%- endfor %}
-      depends_on: {{ [("mysql", RUN_MYSQL), ("mongodb", RUN_MONGODB)]|list_if }}
+      depends_on: {{ [("mysql", RUN_MYSQL), ("mongodb", RUN_MONGODB), ("meilisearch", RUN_MEILISEARCH)]|list_if }}
```
regisb, line 32: I find it weird that we never caught this issue when we were running elasticsearch. What's the error stacktrace? And can you please add a changelog entry?

DawoudSheraz: ES did not have an explicit init job if I remember correctly.

regisb (approved):

Issue #1185, opened 2025-01-02
Sumac deployment fails with `RUN_MEILISEARCH: false`


mrtmm:
**Bug description**
Tutor deployment, running the Sumac release, fails with `ConnectionError: Unable to connect to Meilisearch` when `RUN_MEILISEARCH: false` is set.

I believe there are three steps in the platform launch where the `RUN_MEILISEARCH` setting is not honoured:
1. https://github.com/overhangio/tutor/blob/release/tutor/commands/jobs.py#L49
2. https://github.com/overhangio/tutor/blob/release/tutor/templates/jobs/init/lms.sh#L14
3. https://github.com/overhangio/tutor/blob/release/tutor/templates/jobs/init/cms.sh#L24

**How to reproduce**

1. Set `RUN_MEILISEARCH: false` in `config.yml`
2. Run `tutor config save`
3. Launch the platform as usual

**Environment**
Seeing the same failure with `tutor local` and `tutor k8s`

regisb:
I think you may be mistaken in the meaning of the `RUN_MEILISEARCH` setting. Meilisearch is a required component of the Open edX stack with Tutor in Sumac. Setting RUN_MEILISEARCH=false merely configures Tutor to point to an externally-hosted instance of Meilisearch. Thus, if you set RUN_MEILISEARCH=false, then you also need to set MEILISEARCH_URL and [other settings](https://docs.tutor.edly.io/configuration.html#meilisearch).

fghaas:
@regisb What do you think about flagging a configuration error, and refusing to start, if `RUN_MEILISEARCH` is set to `false` while `MEILISEARCH_URL` is simultaneously set to its default?

mrtmm:
> I think you may be mistaken in the meaning of the RUN_MEILISEARCH setting. Meilisearch is a required component of the Open edX stack with Tutor in Sumac. Setting RUN_MEILISEARCH=false merely configures Tutor to point to an externally-hosted instance of Meilisearch. Thus, if you set RUN_MEILISEARCH=false, then you also need to set MEILISEARCH_URL and [other settings](https://docs.tutor.edly.io/configuration.html#meilisearch).

I see, in that case I misunderstood the setting indeed. From the release notes I read:
> Two recent features, Studio Course Search and the new Content Libraries (beta), require Meilisearch and will be hidden from the UI if Meilisearch is not available. (For the Content Libraries beta, this affects only the UI - developers can still use the APIs without Meilisearch.)

[...]

PR #1186, opened 2025-01-02
feat: Add a plausibility check for Meilisearch configuration

fghaas:
Tutor is meant to be configured with either its own, or an externally preconfigured Meilisearch instance. Thus, the combination of configuring `RUN_MEILISEARCH: false` and leaving `MEILISEARCH_URL` at its default is almost certainly a configuration error.

Add an alert to notify the user of that configuration issue.

Related issue:
https://github.com/overhangio/tutor/issues/1185

### DIFF
--- a/tutor/config.py
+++ b/tutor/config.py
@@ -358,3 +358,23 @@ def _check_preview_lms_host(config: Config) -> None:
             f'Warning: PREVIEW_LMS_HOST="{preview_lms_host}" is not a subdomain of LMS_HOST="{lms_host}". '
             "This configuration is not typically recommended and may lead to unexpected behavior."
         )
+
+
+@hooks.Actions.CONFIG_LOADED.add()
+def _check_run_meilisearch(config: Config) -> None:
+    """
+    In case RUN_MEILISEARCH is set to false, check if
+    MEILISEARCH_URL has been set to a non-default value.
+    If not, print a warning to notify the user.
+    """
+
+    run_meilisearch = get_typed(config, "RUN_MEILISEARCH", bool, True)
+    if not run_meilisearch:
+        meilisearch_url = get_typed(config, "MEILISEARCH_URL", str, "")
+        meilisearch_url_default = get_defaults()["MEILISEARCH_URL"]
+        if meilisearch_url == meilisearch_url_default:
+            fmt.echo_alert(
+                "Warning: RUN_MEILISEARCH is false, but MEILISEARCH_URL is unset. "
+                "This configuration is not typically recommended and may lead to unexpected behavior. "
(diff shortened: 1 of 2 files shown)


tutor/config.py
```diff
@@ -358,3 +358,23 @@ def _check_preview_lms_host(config: Config) -> None:
...
+    """
+    In case RUN_MEILISEARCH is set to false, check if
+    MEILISEARCH_URL has been set to a non-default value.
+    If not, print a warning to notify the user.
+    """
+
+    run_meilisearch = get_typed(config, "RUN_MEILISEARCH", bool, True)
+    if not run_meilisearch:
+        meilisearch_url = get_typed(config, "MEILISEARCH_URL", str, "")
+        meilisearch_url_default = get_defaults()["MEILISEARCH_URL"]
```
regisb, line 374: I'd like to avoid making another call to `get_defaults` here, which is slow. Can we just hardcode "meilisearch"?

regisb (changes requested):
This looks good, with just a minor comment. Could you please rebase on top of the release branch, such that current sumac users can benefit from the change?

tutor/config.py
fghaas: I'm never a big fan of hardcoding the values of defaults, because it means that if the default ever changes, you need to fix it in more than one place. Is there another (better) way to avoid the `get_defaults()` call?

DawoudSheraz: > I'm never a big fan of hardcoding the values of defaults, because it means that if the default ever changes, you need to fix it in more than one place.

[thoughts] While I agree with this, I try to weigh the effort involved in determining how many "changes" are needed. If it is needed in very few places, it can be okay to hardcode.

regisb: I do agree with you Florian that hard-coding values is typically a bad practice. There are other ways to bypass this, but I wouldn't say they are "better". They include:

1) Your approach: calling `get_defaults()`, which is slow.
2) Implementing a `CONFIG_DEFAULTS_LOADED` action that would save a singleton copy of the (unrendered) config defaults, which would be quite complex.

If we ever forget to modify the meilisearch URL default, the warning you added will be present in the console stderr. Also, to make sure we don't forget to update this value, you could add a unit test.

DawoudSheraz: @fghaas Hi, following up on this PR. Thanks

DawoudSheraz: @fghaas Hi, checking up on this. Thanks

ahmed-arb:
Thanks for the original patch @fghaas. Since this has been waiting a while, I've continued it in #1492: rebased on `release`, with the `get_defaults()` call replaced by a hardcoded default plus a unit test that keeps the two in sync, as suggested by @regisb. Your commit is preserved with you as the author. Happy to fold in any further feedback you have there.

PR #1202, opened 2025-03-03
fix: `lms.startup` -> `django.setup`

kdmccormick:
lms/startup.py performs no logic other than setting up django, so it is being removed upstream:
https://github.com/openedx/edx-platform/pull/36302

### DIFF
--- a/tutor/templates/build/openedx/bin/site-configuration
+++ b/tutor/templates/build/openedx/bin/site-configuration
@@ -1,8 +1,8 @@
 #! /usr/bin/env python3
 import argparse
-import lms.startup
+import django
 
-lms.startup.run()
+django.setup()
 
 from django.conf import settings
 from django.contrib.sites.models import Site


regisb:
\o/ After we merge this, let's take this opportunity to remove the site-configuration script. I think that it will become mostly useless in Teak. @DawoudSheraz can you please create a corresponding issue?

ormsbee:
Thanks folks!

DawoudSheraz:
> \o/ After we merge this, let's take this opportunity to remove the site-configuration script. I think that it will become mostly useless in Teak. @DawoudSheraz can you please create a corresponding issue?

Isn't it used in plugins like MFE https://github.com/overhangio/tutor-mfe/blob/release/tutormfe/templates/mfe/tasks/lms/init#L13?

regisb:
I just discovered that this `ENABLE_PROFILE_MICROFRONTEND` site configuration will be removed in Teak. Please see this conversation on Slack for reference: https://openedx.slack.com/archives/CGE253B7V/p1740516391296149
As far as I know, site-configuration is also used in tutor-discovery, but I don't know if that's still necessary.

DawoudSheraz:
Created https://github.com/overhangio/tutor/issues/1207 for the removal of site-config.

PR #1209, opened 2025-03-11
Adds MEILISEARCH_HOST to expose it in ingress

farhaanbukhsh:
MEILISEARCH_HOST is introduced so that we can add it to the ingress controller

**Discussions**: [Slack Discussion](https://openedx.slack.com/archives/C05519HHZKM/p1739341680191669?thread_ts=1739257428.215229&cid=C05519HHZKM)

**Dependencies**: None

~~**Screenshots**: Always include screenshots if there is any change to the UI.~~

**Testing instructions**:

None

Private ref: [BB-9563](https://tasks.opencraft.com/browse/BB-9563)

REF: https://github.com/openedx/openedx-k8s-harmony/pull/150


DawoudSheraz (commented):
please add a changelog-entry to explain why this is needed in tutor core and cannot be done in ingress plugin

farhaanbukhsh:
> please add a changelog-entry to explain why this is needed in tutor core and cannot be done in ingress plugin

Hey @DawoudSheraz thank you for the review, I have added a changelog please have a look :)

changelog.d/20250319_182641_meilisearch_host.md
```diff
@@ -0,0 +1,14 @@
+<!--
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
```
DawoudSheraz, line 8: please remove this comment.

changelog.d/20250319_182641_meilisearch_host.md
```diff
@@ -0,0 +1,14 @@
...
+
+- [Improvement] This change introduces MEILISEARCH_HOST so that it can be
+re-used in the project, this also enables Tutor to configure meilisearch host
+and expose it to be used by other plugins. For example MEILISEARCH_HOST can be
+optinonally used to configure ingress controller in openedx-k8s-harmony like
+LMS_HOST etc.
```
DawoudSheraz, line 14: please add your GitHub username at the end. See other changelogs for reference

DawoudSheraz, line 14: > so that it can be
re-used in the project,

I don't think this wording is helpful. It is implicit that the configs are re-usable. The need of the config, which was explained later, should be good enough.

nit: optinonally --> optionally.

tutor/templates/config/defaults.yml
```diff
@@ -38,8 +38,9 @@ K8S_NAMESPACE: "openedx"
 LANGUAGE_CODE: "en"
 LMS_HOST: "www.myopenedx.com"
 LOCAL_PROJECT_NAME: "{{ TUTOR_APP }}_local"
+MEILISEARCH_HOST: "meilisearch.{{ LMS_HOST }}"
 MEILISEARCH_URL: "http://meilisearch:7700"
-MEILISEARCH_PUBLIC_URL: "{% if ENABLE_HTTPS %}https{% else %}http{% endif %}://meilisearch.{{ LMS_HOST }}"
+MEILISEARCH_PUBLIC_URL: "{% if ENABLE_HTTPS %}https{% else %}http{% endif %}://{{ MEILISEARCH_HOST }}"
```
DawoudSheraz, line 43: Why is MEILISEARCH_PUBLIC_URL not usable? Why do we need the host explicitly?

farhaanbukhsh: Great questions, and thank you for the review. If we already have a convention to use LMS_HOST, PREVIEW_HOST, etc., it only makes sense to have it as MEILISEARCH_HOST.

Adding to that MEILISEARCH_HOST helps us to configure the host URL, MEILISEARCH_PUBLIC_URL does the same but one has to be aware of the protocol which again was not a convention that was followed. Hence it made more sense to rename it and keep the MEILISEARCH_PUBLIC_URL unchanged.

Let me know  if it makes sense 😄

changelog.d/20250319_182641_meilisearch_host.md
farhaanbukhsh: thanks and you are right.

tutor/templates/config/defaults.yml
DawoudSheraz: Good enough. 
Why is this being merged to main and not release? It does not seem to be a breaking change.

farhaanbukhsh: @DawoudSheraz  sorry I wasn't aware of the contribution norms I have read those again and have made the changes.

DawoudSheraz (approved):

PR #1215, opened 2025-03-24
depr!: remove the site-configuration script

### LINKED ISSUE #1207: Remove site-configuration script in main
The ask originated from https://github.com/overhangio/tutor/pull/1202 where the custom startup present in edx-platform was removed in favor of django startup. site-configuration script is used to setup site objects in platform and this script provides a wrapper for that. It is used in tutor-mfe and tutor-discovery. To be able to remove the script from core, we need to ensure tutor-mfe and tutor-discovery are not using site-configuration script and maybe use another way to do so. See https://openedx.slack.com/archives/CGE253B7V/p1740516391296149 for some context, such as why tutor-mfe site-configuration usage might be removable all together.

- [ ] Ensure tutor-discovery main is updated to stop using site-config and has an alternative in place if needed
- [ ] Ensure tutor-mfe main is updated to stop using site-config and has an alternative in place if needed
- [ ] Once above items are done, remove site configuration from main branch

Danyal-Faheem:
Fixes #1207.

This PR is dependent on https://github.com/overhangio/tutor-mfe/pull/246 and https://github.com/overhangio/tutor-discovery/pull/101.

The site-configuration script was required because we needed to utilize the custom lms.startup() call to set the site-configuration. This is no longer required as of https://github.com/openedx/edx-platform/pull/36302. Therefore, we deprecate it in favour of the [create_or_update_site_configuration](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/site_configuration/management/commands/create_or_update_site_configuration.py) management command provided in edx-platform.


changelog.d/20250324_230721_danyal.faheem_remove_site_configuration_script.md
```diff
@@ -0,0 +1 @@
+- [Depreciation] Remove the unnecessary site-configuration script in favour of the create_or_update_site_configuration management command provided in edx-platform. (by @Danyal-Faheem)
```
DawoudSheraz, line 1: - We should mark this a breaking change, just in case
- We can add a note to use lms create_or_update_site_config mgt command as a replacement

Danyal-Faheem: I've already added the name of the management command in the changelog entry and mentioned its provided in edx-platform. Do you think we should add another line with a hyperlink to the command?

DawoudSheraz: Yes, the link can be helpful.

Danyal-Faheem: Added the hyperlink.

DawoudSheraz (approved):

PR #1229, opened 2025-05-13
feat: better stop on start

regisb:
When we run `local/dev start` we expect that any running `dev/local` project will be stopped. Previously, we were running `docker compose ls` twice, and for all running Compose projects on the server. We changed this behaviour:

1. By swapping the project name check and the check for running compose projects, we run `docker compose ls` just once: for dev in local mode, and vice versa.
2. By checking only Compose projects in the context of local/dev. That way we ignore any other non-tutor project.


mlabeeb03 (approved):
lgtm

DawoudSheraz (approved):

PR #1230, opened 2025-05-14
feat: only wait for tutor managed jobs while running k8s

### LINKED ISSUE #1206: K8S: Tutor should use selectors to validate already running jobs
**Bug description**
When Tutor is commanded to perform a K8S job (which uses the [K8sTaskRunner](https://github.com/overhangio/tutor/blob/v19.0.2/tutor/commands/k8s.py#L69) runner), it validates that other jobs [have been completed](https://github.com/overhangio/tutor/blob/v19.0.2/tutor/commands/k8s.py#L91). The problem with this is that it validates the state of [ALL](https://github.com/overhangio/tutor/blob/v19.0.2/tutor/commands/k8s.py#L188) the jobs in the namespace (not only those related to Tutor). If a non-related to Tutor job is running in the namespace, it will delay the execution of the Tutor K8S job till its completion. This is not an ideal situation in a namespace where additional jobs are running independently from Tutor.

A possible solution is to use job selectors so Tutor is able to validate that jobs related to Tutor have been completed.

**How to reproduce**

I used Minikube to test this behavior (used K8S 1.31 version, same kubectl version).

- Create a new namespace in the cluster.
- Install Tutor, create a new Tutor environment and run `tutor config save`.
- Change the kubectl context to point to the namespace you just created in the cluster.
- Create an infinite-running job in the namespace. This can be something like:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: infinite-job
spec:
  template:
    spec:
      restartPolicy: Never  # Can also be "OnFailure"
      containers:
        - name: infinite-loop
          image: busybox
          command: ["sh", "-c", "while true; do echo 'Running...'; sleep 60; done"]
  backoffLimit: 0
```
- Initialize tutor in K8S by running `tutor k8s launch -I`
- Check that the services initialization does not start due to the existence of the previously created job:

Image

- Delete the infinite-running job, once this is done, the K8S initialization can be executed with no issues

mlabeeb03:
closes #1206
This PR is an improvement that fixes the issue mentioned in #1206 
Now we are only filtering jobs that are managed by tutor. This allows running tutor jobs even if other non tutor jobs are running the same namespace.

### DIFF
--- a/changelog.d/20250514_161945_muhammad.labeeb_only_filter_tutor_jobs_in_k8s.md
+++ b/changelog.d/20250514_161945_muhammad.labeeb_only_filter_tutor_jobs_in_k8s.md
@@ -1,0 +1,1 @@
+-   [Improvement] Do not wait for non tutor jobs while running k8s. (by @mlabeeb03)
--- a/tutor/commands/k8s.py
+++ b/tutor/commands/k8s.py
@@ -7,9 +7,8 @@ import click
 
 from tutor import config as tutor_config
 from tutor import env as tutor_env
-from tutor import exceptions, fmt, hooks
+from tutor import exceptions, fmt, hooks, serialize, utils
 from tutor import interactive as interactive_config
-from tutor import serialize, utils
 from tutor.commands import jobs
 from tutor.commands.config import save as config_save_command
 from tutor.commands.context import BaseTaskContext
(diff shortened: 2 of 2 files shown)


regisb (approved):
Yup, all good :)

DawoudSheraz (approved):

PR #1239, opened 2025-05-22
fix: Fix the 'ln' command that fails silently

lewisemm:
- This command takes the form of: `ln -s source_directory target_directory`.

- If the `target_directory` already exists on the file system, the command will fail with a message that is similar to what is shown below.

	`ln: failed to create symbolic link 'target_directory': File exists`

- This message is not being logged (most likely due to how Docker's buildkit logs messages) and the build process fails when the '/scripts/copy-node-modules.sh' script is executed.

- This fix adds a 'rm' command to first remove the 'target_directory' before the 'ln' command is executed.

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -172,7 +172,8 @@ COPY --link --chown=$APP_USER_ID:$APP_USER_ID --from=nodejs-requirements /opened
 COPY --link --chown=$APP_USER_ID:$APP_USER_ID --from=nodejs-requirements /openedx/edx-platform/node_modules /openedx/node_modules
 
 # Symlink node_modules such that we can bind-mount the edx-platform repository
-RUN ln -s /openedx/node_modules /openedx/edx-platform/node_modules
+RUN rm -rf /openedx/edx-platform/node_modules && \
+    ln -s /openedx/node_modules /openedx/edx-platform/node_modules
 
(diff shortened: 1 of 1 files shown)


tutor/templates/build/openedx/Dockerfile
```diff
@@ -172,7 +172,8 @@ COPY --link --chown=$APP_USER_ID:$APP_USER_ID --from=nodejs-requirements /opened
 COPY --link --chown=$APP_USER_ID:$APP_USER_ID --from=nodejs-requirements /openedx/edx-platform/node_modules /openedx/node_modules
 
 # Symlink node_modules such that we can bind-mount the edx-platform repository
-RUN ln -s /openedx/node_modules /openedx/edx-platform/node_modules
+RUN rm -rf /openedx/edx-platform/node_modules && \
```
DawoudSheraz, line 175: Under what circumstances would /openedx/edx-platform/node_modules exist beforehand?

lewisemm: Sorry about this.

I have been trying to build the `openedx-dev` image locally using the following command.

```
tutor images build openedx-dev
```

The command above actually passes the `--build-context` argument when it runs `docker buildx build...`

```
docker buildx build .... --build-context=edx-platform=/path/to/local/repo/edx-platform
```

In the end, I figured out that my local repo had a residual `node_modules` directory that was empty and causing the `openedx-dev` image build to fail.

After removing the `node_modules` directory in my local repo, the `openedx-dev` image builds successfully.

I'll go a head and close this PR.

PR #1264, opened 2025-08-24
feat: add flag to skip reindexing

kaustavb12:
## Description

The [reindex commands](https://github.com/overhangio/tutor/blob/aef818912bbece5b5e4f38f8eb5273d5c6e7fa54/tutor/templates/jobs/init/cms.sh#L20-L22) which are run as part of the CMS init job, are not really required after the initial setup or initial move to meilisearch. Also, for very large instances, these jobs take multiple hours to multiple days to run, which makes it impossible to do deployments without commenting out the reindex steps.

This PR introduces a `SKIP_REINDEX` flag, which, when set to `true`, skips the re-index steps. The default value of this flag is `false` to preserve backwards compatibility, but it gives the instance operator the option to skip the re-index steps if they so choose.

## Testing instructions

Add `SKIP_REINDEX: true` to your config file and do a deployment. Verify that the CMS init job does not contain the re-index steps

## Other information

Private Ref: [BB-9778](https://tasks.opencraft.com/browse/BB-9778)


ahmed-arb (approved):
Thank you, @kaustavb12, for this contribution. It looks good to me.

regisb:
Hi @kaustavb12,
Adding a configuration setting to Tutor is always a red flag for me. In this particular case, it's a boolean flag that toggles a certain behaviour. I understand that this is an easy way to address the problem, but it's almost certainly not the "right" one. Instead, we could be doing one of the following:

1. Improve indexing performance
2. Only reindex what's necessary
3. On large platforms, skip reindex automatically and print a large warning saying user has to do it themselves

etc.

Basically, we should try to improve the default behaviour. Option 3 is probably the easiest, but I'd be interested in ideas to on options 1 and 2.

bradenmacdonald:
@kaustavb12  I thought you had already seen this, but maybe you haven't: https://github.com/openedx/edx-platform/issues/36868

It should be linked here.

bradenmacdonald:
> In tutor how would we detect that a platform is large?

The number of courses should be a reasonable proxy for the time it will take to run the index command.

PR #1267, opened 2025-08-26
Add auto-mounts for openedx-events and opaque-keys

bradenmacdonald:
Sometimes I need to work on repos like `openedx-events` or `opaque-keys`, and it's inconvenient to do so without Tutor knowing to automatically include those in all of the edx-platform images. This just adds them to the list of repos that are auto-detected as being used in the `edx-platform` images like LMS and CMS images.

With this change, users can just use e.g. `tutor mounts add openedx-events`

### DIFF
--- a/tutor/plugins/openedx.py
+++ b/tutor/plugins/openedx.py
@@ -111,6 +111,8 @@ hooks.Filters.MOUNTED_DIRECTORIES.add_items(
         ("openedx", "edx-ora2"),
         ("openedx", "edx-search"),
         ("openedx", "openedx-learning"),
+        ("openedx", "openedx-events"),
+        ("openedx", "opaque-keys"),
         ("openedx", r"platform-plugin-.*"),
     ]
 )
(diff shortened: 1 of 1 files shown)


tutor/plugins/openedx.py
```diff
@@ -111,6 +111,8 @@ def _mount_edx_platform_compose(
         ("openedx", "edx-ora2"),
         ("openedx", "edx-search"),
         ("openedx", "openedx-learning"),
+        ("openedx", "openedx-events"),
+        ("openedx", "opaque-keys"),
```
DawoudSheraz, line 115: What if instead of hard-coding, we add the regex, such as openedx-*? The same has been done for platform-plugin.

DawoudSheraz: I can't re-call anything right now. I think this should be good to go. 
@ahmed-arb What do you think?

bradenmacdonald: No it still only mounts repos that you explicitly request with `tutor mounts add`. This just tells tutor to make them available to all the `openedx` images: `lms`, `cms`, `lms-worker`, `cms-worker`, etc. and to install them when building python deps for those images.

DawoudSheraz (approved):

ahmed-arb (approved):

PR #1268, opened 2025-09-02
fix: improve security of random_string

samuelallan72:
## Description

Update the `random_string` function to use a cryptographically secure random number generator.

The `random_string` function is used for generating secrets, such as API keys or database passwords, so it should use a secure random number generator. The `random` module is a pseudo-random generator, which isn't suitable for this use case. Instead, switch to the `secrets` module, which provides the required randomness for this case.

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-9972

## Test instructions

- `secrets.choice` appears to be a drop-in replacement for `random.choice`. Please verify this is the case.
- Verify the `random_string` function generates secrets in the same way (code review + the unit tests should confirm this).

### DIFF
--- a/changelog.d/20250902_204307_samuel_improve_random_string.md
+++ b/changelog.d/20250902_204307_samuel_improve_random_string.md
@@ -1,0 +1,1 @@
+- [Security] Switch to a cryptographically secure number generator for the `random_string` function. (by @samuelallan72)
--- a/tutor/utils.py
+++ b/tutor/utils.py
@@ -3,8 +3,8 @@ import hashlib
 import hmac
 import json
 import os
-import random
 import re
+import secrets
 import shlex
 import shutil
 import string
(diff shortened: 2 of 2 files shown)


DawoudSheraz:
> @ahmed-arb sure, added! 🙂
> 
> I also saw this in https://docs.tutor.edly.io/tutor.html#contributing-to-tutor
> 
> > Target release if your change is compatible with the latest official Open edX release and it carries no major backwards-incompatibility nor risk of regression.
> 
> Please let me know if I should rebase on `release` for this patch. :)

Is this a breaking change? If not, then this branch can be rebased on release.

samuelallan72:
@DawoudSheraz I don't believe it would be a breaking change, so I've rebased on release. Thanks! :)

ahmed-arb (approved):

PR #1274, opened 2025-09-10
add option to provide a redis transport protocol

so-jd:
My redis server is set to only accept secure redis requests (i.e rediss:// vs redis://), and there was not easy way to set it, I made it variable with a default of `redis` so it will remain backwards compatible and will not introduce breaking changes.


DawoudSheraz (commented):
You can use openedx-common-settings patch to override settings.  I don't think the new config is needed since Tutor provides a way to override the behavior for one's particular requirements.

PR #1282, opened 2025-10-07
fix: Handle ImportError for Django deprecation warnings

feanil:
Wrap import and warning filter statements in a try-except block to handle potential ImportError gracefully. This can happen around django upgrade boundaries like the 4.2 -> 5.2 upgrade that's about to land.


feanil:
Updated the base to main for the PR.

PR #1290, opened 2025-10-21
[FC-0099] feat: add initialization tasks for the openedx-authz framework

mariajgrimaldi:
## Description

One of the main components of the Authorization Framework is the set of default policies that define how roles and permissions behave during authorization checks. These policies should be loaded at initialization to ensure they're available once the services are running and performing authorization checks.

This is what loading the default policies means:

1. The file https://github.com/openedx/openedx-authz/blob/main/openedx_authz/engine/config/authz.policy defines the default roles and permissions for managing content libraries.
2. To load these policies, run `python manage.py {lms|cms} load_policies`.
3. This command reads the `authz.policy` file and loads its contents into the `CasbinRules` table. It skips duplicates, making the command idempotent.

https://github.com/openedx/openedx-authz/issues/71

The authorization framework implemented here will start being used in the **Ulmo release**.

### DIFF
--- a/changelog.d/20251103_123634_maria.grimaldi_authz_load_policy.md
+++ b/changelog.d/20251103_123634_maria.grimaldi_authz_load_policy.md
@@ -1,0 +1,1 @@
+- [Feature] Add authorization policy loading functionality to LMS/CMS services. (by @mariajgrimaldi)


ahmed-arb (approved):
nit: please add a changelog entry.

mariajgrimaldi:
@ahmed-arb thanks! I'll test this one last time and add the changelog entry.

PR #1302, opened 2025-11-10
feat: Add ENV_SAVED Action

xitij2000:
Adds an ENV_SAVED Action hook that allows hooking into the save mechanism and running custom code after saving an environemnt.

### DIFF
--- a/changelog.d/20251111_011520_kshitij_env_saved_action.md
+++ b/changelog.d/20251111_011520_kshitij_env_saved_action.md
@@ -1,0 +1,2 @@
+- [Feature] Add ENV_SAVED Action that allows hooks to run after a save
+  operation. (by @xitij2000)
--- a/tutor/env.py
+++ b/tutor/env.py
@@ -342,6 +342,15 @@ def save(root: str, config: Config) -> None:
         save_all_from(src, os.path.join(root_env, dst), config)
 
     upgrade_obsolete(root)
+
+    config_copy = deepcopy(config)
+    hooks.Actions.ENV_SAVED.do(root_env, config_copy)
+    if config_copy != config:
+        fmt.echo_alert(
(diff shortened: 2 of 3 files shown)


tutor/env.py
```diff
@@ -342,6 +342,8 @@ def save(root: str, config: Config) -> None:
         save_all_from(src, os.path.join(root_env, dst), config)
 
     upgrade_obsolete(root)
+    # print("asdasd")
+    hooks.Actions.ENV_SAVED.do(root_env, config)
```
Kelketek: The docs you add mention that modifying this `config` variable won't result in the configuration being resaved and that it should be treated as read-only.

However, the config object being passed into this function is handed to this function. I'm not sure what the lifetime is of this variable outside of the function, but presumably, the configuration could be modified by this function and affect later code execution, even if the result isn't saved.

Should we add in a bit of enforcement here, by cloning the object before passing it? If so, might we do an equality check between the config before and after the change and add a warning saying the changes will not be saved?

tutor/env.py
```diff
@@ -333,6 +333,10 @@ def is_rendered(path: str) -> bool:
 )
 
 
+def deepcompare(val1, val2):
```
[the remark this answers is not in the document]
xitij2000: @Kelketek Yeah, I started implementing a more thorough compare mechanism that would show an actual diff, but realised it isn't worth it currently.

changelog.d/20251111_011520_kshitij_env_saved_action.md
```diff
@@ -0,0 +1,2 @@
+- [Feature] Add ENV_SAVED Action that allows hooks to run after a save operation.
```
Faraz32123: can you please add your GH username at the end of the change log entry.

PR #1305, opened 2025-11-16
feat: add K8S_CONTEXT configuration option

m-k8s:
## Summary
Add support for configuring Kubernetes context via `K8S_CONTEXT` setting, similar to the existing `K8S_NAMESPACE` configuration.

## Changes
- Add `K8S_CONTEXT` configuration option in defaults (empty by default)
- Update all kubectl commands to use `--context` flag when configured
- Update Kubernetes Python API client to load the specified context
- Add `kubectl_base_args()` helper function following the DRY principle

## Behavior
- When `K8S_CONTEXT` is empty (default): uses the current context from kubeconfig
- When `K8S_CONTEXT` is set: all kubectl commands and API calls use the specified context

## Testing
Tested with a fake context set as default in kubeconfig to ensure Tutor uses the configured `K8S_CONTEXT` instead of the kubeconfig's current-context.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -310,6 +310,10 @@ Kubernetes
 
 This configuration parameter sets the Kubernetes Namespace.
 
+- ``K8S_CONTEXT`` (default: ``""``)
+
+This configuration parameter sets the Kubernetes context to use for all kubectl commands and API calls. When empty (default), Tutor uses the current context from your kubeconfig file.
+
 Miscellaneous Project Settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


mlabeeb03:
Hey @m-k8s just checking in. If you can provide additional context of why you need it and what problems you are facing with `kubectl config use-context` then we can move forward with the review.

mlabeeb03:
Lets not add this config for now.

m-k8s:
Hi @mlabeeb03,

Apologies for the late response. Let me clarify the use case for this feature.

## Context

The need for this feature arises when working with **dozens or even hundreds of different Kubernetes contexts** across multiple projects simultaneously. In such scenarios, many users (including myself) prefer **not to switch the default context** using `kubectl config use-context`.

## Workflow

Instead, we systematically use the `--context` flag with every `kubectl` command to explicitly specify which cluster we're targeting:

```bash
kubectl --context prod-cluster-1 get pods
kubectl --context dev-cluster-2 apply -f deployment.yaml
kubectl --context staging-cluster-3 logs my-pod
```

## Benefits

This approach provides several advantages:

1. **Safety**: Avoids accidentally running commands on the wrong cluster when the default context hasn't been updated
2. **Clarity**: Makes it explicit which cluster each command targets, especially in scripts and automation
3. **Parallel workflows**: Allows working on multiple projects/contexts without constantly switching the default context back and forth

## Proposal

This PR would enable the same explicit context specification for Tutor's `k8s` commands, maintaining consistency with how we already use `kubectl` directly.

That said, I understand if this is too niche for inclusion in Tutor. Happy to discuss alternative approaches or keep this as a local patch if needed.

PR #1313, opened 2025-12-05
Ahtisham/add notification requirement

AhtishamShahid:
#### Description

This PR adds the Notifications plugin as the default offering  in tutor

### DIFF
--- a/requirements/plugins.txt
+++ b/requirements/plugins.txt
@@ -12,3 +12,4 @@ tutor-mfe@git+https://github.com/overhangio/tutor-mfe@main
 tutor-minio@git+https://github.com/overhangio/tutor-minio@main
 tutor-notes@git+https://github.com/overhangio/tutor-notes@main
 tutor-xqueue@git+https://github.com/overhangio/tutor-xqueue@main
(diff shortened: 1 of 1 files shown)


requirements/plugins.txt
```diff
@@ -12,3 +12,4 @@ tutor-mfe@git+https://github.com/overhangio/tutor-mfe@main
 tutor-minio@git+https://github.com/overhangio/tutor-minio@main
 tutor-notes@git+https://github.com/overhangio/tutor-notes@main
 tutor-xqueue@git+https://github.com/overhangio/tutor-xqueue@main
+tutor-contrib-notifications@git+https://github.com/openedx/tutor-contrib-notifications.git@main
```
ahmed-arb: There's a small inconsistency here: all the other plugins in this list are hosted under `overhangio/*`, but `tutor-contrib-notifications` is under `openedx/*`. 

Should we align the repository ownership before making this a default plugin? Or is there a reason to keep it under the `openedx` org?

Also, official plugins don't have `*contrib*` in their name.

sarina:
I think this PR can be closed because we're planning to make this a default within the platform (as opposed to a Tutor default) per recommendation from @regisb - is my understanding correct, @arbrandes ?

PR #1116, opened 2024-08-27
fix: uwsgi closed connection

Ian2012:
### Description

This PR modifies the uwsgi config file to improve lms/cms deployments. It follows the recommendations from [Bloomberg uWSGI production deployment](https://www.bloomberg.com/company/stories/configuring-uwsgi-production-deployment/) and solvesa 502 error based on the issue: https://github.com/unbit/uwsgi/issues/2133.

### The issue

Sometimes after a POST request, the connection between Caddy and the LMS/CMS pod is closed prematurely which returns a 502 error. this is fixed by setting `http-keepalive` plus manually adding the header `Connection: Keep-Alive`. Please take a look at this comment for more context: https://github.com/unbit/uwsgi/issues/1097#issuecomment-1323662862.

### Improvements

Following the best practices for production uwsgi deployments in docker we applied the following changes:

```
# Better startup/shutdown in docker
die-on-term = true
lazy-apps = false
need-app = true
no-defer-accept = true

# Other settings
master = true
py-call-osafterfork = true
vacuum = true
```

### Benchmarks

For the load test, we used k6 scripts with 100 virtual users who logged in to the platform and answered a problem. The target installation was running redwood in K8S.

You can find the k6 tests here: https://github.com/eduNEXT/openedx-k6-loadtesting

Notice how the 502 errors are reduced to 0 after the fixes
```
1. Default k8s deployments:

     data_received..................: 8.8 MB 64 kB/s
     data_sent......................: 294 kB 2.1 kB/s
     http_req_blocked...............: avg=54ms    min=455ns    med=1.13µs   max=802.08ms p(90)=2.17µs   p(95)=629.69ms
     http_req_connecting............: avg=11.24ms min=0s       med=0s       max=163.75ms p(90)=0s       p(95)=132.52ms
     http_req_duration..............: avg=3.9s    min=126.72ms med=146.21ms max=45.4s    p(90)=18.04s   p(95)=31.13s  
       { expected_response:true }...: avg=3.93s   min=126.72ms med=146.3ms  max=45.4s    p(90)=18.16s   p(95)=31.13s  
     http_req_failed................: 0.98%  ✓ 12       ✗ 1207 
     http_req_receiving.............: avg=23.08ms min=58.45µs  med=208.32µs max=234.44ms p(90)=129.57ms p(95)=137.2ms 
     http_req_sending...............: avg=176.7µs min=55.29µs  med=173.53µs max=523.46µs p(90)=246.48µs p(95)=268.91µs
[...]

PR #1321, opened 2025-12-31
[WIP] feat: replace uwsgi with granian

### LINKED ISSUE #937: We must replace uwsgi by something else

mlabeeb03:
PLEASE REACH OUT TO @ahmed-arb FOR UPDATES ON THIS PR.

closes #937 

This PR replaces uwsgi with [granian](https://github.com/emmett-framework/granian).

All uwsgi related docs have been updated. 
All uwsgi related patches and configs values have been renamed.
All uwsgi related files have been deleted and replaced with granian equivalents.

Here are the uwsgi configs/features we are currently using and their granian equivalents.

- [x] static-map = /static=/openedx/staticfiles/
- [x] static-map = /media=/openedx/media/
Even though `--static-path-route` & `--static-path-mount` are available, they can only server from one path, in order to server both of them we must use caddy.

- [x] http = 0.0.0.0:8000
`--host` & `--port`

- [x] buffer-size = 8192
--http1-buffer-size 8192

- [x] wsgi-file = $(SERVICE_VARIANT)/wsgi.py
Use this parameter `$SERVICE_VARIANT.wsgi:application`

- [x] processes = $(UWSGI_WORKERS)
`--workers`

- [x] ⚠️ thunder-lock = true
Apparently granian does not have thundering herd problem, see [Related Granian discussion](https://github.com/emmett-framework/granian/discussions/75)
- [x] single-interpreter = true
- [x] enable-threads = true
Threads are enabled by default. You can explicitly set them with `--runtime-mode`. Set the number of threads with `--runtime-threads`.
- [x] http-keepalive = 1
--http1-keep-alive enabled by default
- [x] ❌ add-header = Connection: Keep-Alive
Can not add arbitrary headers with granian, does this affect us if we have `--http1-keep-alive`?
- [x] die-on-term = true
Granian handles SIGTERM as a shutdown signal by default
- [x] lazy-apps = false
- [x] need-app = true
- [x] no-defer-accept = true
- [x] master = true
- [x] py-call-osafterfork = true
- [x] vacuum = true

- [x] Reload
We can define a file in the `--reload-paths` and changing that file will restart the server. We also need to install the `reload` dependency and also unable the `--reload` config.


ormsbee:
Just a warning that I have a suspicion that MIT's switch to granian _may_ have been responsible for exposing [a multi-threading related bug in grading](https://github.com/openedx/edx-platform/pull/37825). This wouldn't be any fault of granian's, but more a case where granian's deployment strategy exposes already-existing problems. This is *highly speculative* on my part, and the investigation is ongoing. I just wanted to bring this to your attention early. The bug needs to be fixed regardless, so I don't expect it to block granian in the long term, even if this hypothesis turns out to be true.

tutor/templates/build/openedx/settings/granian-config
```diff
@@ -0,0 +1,10 @@
+--interface wsgi
+--host 0.0.0.0
+--port 8000
+--workers $GRANIAN_WORKERS
+--static-path-route /static
+--static-path-mount /openedx/staticfiles/
+--reload
+--reload-paths /openedx/granian-reload-path
```
ormsbee: This is for watching a directory and reloading if there are file changes detected, right? Just out of curiosity, why enable this in a prod environment?

ormsbee:
Quick update: We believe the multi-threading issue has been fixed, though we're still waiting for MIT to deploy it on their instance next week to verify that it resolves the issues they've been seeing there.

tutor/templates/build/openedx/settings/granian-config
mlabeeb03: We had a `reload-uwsgi` command as well. [Here](https://github.com/overhangio/tutor/blob/b2bc431fee59a878de1fd752a841a8f945adef61/docs/local.rst#reloading-open-edx-settings) is the rational behind doing it. This is just a replacement for that. Now the `reload-granian` command puts current timestamp in the file mentioned in `--reload-paths` to trigger a reload.

Danyal-Faheem:
The [issue](https://github.com/emmett-framework/granian/issues/747) of Granian not being able to manage serving from multiple paths has been fixed with this [PR](https://github.com/emmett-framework/granian/pull/794) as of granian [v2.7.0](https://github.com/emmett-framework/granian/releases/tag/v2.7.0).

This means we can move forward with the current approach if we can find alternatives to the other options.

HammadYousaf01:
I've bumped Granian to [2.7.4](https://github.com/emmett-framework/granian/releases/tag/v2.7.4). I verified that it now serves static files correctly for both /media and /static.

As for the `add-header = Connection: Keep-Alive`, we don't need it now. That header was added to work around the uwsgi issue in https://github.com/overhangio/tutor/pull/1116. Granian does not have that bug, and its HTTP/1.1 keep-alive behavior is correct by default.

I validated this with the same load test benchmark used in the earlier [PR](https://github.com/overhangio/tutor/pull/1116). Granian completed the test without any 502 responses, so I think we can mark this action item done as well.

`http_req_failed................: 0.00% 0 out of 9272`
`http_reqs......................: 9272  21.832574/s`
`{ status:502 }...............: 0     0/s`

The only remaining action items are the flags, and we still need to scope them and decide what to do with them.

tutor/templates/build/openedx/bin/run-granian
```diff
@@ -0,0 +1,33 @@
...
+                # flags without values
+                args.append(line)
+    return args
+
+
+def run():
+    cli(args=parse_args_file(), standalone_mode=False)
+
+if __name__ == "__main__":
+    run()
```
ormsbee, line 33: In general, please add a trailing new line in your files.

tutor/templates/build/openedx/bin/run-granian
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
ormsbee, line 1: Please add a docstring explaining why this is necessary (I realize it's because granian doesn't support a config file format, but this might be confusing to people who don't know that.)

tutor/templates/build/openedx/bin/run-granian
```diff
@@ -0,0 +1,33 @@
...
+                args.append(option)
+                args.append(value)
+            else:
+                # flags without values
+                args.append(line)
+    return args
+
+
+def run():
+    cli(args=parse_args_file(), standalone_mode=False)
```
ormsbee, line 30: Likewise, please comment what `standalone_mode=False` is, so that people understand that it's related to `click` and the process, and not a separate arg for granian.

tutor/templates/build/openedx/settings/granian-config
```diff
@@ -0,0 +1,12 @@
+--interface wsgi
+--host 0.0.0.0
+--port 8000
+--workers $GRANIAN_WORKERS
```
ormsbee, line 4: Not blocking, but I'm curious if you folks are planning to experiment with different threading values?

tutor/templates/build/openedx/settings/granian-config
```diff
@@ -0,0 +1,12 @@
...
+--static-path-route /static
+--static-path-mount /openedx/staticfiles/
+--static-path-route /media
+--static-path-mount /openedx/media/
+--reload
+--reload-paths /openedx/granian-reload-path
+--http1-buffer-size 8192
```
ormsbee, line 11: This seems really low? Is this because we're not actually expecting to get HTTP 1.1 requests?

ormsbee (commented):
A few more questions, but it's looking great. I'm so glad to hear that we can map multiple static dirs now.

> The only remaining action items are the flags, and we still need to scope them and decide what to do with them.

Do you mean that you're still looking at the exact config values that should ship by default, or does this mean something else?

Thank you!

PR #1335, opened 2026-01-21
Update python version.

feanil:
The edx-platform requirements are now building correctly on both version of Python.


feanil:
@ahmed-arb sounds good I'll just drop that commit from this PR.

PR #1348, opened 2026-02-10
fix: Cherry-pick three critical ulmo.1 fixes

kdmccormick:
## Description

Bumps tutor version from 21.0.0 to 21.0.1

Cherry-picks in:

* https://github.com/openedx/openedx-platform/pull/37949
* https://github.com/openedx/openedx-platform/pull/37991
* https://github.com/openedx/openedx-platform/pull/37975

## Changelog

@overhangio/tutor-maintainers  - I haven't bumped a tutor release version myself in a while. Is this the correct way to do it? Do I need to compile the changelog or anything?

## Testing

I haven't tested this yet. @overhangio/tutor-maintainers , I could use your help testing the build if you have time.

## Notes

* I'd prefer if [ulmo.2 were released now](https://discuss.openedx.org/t/a-proposal-for-btr-tag-point-releases-whenever-theyre-needed/18416) so that we could just update OPENEDX_COMMON_VERSION instead of cherry-picking these in, but I don't think we should wait on that discussion in order to release these patches.

* These patches aren't needed on Main-mode since they have all been merged to openedx-platform master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@ instructions, because git commits are used to generate release notes:
 
 <!-- scriv-insert-here -->
 
+<a id='changelog-21.0.1'></a>
+## v21.0.1 (2026-02-11)
+
+- [Feature] Add support for mounting `openedx-platform`, the which is new name of `edx-platform`. Mounting by the old name, `edx-platform`, will continue to work for now as well (by @kdmccormick).
+
+- [Bugfix] Upgrade edx-search to 4.4.0 to fix Catalog MFE (by @kdmccormick)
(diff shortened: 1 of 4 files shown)


kdmccormick:
Thanks @ahmed-arb !

I've pushed a new commit with the build changelog.

I've also dropped this entry from the changelog: https://github.com/overhangio/tutor/pull/1348/changes/1acbb84ad2e0dda71b999478af7399177615b9f9 . My reasoning is that the Changelog is for people who use tutor, not for us Tutor maintainers (the commit log is for us maintainers). The users of tutor don't need to know that we updated the version of a github action runner's OS. Let me know if you disagree, and I'll put it back.

ahmed-arb (approved):

PR #1366, opened 2026-04-09
fix: apply upstream security patch to remove activation_key exposure from account API

Gi-ron:
### Description
This PR applies an upstream security fix from Open edX by backporting the following commit into the Tutor Open edX Docker image, which removes the activation_key field from the /api/user/v1/accounts/{username} response to prevent a vulnerability where attackers could bypass email verification by combining OAuth2 password grant access for inactive users with direct account activation:

https://github.com/openedx/openedx-platform/commit/21cead238466ca398ba368518f1d3288431d68f4

This issue is part of the security advisory discussed here:
https://discuss.openedx.org/t/security-upcoming-security-release-for-openedx-platform-2026-03-27/18655

This PR is proposed because the fix is not yet included in an official Open edX Ulmo release used by Tutor. Applying it at build time ensures that Tutor deployments are protected until the fix is available upstream in a stable release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ instructions, because git commits are used to generate release notes:
 
 <!-- scriv-insert-here -->
 
+<a id='changelog-21.0.4'></a>
+## v21.0.4 (2026-04-10)
+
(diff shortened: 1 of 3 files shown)


Gi-ron:
@regisb could you please take a look at this security backport?

ahmed-arb:
Thank you @Gi-ron, could you please verify if you have tested image builds with this patch?

Gi-ron:
@ahmed-arb Yes I have tested the image builds with this patch and everything is working correctly. No issues found.

ahmed-arb:
Okay, merging it then.

PR #1375, opened 2026-04-29
fix: fixed wrong lms_host

### LINKED ISSUE #1372: First dev setup with wrong lms_host
It's an old one, but running tutor dev in non interactive will result in wrong lms host:
How to reproduce:

```bash
# in a venv
pip install tutor tutor-mfe tutor-indigo
tutor config save
tutor dev launch -I
```
will result in not setting the LMS_HOST to local.openedx.io, and we end up in

```
All services initialised.
The platform is now running and can be accessed at the following urls:

    http://www.myopenedx.com:8000
    http://studio.www.myopenedx.com:8001
    http://meilisearch.www.myopenedx.com:7700
    http://apps.www.myopenedx.com:2025/admin-console
    http://apps.www.myopenedx.com:1999/authn
    http://apps.www.myopenedx.com:2001/authoring
    http://apps.www.myopenedx.com:1997/account
    http://apps.www.myopenedx.com:1984/communications
    http://apps.www.myopenedx.com:2002/discussions
    http://apps.www.myopenedx.com:1994/gradebook
    http://apps.www.myopenedx.com:1996/learner-dashboard
    http://apps.www.myopenedx.com:2000/learning
    http://apps.www.myopenedx.com:1993/ora-grading
    http://apps.www.myopenedx.com:1995/profile

```

muhammadadeeltajamul:
Fix : [Issue#1372](https://github.com/overhangio/tutor/issues/1372)

Fixes: Running `tutor dev launch -I` sets default `LMS_HOST` not `local.openedx.io`

### DIFF
--- a/tutor/commands/compose.py
+++ b/tutor/commands/compose.py
@@ -228,6 +228,8 @@ def interactive_configuration(
             config,
             run_for_prod=run_for_prod,
         )
+    elif not run_for_prod:
+        interactive_config.set_run_mode_defaults(config)
     tutor_config.save_config_file(context.obj.root, config)
     config = tutor_config.load_full(context.obj.root)
     tutor_env.save(context.obj.root, config)
--- a/tutor/interactive.py
+++ b/tutor/interactive.py
@@ -7,6 +7,16 @@ from . import env, exceptions, fmt, hooks
 from .types import Config, get_typed
 
 
+def set_run_mode_defaults(config: Config) -> None:
+    dev_values: Config = {
+        "LMS_HOST": "local.openedx.io",
+        "CMS_HOST": "studio.local.openedx.io",
+        "ENABLE_HTTPS": False,
+    }
+    for key, value in dev_values.items():
+        config[key] = value
+
+
 def ask_questions(config: Config, run_for_prod: Optional[bool] = None) -> None:
     """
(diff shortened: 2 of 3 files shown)


tutor/interactive.py
```diff
@@ -7,6 +7,18 @@
 from .types import Config, get_typed
 
 
+def set_run_mode_defaults(config: Config, run_for_prod: bool) -> None:
+    if run_for_prod:
+        return
```
Faraz32123: isn't this condition redundant when we are already checking `run_for_prod` before calling the function.

tutor/commands/compose.py
```diff
@@ -228,6 +228,8 @@ def interactive_configuration(
             config,
             run_for_prod=run_for_prod,
         )
+    elif run_for_prod is not None:
```
Faraz32123: nit: why didn't we specifically used `elif not run_for_prod` and remove the if condition from inside the function because it will return early from the function.

Faraz32123 (changes requested):
Can we also add the changelog entry?

muhammadadeeltajamul:
> Can we also add the changelog entry?

It can be added in next release changelog

Faraz32123 (approved):

ahmed-arb (approved):

PR #1376, opened 2026-04-29
fix: apply activation_key security patch only in release mode

ahmed-arb:
### Description

Follow-up to #1366. The security patch added in that PR fails when building against `master` edx-platform because the upstream commit is already present there, so `git am` fails with "patch already applied". This breaks builds on Tutor Main.

This PR moves the RUN command inside the non-Main (Release mode) branch of the existing `TUTOR_BRANCH_IS_MAIN` conditional, matching the pattern that the template already establishes for git patches.

### Testing

- Build openedx image in release mode (`EDX_PLATFORM_VERSION` pointing to `release/ulmo.x`): patch applies as before.
- Build openedx image in main mode (`EDX_PLATFORM_VERSION=master`): patch is skipped, build no longer fails.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tutor/templates/build/openedx/Dockerfile
+++ b/tutor/templates/build/openedx/Dockerfile
@@ -58,14 +58,14 @@ RUN git config --global user.email "tutor@overhang.io" \
 # Patches in Main node
 {%- else %}
 # Patches in non-Main mode (i.e., Release mode)
+
+# SECURITY FIX: remove activation_key exposure from account API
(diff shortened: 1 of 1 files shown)


Agrendalath:
@ahmed-arb, could you please create a release that includes this change? I'd like to remove this patch locally via `openedx-dockerfile-git-patches-default`.
The builds are failing for `release/ulmo.3`, because the security fix is already present there.

ahmed-arb:
> The builds are failing for `release/ulmo.3`, because the security fix is already present there.

@Agrendalath, I am making a new release to fix this: https://github.com/overhangio/tutor/pull/1379

PR #1377, opened 2026-04-29
fix!: Raise exceptions when a plugin fails to load.

xitij2000:
When a plugin is missing or throws an error while loading, Tutor continues as normal. At times this has made us recognise issues far later than we'd like. 

This change will raise an error causing the command to return an error and the break the CI flow rather than deploying a broken instance with missing plugins. 

We've considered few alternatives. 

1. Put this behind an option flag such as `--check`, however this flag would need to be added in a lot of places that cause plugins to load. 
2. Adding another command like `tutor doctor` this is a convention that many such tools follow that check the config and environment for errors and suggest fixes. This issue with this is that it will require updating existing usage of tutor to call tutor doctor first. 
3. Creating a plugin for this. We think this is a core feature hat should be part of tutor, but this could potentially be done in a plugin.

### DIFF
--- a/tests/commands/test_cli.py
+++ b/tests/commands/test_cli.py
@@ -1,6 +1,9 @@
 import unittest
+from unittest.mock import patch
 
+from tutor import hooks
 from tutor.__about__ import __version__
+from tutor.commands.cli import check_plugin_errors, cli, main
 
 from .base import TestCommandMixin
 
@@ -21,3 +24,43 @@ class CliTests(unittest.TestCase, TestCommandMixin):
         self.assertEqual(0, result.exit_code)
         self.assertIsNone(result.exception)
         self.assertRegex(result.output, rf"cli, version {__version__}\n")
+
+    def test_ignore_plugin_errors_flag_in_help(self) -> None:
+        """The --ignore-plugin-errors flag should appear in help output."""
+        result = self.invoke(["--help"])
+        self.assertEqual(0, result.exit_code)
+        self.assertIn("ignore-plugin-errors", result.output)
+
+
+class PluginErrorExitCodeTests(unittest.TestCase):
+    """Tests for plugin error exit code behavior."""
+
+    def _add_plugin_error(self) -> None:
+        """Add a test plugin error."""
+        hooks.Filters.PLUGIN_ERRORS.add_item(("badplugin", "test error"))
+
+    def _clear_plugin_errors(self) -> None:
+        """Clear all plugin errors by clearing all callbacks."""
+        hooks.Filters.PLUGIN_ERRORS.clear()
+
+    def setUp(self) -> None:
+        self._clear_plugin_errors()
+        self.addCleanup(self._clear_plugin_errors)
+
+    def test_check_plugin_errors_exits_with_code_1(self) -> None:
+        """_check_plugin_errors should exit with code 1 when errors exist."""
+        self._add_plugin_error()
+        with self.assertRaises(SystemExit) as cm:
+            check_plugin_errors(False)
+        self.assertEqual(cm.exception.code, 1)
+
+    def test_check_plugin_errors_no_exit_when_ignored(self) -> None:
+        """_check_plugin_errors should not exit when --ignore-plugin-errors is set."""
+        self._add_plugin_error()
+        # Should not raise
+        check_plugin_errors(True)
+
+    def test_check_plugin_errors_no_exit_when_no_errors(self) -> None:
+        """_check_plugin_errors should not exit when there are no errors."""
+        # Should not raise
+        check_plugin_errors(False)
--- a/tests/commands/test_cli.py
+++ b/tests/commands/test_cli.py
@@ -46,7 +46,9 @@ class PluginErrorExitCodeTests(PluginsTestCase):
             hooks.Filters.PLUGIN_ERRORS.add_item(("badplugin", "test error"))
 
     @contextmanager
-    def _prev_config(self, ignore_errors=False) -> Iterator[tuple[Context, Config]]:
+    def _prev_config(
+        self, ignore_errors: bool = False
+    ) -> Iterator[tuple[Context, Config]]:
         with temporary_root() as root:
             context = Context(root, ignore_errors)
             yield context, tutor_config.get_user(root)
--- a/tests/commands/test_cli.py
+++ b/tests/commands/test_cli.py
@@ -1,6 +1,12 @@
 import unittest
+from contextlib import contextmanager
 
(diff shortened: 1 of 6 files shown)


xitij2000:
Note this PR is to get the conversation started, we can improve the error messaging etc.

Kelketek:
@xitij2000 It looks like you can't disable the plugin using the command line after running into this problem. I could try modifying the config, but I can't ask it for the config root, either. Is it possible to keep this a warning for configuration commands but an error for things like image building and launching?

xitij2000:
> @xitij2000 It looks like you can't disable the plugin using the command line after running into this problem. I could try modifying the config, but I can't ask it for the config root, either. Is it possible to keep this a warning for configuration commands but an error for things like image building and launching?

I'll look into that. For now I just wanted to get some feedback. Is there is a reason to not do this at all? Or move it to a separate command or plugin.

ahmed-arb:
The fail-fast goal makes sense, but raising at the plugin-loading layer is too low: `ensure_plugins_enabled` runs on every invocation, so a plugin that fails to load doesn't just break launch; it also breaks:

- `tutor plugins disable <bad-plugin>`
- `tutor config`
- `tutor --help`

The commands you'd use to recover all trigger plugin loading first, so the user gets bricked into hand-editing `config.yml` or running `pip uninstall` (which is what @Kelketek hit above).

Also worth noting: the raise lives inside the `for name in names` loop in `load_all`, so the first failing plugin aborts loading of all remaining plugins and skips `PLUGINS_LOADED.do()`. As a result, one broken plugin takes down healthy ones too.

I'd suggest keeping loading (warn + record failures) and moving the hard failure up to deployment/build commands (`launch`, `start`, `images build`) while leaving inspection/recovery commands as warnings.

xitij2000:
@ahmed-arb Thanks for having a look! 

I agree with all the points you have raised, I just wanted to start discussing this issue with this PR. I plan to make the solution more robust when I get back to this.

tests/commands/test_cli.py
```diff
@@ -21,3 +24,43 @@ def test_cli_version(self) -> None:
...
+        hooks.Filters.PLUGIN_ERRORS.clear()
+
+    def setUp(self) -> None:
+        self._clear_plugin_errors()
+        self.addCleanup(self._clear_plugin_errors)
+
+    def test_check_plugin_errors_exits_with_code_1(self) -> None:
+        """_check_plugin_errors should exit with code 1 when errors exist."""
+        self._add_plugin_error()
+        with self.assertRaises(SystemExit) as cm:
```
Kelketek: Oh, cool. I didn't know you could catch sys.exit like this.

Kelketek:
@xitij2000 Thanks for these updates. Image building succeeds, though it should be one of the commands that fails if an enabled plugin is missing, in my opinion. Actually, launch also gets off the ground right now. Can you double-check that things are actually failing for you where they need to be?

CI is also failing.

You noted to me over our internal ticket that this set of errors would cause a freshly installed tutor launch to fail due to installing only tutor leaving out plugins like mfe and indigo. However, I actually think this is preferable-- if someone accidentally installs only tutor and launches a brand new install, they're not following the tutorial closely and have almost certainly made a mistake.

Anyone who is deliberately only installing Tutor will know what to do about this because they'll be familiar, and it's very likely they already have a config.yml file without those entries in the plugin list.

xitij2000:
> @xitij2000 Thanks for these updates. Image building succeeds, though it should be one of the commands that fails if an enabled plugin is missing, in my opinion. Actually, launch also gets off the ground right now. Can you double-check that things are actually failing for you where they need to be?
> 
> CI is also failing.
> 
> You noted to me over our internal ticket that this set of errors would cause a freshly installed tutor launch to fail due to installing only tutor leaving out plugins like mfe and indigo. However, I actually think this is preferable-- if someone accidentally installs only tutor and launches a brand new install, they're not following the tutorial closely and have almost certainly made a mistake.
> 
> Anyone who is deliberately only installing Tutor will know what to do about this because they'll be familiar, and it's very likely they already have a config.yml file without those entries in the plugin list.

Now command should run but produce an error at the end and return a error code. While this is not ideal, raising an error first means that you can't even run a command to fix the error, as was the case before. 

This way if you apply a config change and that raises a plugin error it will raise immediately. 

While I agree about the missing plugins raising an error by default it is a breaking change and I'd want some upstream feedback on this.

tutor/commands/cli.py
```diff
@@ -32,6 +33,45 @@ def main() -> None:
...
+    if has_config_changed:
+        fmt.echo_alert(
+            "The Config file was changed by running this command "
+            "so the above errors might no longer be valid. \n"
+            "Please run 'tutor config save' to check if the errors persist."
+        )
+    # Only exit with an error code if:
+    #   - errors are not being ignored
+    #   - the config file hasn't changed
+    #   - no exception has already been raised that would return its own error codek
```
Kelketek, line 63:
```suggestion
    #   - no exception has already been raised that would return its own error code
```

Kelketek (changes requested):
@xitij2000 It is indeed failing commands after running them if there is a plugin error. However, I'm still unhappy with it going all the way through commands like image building and launching. It greatly undermines the utility of the change, as the original reason we bothered with this at all was to avoid waiting hours just to find out that the build couldn't be correct.

Note also that if an image building succeeds, that image is saved, and so it will be the image rolled next restart. Someone could deploy an image they don't want to because they haven't fixed it and rebuilt. That shouldn't be possible-- if the configuration is bad and we're returning an error code, we don't want the state to be left worse than when we found it.

You mentioned needing input from upstream on this, bearing in mind that the default settings would break in the case that someone installed Tutor and initialized a new configuration while not following the tutorial and instead running the commands raw. I'm still of the mind that this is fine, given my reasoning above.

However, I want to ping @ahmed-arb and get his take. @ahmed-arb , what are your thoughts on [my suggestions here](https://github.com/overhangio/tutor/pull/1377#issuecomment-4877690464)?

xitij2000:
@Kelketek I think the only solution for this is to special case certain commands to run despite plugin errors, such as config editing, and enabling / disabling plugins.

Kelketek:
@xitij2000 Maybe it could be done by a special 'configuration_exempt' decorator which adds an attribute to the function which the preflight can check for? That way any other plugin that includes a command that should be able to run with broken plugins can use it.

xitij2000:
@Kelketek I think that approach makes sense. The decorator can simply add an attribute to the function which can be looked up and for functions that are marked this way we can skip the early exit.

Issue #1218, opened 2025-04-07
Onboarding Developer Documentation


tonybusa:
**Is your feature request related to a problem? Please describe.**
I am working to onboard 10 engineers (myself included) to the Open edX platform and each individual has struggled through the tutor documentation for development. Particularly, the initial getting started documentation seem to have opportunity for consolidation and/or refactoring.

**Describe the solution you'd like**
Here are the two pages that I see the best opportunity for consolidation:
- https://docs.tutor.edly.io/dev.html#open-edx-development
- https://docs.tutor.edly.io/tutorials/edx-platform.html#working-on-edx-platform-as-a-developer
There are important explanations in each of these pages but there are a few inconsistencies that I think could be cleared up. For instance:
- The "tutorials" page explains step-by-step how to reference your local workspace, while the "first-time" page assumes you have it already forked.
- The "tutorials" page explains what "tutor mounts add ..." does while the "first-time" page explains what "tutor dev launch" does.
- Both documentation references "tutor local" halfway through the page but neither links to or explains what the difference between "tutor local" and "tutor dev".
- Lastly, I think the "first-time" document should reference the "tutorials" link _at the _end_ of the documentation and so it flows into that page as next steps, rather than two docs that start off similarly.

I also believe it would be better to include this page about installing tutor main, as a sub-page or section to the installing tutor documentation:
- (Tutor main): http://docs.tutor.edly.io/tutorials/main.html#installing-tutor-main
- (Tutor install): https://docs.tutor.edly.io/install.html#installing-tutor
The reason being was by the time I found out what tutor main was in the tutorial I was confused as to why I would want to be developer contributing to open edx and not working on the master branch.

**Describe alternatives you've considered**
Alternatively, there could be a clearer navigation menu that would include ability to see more sub-pages at once rather than just the one selected. I had hard time realizing when I was moving into a new section when I clicked on different links throughout. 
Or there could be a one-stop-shop menu for all engineering / developer related content for contributing to the platform.

**Additional context**
I have now been able to run a release branch of the platform locally and make changes. However, I still had questions about what really was tutor doing. I found this video extremely helpful in solidifying my understanding of not just what tutor is, but why it came about: https://www.youtube.com/watch?v=BzNcrmyFpw4. I would recommend having this video embedded in one of the first pages on the tutor documentation.

[...]

PR #1385, opened 2026-05-08
docs: add instructions for running migrations

Danyal-Faheem:
One of the steps in improving our developer documentation as part of #1218. This was raised by Axim where they mentioned we have no mention of how to run migrations in our docs using tutor.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -366,6 +366,14 @@ Then, the ``openedx`` docker image must be rebuilt::
 
     tutor images build openedx
 
+Then, restart tutor for the changes to take effect::
+
+    tutor local start -d
+
+If your extra requirements have database migrations, those need to be run as well. If you are unsure of whether this is required, it is always good to run this command as it is idempotent::
+
+    tutor local do init
+
 .. _edx_platform_fork:
 
 Running a fork of ``edx-platform``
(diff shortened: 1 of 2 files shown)


Danyal-Faheem:
@ayub02 Can you take a look at this?

docs/configuration.rst
```diff
@@ -366,6 +366,14 @@ Then, the ``openedx`` docker image must be rebuilt::
...
     tutor images build openedx
 
+Then, restart tutor for the changes to take effect::
+
+    tutor local start -d
+
+If your extra requirements have database migrations, those need to be run as well. If you are unsure of whether this is required, it is always good to run this command as it is idempotent::
+
+    tutor local do init
+
```
ahmed-arb, line 376: I think we should move the init part before the restart command. If the user needs to run init, they should do it before the restart.

docs/configuration.rst
```diff
@@ -366,6 +366,14 @@ Then, the ``openedx`` docker image must be rebuilt::
 
     tutor images build openedx
 
+Then, restart tutor for the changes to take effect::
+
+    tutor local start -d
```
ahmed-arb: If we're suggesting a restart, shouldn't the command be:
```bash
tutor local reboot -d
```

docs/configuration.rst
Danyal-Faheem: For this, I guess it makes more sense to restart the platform with the updated image before applying any migrations.

docs/configuration.rst
Danyal-Faheem: Sure, I will update this.

Danyal-Faheem: Updated.

PR #1387, opened 2026-05-11
feat: add TESTS filter

### LINKED ISSUE #1386: Feature Request: Add a pluggable format to include smoke/integration/e2e tests for tutor and plugins
**Is your feature request related to a problem? Please describe.**

Running tutor is seamless thanks to its one click policy, however, after setup, users tend to find that their system is currently not running perfectly or has issues. This is especially evident when one is developing using tutor where latest (main) branches tend to break certain functionalities. Finding these bugs requires testing by pulling new branches, building images then manually testing certain features.

While some work has been done and is in active development for a full E2E testing suite for Open edX, reference: https://github.com/openedx/cypress-e2e-tests, this work is still in its infancy and requires much work before it can be considered "production ready". And, there is currently no way to run these e2e tests on tutor other than manual confugration and setup either.

This would also streamline our bi-annual release process for Open edX as well as for minor releases where we are constantly looking for testing volunteers. Of course, we would still require testing but adding tests would hasten the process considerably.

**Describe the solution you'd like**

An extendable and pluggable Filter that allows tutor and tutor plugin developers to add their own tests, be it smoke, integration or e2e tests that can be executed using a single command, keeping within standard of the current one click tutor install procedure.

The exact technical details can be discussed over but an extendable format where plugins can add their own tests and tests written across all plugins can be executed using a single idempotent command is what I would like, similar to how the init jobs currently work.

These tests could be as simple as a curl healthcheck call or something as complicated as running a full e2e testing suite, akin to that to of https://github.com/openedx/cypress-e2e-tests. 

It would also allow developers who have made customizations on top of Open edX using Open edX and Tutor plugins to seamlessly integrate their testing suites within tutor rather than by running management commands through the shell one by one.

Danyal-Faheem:
fixes #1386.

We add the following changes:

1. Add a new `TESTS` Filter that will accumulate tests across tutor core and plugins.
2. Add a new `tutor local do tests [suite] --limit=<service>` command to run tests across core and plugins. For example, to run smoke tests across plugins: `tutor local do tests smoke`
3. Adds smoke tests for tutor core that include:
i. User Management and Authentication
ii. Course import, course creation
iii. User enrollment
iv. Basic healthchecks
v. OAuth Application creation


tutor/commands/jobs.py
```diff
@@ -467,6 +476,169 @@ def update_mysql_authentication_plugin(
...
+        fmt.echo_alert(
+            f"About to delete smoke test artifacts from the database: "
+            f"user '{smoke_username}', course '{smoke_course_id}'."
+        )
+        if not non_interactive:
+            if not click.confirm("Proceed with cleanup?", prompt_suffix=" "):
+                cleanup = False
+    if cleanup:
+        yield ("lms", tests_teardown_lms_template(merged["SMOKE_USERNAME"]))
+        yield ("cms", tests_teardown_cms_template(merged["SMOKE_COURSE_ID"]))
```
Faraz32123, line 645: As `TEST_DEFAULTS` doesn't have `SMOKE_USERNAME` & `SMOKE_COURSE_ID`, can we make sure cleanup job doesn't fail with key error? Or I missed something?

Danyal-Faheem: I was testing this with some smoke tests included and I removed them for the PR. I think I will add them to the PR itself which would make it easier to understand.

ahmed-arb (approved):

ahmed-arb:
Let's resolve conflicts and merge this.

ahmed-arb (changes requested):
Just noticed, changelog entry is missing.

Danyal-Faheem:
Added changelog entries and resolved merge conflicts.

ahmed-arb (approved):

PR #1389, opened 2026-05-20
fix: fixed wrong site in django shell

muhammadadeeltajamul:
Fixes incorrect `SITE_ID` when site is accessed from management command or celery task. 
After this change, it will return `LMS_BASE` site object when command or task is in lms and `CMS_BASE` site object when command or task is in cms instead of `example.com` site object

### DIFF
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -20,11 +20,15 @@ def _configure_site_id():
     from django.conf import settings
     from django.contrib.sites.models import Site
 
-    site, _ = Site.objects.get_or_create(
-        domain=domain,
-        defaults={"domain": domain, "name": domain},
-    )
-    settings.SITE_ID = site.pk
+    try:
+        site, _ = Site.objects.get_or_create(
+            domain=domain,
+            defaults={"name": domain},
+        )
+        settings.SITE_ID = site.pk
+    except Exception:
+        # use default SITE_ID. This can happen if the database is not yet migrated, or if the Sites framework is not installed.
+        pass
 
 
 def _patch_django_setup_with_site_id():
--- a/tutor/templates/apps/openedx/settings/partials/common_all.py
+++ b/tutor/templates/apps/openedx/settings/partials/common_all.py
@@ -4,6 +4,46 @@ import os
 
 from xmodule.modulestore.modulestore_settings import update_module_store_settings
 
+
+def _get_default_site_domain():
+    service_variant = os.environ.get("SERVICE_VARIANT", "lms")
+    if service_variant == "cms":
+        return globals().get("CMS_BASE")
+    return globals().get("LMS_BASE")
+
+
+def _configure_site_id():
+    domain = _get_default_site_domain()
+    if not domain:
+        return
+
+    from django.conf import settings
+    from django.contrib.sites.models import Site
+
+    site, _ = Site.objects.get_or_create(
+        domain=domain,
+        defaults={"domain": domain, "name": domain},
+    )
+    settings.SITE_ID = site.pk
+
+
+def _patch_django_setup_with_site_id():
+    import django
+
+    if getattr(django.setup, "_tutor_site_id_patched", False):
+        return
+
+    original_setup = django.setup
+
+    def setup_with_site_id(*args, **kwargs):
+        result = original_setup(*args, **kwargs)
+        _configure_site_id()
+        return result
+
+    setup_with_site_id._tutor_site_id_patched = True
+    django.setup = setup_with_site_id
+
+
 # Mongodb connection parameters: simply modify `mongodb_parameters` to affect all connections to MongoDb.
 mongodb_parameters = {
     "db": "{{ MONGODB_DATABASE }}",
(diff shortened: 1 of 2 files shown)


tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -4,6 +4,46 @@
...
+    return globals().get("LMS_BASE")
+
+
+def _configure_site_id():
+    domain = _get_default_site_domain()
+    if not domain:
+        return
+
+    from django.conf import settings
+    from django.contrib.sites.models import Site
```
Faraz32123, line 21: Can we move these imports to top of the file? or any reason to add it here?

tutor/templates/apps/openedx/settings/partials/common_all.py
```diff
@@ -4,6 +4,46 @@
...
+
+    site, _ = Site.objects.get_or_create(
+        domain=domain,
+        defaults={"domain": domain, "name": domain},
+    )
```
Faraz32123: can we remove domain from from defaults? as it is already used for lookup.

tutor/templates/apps/openedx/settings/partials/common_all.py
muhammadadeeltajamul: These imports are inside function so that model and settings gets imported once the django.setup is completed. If I move these imports to the top, it will throw error that app is not loaded.

tutor/templates/apps/openedx/settings/partials/common_all.py
Faraz32123: [Question] Did we tried a fresh setup with it? When there are no migration applied, just to be sure?

muhammadadeeltajamul: its a small nit. I write this way for better readability but it will work either way.

muhammadadeeltajamul: I have not tried it with the new setup. I tested it on running instances of `tutor local` and `tutor dev`  for both cms and lms.

muhammadadeeltajamul: Fixed nit.
`launch` was failing because site model was not migrated. I have fixed this issue

Faraz32123 (approved):
LGTM!

ahmed-arb (commented):
I want to push back on the approach here before we merge, given how PR #1323 played out. That PR was reverted because it shifted everyone's SITE_ID from 2 to 1 and broke existing instances where no Site row with id=1 existed. We need to be careful that this fix doesn't introduce a similar class of footgun.

A few concerns with doing this in the Django settings file:

1. Monkey-patching django.setup() from a settings template is fragile. Settings are supposed to be declarative. Wrapping a core Django bootstrap function so that it does a DB write on every process start is a strong code smell. Any future Django change to the setup flow can break this silently, and it's a surprising thing for an operator to find when they cat lms settings and try to understand their config.
2. Mutating settings.SITE_ID post-load is racy with imports. Plenty of edx-platform and plugin code reads SITE_ID at module import time or caches derived values. Whatever gets imported and resolved before the patched django.setup finishes running keeps SITE_ID=2; everything after sees the new value. Ordering relative to AppConfig.ready() across plugins isn't guaranteed.
3. This still treats a data-state problem with a runtime workaround. Per @kdmccormick's analysis in the issue, the actual root cause is that our init creates two Site rows: Django's auto-created example.com at id=SITE_ID, and then a separate my.lms.org row at id=SITE_ID+1. Django's own docs recommend exactly the opposite: use a data migration to modify the existing example Site row in place (https://docs.djangoproject.com/en/5.2/ref/contrib/sites/#enabling-the-sites-framework) so that there's only ever one Site, and SITE_ID naturally points at it.

What I'd propose instead:

Fix the data, once, in the right layer. Tutor's post-migration init step (alongside createsuperuser, oauth client setup, etc.) for [lms](https://github.com/overhangio/tutor/blob/main/tutor/templates/jobs/init/lms.sh) and [cms](https://github.com/overhangio/tutor/blob/main/tutor/templates/jobs/init/cms.sh). A small idempotent step that:

- Ensures a Site row exists at id=SITE_ID (=2) with domain=LMS_BASE / CMS_BASE.
- If `example.com` is squatting on id=2, rename it in place to the real domain (Django's recommended pattern), or move/delete it.
- Logs what it did.

This way we:

- Fix both fresh installs and existing instances without an id=1-style breakage.
- Keep settings declarative.
- Run once per tutor launch instead of on every process boot.
- Make it observable in launch logs instead of behind a bare except.
- Avoids touching django.setup internals.

kdmccormick:
@ahmed-arb , I think that's a wise analysis and a good approach.

My only concern is that some instances may be in a situation where they have Site objects for both `example.com` _and_ for `<LMS_BASE>`. There might be data tied to each of these Site objects, and it would be hard for the Tutor init script to know which one is the "right" Site to keep. In that situation, I think it is best for the init script to:
* leave the database alone
* warn the user that both Sites exist and print their IDs
* recommend that the user deletes one of them and sets SITE_ID to the other one.

What do you think?

PR #1393, opened 2026-05-26
fix: volume bind-mounts on selinux/fedora

regisb:
I recently started running tutor on Fedora and realised that it was not working out of the box because of SELinux. That's because some bind-mounted volumes are shared between different containers. Without the "z" option, this sharing is disabled.

The "z" label is ignored in environments that do not use SELinux.

This feature is documented here:
https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label

### DIFF
--- a/changelog.d/20260526_124321_regis_fedora_selinux.md
+++ b/changelog.d/20260526_124321_regis_fedora_selinux.md
@@ -1,0 +1,1 @@
+- [Bugfix] Compatibility with Fedora/SELinux systems: volume bind-mounts must indicate that they are shared between different containers. Otherwise, SELinux fails to mount them. (by @regisb)
--- a/tutor/templates/dev/docker-compose.yml
+++ b/tutor/templates/dev/docker-compose.yml
@@ -5,7 +5,7 @@ x-openedx-service:
   tty: true
   volumes:
     # theme files
-    - ../build/openedx/themes:/openedx/themes
+    - ../build/openedx/themes:/openedx/themes:z
 
 services:
   permissions:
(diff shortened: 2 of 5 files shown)


ahmed-arb (approved):

muhammadadeeltajamul (approved):

Faraz32123 (approved):

Danyal-Faheem:
While this PR LGTM, I have concerns about this being an issue with every single plugin that will be used alongside tutor. Anyone running SELinux will have to hope that the maintainer resolves this upstream or that they manually override each volume mount.

Would it be possible to check that we are running SELinux (by running `uname` or otherwise) and add a `z` in front of all volume paths as well for all plugins installed in tutor?

Alternatively, we could just document that plugin developers should add a `z` for SELinux users and it will be ignored for the others.

ahmed-arb:
> Would it be possible to check that we are running SELinux (by running `uname` or otherwise) and add a `z` in front of all volume paths as well for all plugins installed in tutor?

We can check the OS, but post-processing the rendered Docker Compose YAML to append the `z` suffix to hard-coded volumes just sounds bad: we'd have to skip named volumes, avoid double-labeling, and handle the long-form volumes: syntax.

I think we should just document this and rely on maintainers to do their part in making their plugins SELinux-compatible.

Danyal-Faheem:
> We can check the OS, but post-processing the rendered Docker Compose YAML to append the z suffix to hard-coded volumes just sounds bad: we'd have to skip named volumes, avoid double-labeling, and handle the long-form volumes: syntax.

I agree but we should definitely update this in the `tutor-mfe` and `tutor-indigo` atleast (if it has docker templates) as well. Mfe is basically a necessary plugin to run openedx now.

ahmed-arb:
Let's merge this and open an issue to track this.

PR #1397, opened 2026-06-04
ci: add workflow to generate PRs for version bumps

Danyal-Faheem:
A common theme is creating a new PR for a version bump which takes precious time out of a maintainers day Since the actual workflow is pretty predictable, we just added a small github action to bump version and create a new PR instead

This workflow is only run manually and requires inputs for the updated version, updated OPENEDX_COMMON_VERSION (optional)

The commit message, branch name, PR title are all named the new version as per previously used convention

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,0 +1,81 @@
+name: Bump Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "New version (e.g. 21.0.8)"
+        required: true
+        type: string
+      openedx_common_version:
+        description: "OPENEDX_COMMON_VERSION (e.g. release/ulmo.4) Leave blank to use current version"
+        required: false
+        type: string
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set version variables
+        run: |
+          echo "VERSION=${{ github.event.inputs.version }}" >> "$GITHUB_ENV"
+          echo "RELEASE_TAG=v${{ github.event.inputs.version }}" >> "$GITHUB_ENV"
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
+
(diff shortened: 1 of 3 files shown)


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,84 @@
+name: Bump Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "New version (e.g. 21.0.8)"
```
ahmed-arb, line 10: In the comments or in the variable description, could we please link this [doc on versioning](https://github.com/overhangio/tutor/blob/6a86d2c625d5a2afc6f04452438add1b7dfd7f18/docs/tutor.rst#versioning) in Tutor.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Set version variables
+        run: |
+          echo "VERSION=${{ github.event.inputs.version }}" >> "$GITHUB_ENV"
+          echo "RELEASE_TAG=v${{ github.event.inputs.version }}" >> "$GITHUB_ENV"
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
```
ahmed-arb: In the next major release, we are deprecating Python 3.9. We should add a TODO here.

ahmed-arb (commented):
We should clarify in the action name/description/comment that this action should only be used for MAJOR and MINOR version bumps, *not* for RELEASE version. Better yet, we add a check that blocks release version bump in it.

Also, let's add a link to this action in the [versioning doc](https://github.com/overhangio/tutor/blob/6a86d2c625d5a2afc6f04452438add1b7dfd7f18/docs/tutor.rst#versioning).

.github/workflows/bump-version.yml
Danyal-Faheem: Added a few lines of comments.

.github/workflows/bump-version.yml
Danyal-Faheem: Since this python version does not deal with any code in tutor, but rather just installing and running scriv for collecting changelog entries, I've insted updated to the LTS release instead which is python 3.14.

ahmed-arb (approved):

Issue #1400, opened 2026-06-11
Upgrade outdated dependencies for the Verawood release


ahmed-arb:
Tracking issue for bringing Tutor's pinned dependencies up to date ahead of the Open edX Verawood release. Rather than landing everything at once, each dependency will be upgraded in its **own PR** and linked back here so they can be reviewed and tested independently.

Targets stay on the version line that Verawood actually tests/supports (latest stable patch within the current major), not absolute-latest. Django and React are pulled in upstream via `OPENEDX_COMMON_VERSION` (`release/verawood`) and are out of scope. Host/cluster tooling (Docker, Docker Compose, Kubernetes) is operator-managed and not pinned in this repo.

### Safe pin bumps (one-line, no migration logic)

- [x] **MySQL** `8.4.0` → `8.4.9` — latest 8.4 **LTS** patch (stay on 8.4; 9.x is a non-LTS Innovation series, not what Open edX tests) — `defaults.yml`
- [x] **MongoDB** `7.0.28` → `7.0.35` — latest 7.0 patch, pulls in upstream security patches (stay on 7.0 line) — `defaults.yml`
- [x] **Redis** `7.4.5` → `7.4.9` — latest 7.4 patch, pulls in upstream security patches (stay on 7.4 line) — `defaults.yml`
- [x] **Caddy** `2.7.4` → `2.11.4` — latest stable Caddy 2.x — `tutor/templates/config/defaults.yml`
- [x] **Python** `3.12.12` → `3.12.13` — latest 3.12 patch — `tutor/templates/build/openedx/Dockerfile`
- [x] **Node.js** `24.12.0` → `24.16.0` — latest Node 24 Active LTS — `Dockerfile`

### Meilisearch (needs re-index / migration validation)

- [x] **Meilisearch** `v1.8.4` → `v1.36.0` — matches the client (`meilisearch==0.41.0`) shipped by edx-platform on `release/verawood`. This jump changes the on-disk index format; indexes are rebuilt by the standard `init` jobs, so the PR must verify a re-init and exercise search (course discovery, courseware, Studio).

### Deferred — major-version upgrades requiring migration code (separate, dedicated PRs)

Intentionally **not** simple pin bumps; should not be bundled with the above:

[...]

PR #1401, opened 2026-06-11
Upgrade MySQL to v8.4.9

ahmed-arb:
Bumps the default MySQL image from `8.4.0` to `8.4.9`, the latest patch on the **8.4 LTS** line.

Open edX tests and ships against MySQL 8.4 LTS, so we stay on that line (rather than the non-LTS 9.x Innovation series) and move to the newest patch. No upgrade-command changes are needed for a patch bump — this PR is intentionally kept separate from the other dependency bumps in case 8.4-specific upgrade logic is needed later.

Part of #1400.

### Testing
- [x] `tutor images build openedx`
- [x] `tutor local launch` and confirm MySQL starts and migrations run

### DIFF
--- a/changelog.d/20260611_upgrade_mysql.md
+++ b/changelog.d/20260611_upgrade_mysql.md
@@ -1,0 +1,1 @@
+- [Improvement] Upgrade MySQL to v8.4.9, the latest patch on the 8.4 LTS line. (by @ahmed-arb)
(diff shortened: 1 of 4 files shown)


HammadYousaf01:
Verified this on both `dev` and `local`, no errors in either case.

Existing data survives the upgrade. MySQL does the in-place upgrade on first boot (`80400 → 80409` in the logs) and everything comes up fine. Fresh install also initializes cleanly on 8.4.9, migrations apply, LMS and Studio both healthy.

PR #1402, opened 2026-06-11
Upgrade MongoDB to v7.0.35

ahmed-arb:
Bumps the default MongoDB image from `7.0.28` to `7.0.35`, the latest patch on the **7.0** line, pulling in upstream security fixes.

We stay on 7.0 deliberately: Open edX Verawood's CI tests against `mongo-version: "7.0"` and pins `pymongo==4.4.0`, which predates MongoDB 8.0 server support. A future 7.0 → 8.0 major upgrade is tracked separately in #1400 because it requires `setFeatureCompatibilityVersion` stepping logic in `tutor/commands/upgrade/common.py` — kept this PR isolated so that logic can be added here if/when we go to 8.0.

Part of #1400.

### Testing
- [x] `tutor local launch` and confirm MongoDB starts cleanly (no featureCompatibilityVersion errors)

### DIFF
--- a/changelog.d/20260611_upgrade_mongodb.md
+++ b/changelog.d/20260611_upgrade_mongodb.md
@@ -1,0 +1,1 @@
+- [Security] Upgrade MongoDB to v7.0.35 to pull in upstream security patches, remaining on the 7.0 line that Open edX tests against. (by @ahmed-arb)
--- a/tutor/templates/config/defaults.yml
+++ b/tutor/templates/config/defaults.yml
@@ -19,7 +19,7 @@ DOCKER_IMAGE_CADDY: "docker.io/caddy:2.7.4"
 # https://hub.docker.com/r/getmeili/meilisearch/tags
 DOCKER_IMAGE_MEILISEARCH: "docker.io/getmeili/meilisearch:v1.8.4"
 # https://hub.docker.com/_/mongo/tags
-DOCKER_IMAGE_MONGODB: "docker.io/mongo:7.0.28"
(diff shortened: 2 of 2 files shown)


HammadYousaf01:
I've tested this locally. Clean upgrade, no breaking changes, rollback is safe.

- Launched against an existing 7.0.28 data volume (demo course imported beforehand). MongoDB started cleanly. WiredTiger opened the existing files with no recovery issues, featureCompatibilityVersion stayed at 7.0, no FCV errors, RestartCount=0.
- No errors in the mongod log (only the usual access-control / THP / vm.max_map_count startup warnings, same as 7.0.28). No pymongo/connection errors in LMS or CMS.
 
Mongo code paths exercised on 7.0.35
- Modulestore reads: MFE rendering, xblocks, and problem xblocks all work.
- Modulestore writes: Added a unit with mixed xblock types; new structures/definitions documents written as expected.
- GridFS read + write: uploaded a new asset and fetched it back through the LMS byte-exact.
- ensure_indexes: created contentstore/modulestore indexes with no errors.
- reindex_course --all: 406 blocks reindexed into Meilisearch, 0 failures.

Rollback (7.0.35 -> 7.0.28)
- Reverted the image tag and relaunched. 7.0.28 booted clean and read back everything written by 7.0.35 (modulestore docs, the uploaded asset byte-exact, and the indexes created under 7.0.35). Confirms the upgrade is reversible by just reverting DOCKER_IMAGE_MONGODB, no data surgery.

PR #1403, opened 2026-06-11
Upgrade Caddy, Redis, Python and Node

ahmed-arb:
Bundled patch/minor bumps for the dependencies that don't need any upgrade-command logic, each moving to the latest stable on its current line:

| Component | From | To | Notes |
|---|---|---|---|
| Caddy | 2.7.4 | 2.11.4 | latest stable Caddy 2.x |
| Redis | 7.4.5 | 7.4.9 | latest 7.4 patch (security fixes); stays on the OSS 7.x line |
| Python | 3.12.12 | 3.12.13 | latest 3.12 patch |
| Node.js | 24.12.0 | 24.16.0 | latest Node 24 Active LTS |

MySQL and MongoDB are handled in separate PRs (#1401, #1402), so their upgrade-command logic can evolve independently. Redis 8.x and MongoDB 8.0 remain deferred — see #1400.

Part of #1400.

### Testing
- [x] `tutor images build openedx` (Python + Node)
- [x] `tutor local launch`; verify Caddy proxying and Redis-backed cache/Celery.

### DIFF
--- a/changelog.d/20260611_upgrade_dependencies.md
+++ b/changelog.d/20260611_upgrade_dependencies.md
@@ -1,0 +1,2 @@
+- [Improvement] Upgrade Caddy to v2.11.4, Redis to v7.4.9, and Meilisearch to v1.36.0. The Meilisearch jump changes the on-disk index format; the standard `init` jobs rebuild the indexes, so re-run platform initialisation after upgrading. (by @ahmed-arb)
+- [Improvement] Upgrade Python to v3.12.13 and Node.js to v24.16.0 in the Open edX Docker image. (by @ahmed-arb)
--- a/changelog.d/20260611_upgrade_dependencies.md
+++ b/changelog.d/20260611_upgrade_dependencies.md
@@ -1,2 +1,2 @@
-- [Improvement] Upgrade Caddy to v2.11.4, Redis to v7.4.9, and Meilisearch to v1.36.0. The Meilisearch jump changes the on-disk index format; the standard `init` jobs rebuild the indexes, so re-run platform initialisation after upgrading. (by @ahmed-arb)
+- [Improvement] Upgrade Caddy to v2.11.4 and Redis to v7.4.9. (by @ahmed-arb)
 - [Improvement] Upgrade Python to v3.12.13 and Node.js to v24.16.0 in the Open edX Docker image. (by @ahmed-arb)
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -67,7 +67,7 @@ This configuration parameter defines the name of the Docker image to run the dev
 
 .. https://hub.docker.com/r/devture/exim-relay/tags
 
-- ``DOCKER_IMAGE_CADDY`` (default: ``"docker.io/caddy:2.6.2"``)
+- ``DOCKER_IMAGE_CADDY`` (default: ``"docker.io/caddy:2.11.4"``)
 
 This configuration parameter defines which Caddy Docker image to use.
 
@@ -87,7 +87,7 @@ This configuration parameter defines which MySQL Docker image to use.
 
 .. https://hub.docker.com/_/redis/tags
 
-- ``DOCKER_IMAGE_REDIS`` (default: ``"docker.io/redis:7.4.5"``)
+- ``DOCKER_IMAGE_REDIS`` (default: ``"docker.io/redis:7.4.9"``)
 
 This configuration parameter defines which Redis Docker image to use.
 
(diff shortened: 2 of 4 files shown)


HammadYousaf01:
Tested this locally against the existing `tutor local` deployment (demo course).

Redis came up cleanly on 7.4.9, loaded the existing 7.4.5 RDB/AOF without any issues, and stayed stable throughout testing. Since we're using standalone Redis, the `--cluster-announce-ip` change in 7.4.9 doesn't affect this deployment.

I exercised the Redis-backed paths that Openedx depends on:

* Cache read/write across all `django-redis` aliases (`default`, `general`, `course_structure_cache`, `configuration`).
* Course structure cache invalidation by renaming and publishing a section in Studio, then verifying the updated title appeared in the LMS outline.
* Browser login/session creation, including confirming the session remained valid after restarting Redis.
* Celery reconnect after a Redis restart, with both grade report and bulk email tasks completing successfully.
* `django-ratelimit` counters and enforcement.
* Memory eviction by reducing `maxmemory` to 30 MB and confirming `allkeys-lru` evicted cache entries without any `OOM command not allowed` errors while Celery continued processing tasks.
* Redis Streams used by `edx-event-bus-redis`, including producing, consuming, and acknowledging an `xblock.deleted.v1` event.
* Persistence by successfully running `BGREWRITEAOF` and verifying data survived the upgrade.

I also tested rollback by reverting the image back to 7.4.5 and recreating the Redis container. It started normally, loaded the data written by 7.4.9, and upgrading back to 7.4.9 worked as well. Rollback is just reverting `DOCKER_IMAGE_REDIS`; no data migration or manual cleanup was needed.

I'm confident about the upgrade.

HammadYousaf01:
Tested this locally against the existing `tutor local` deployment in HTTP mode.

Caddy recreated cleanly on 2.11.4 and stayed up. The Caddyfile validated successfully, and a `caddy adapt` diff against 2.7.4 only showed cosmetic JSON differences. I didn't see any changes to routing, headers, or TLS behavior. The logs were clean apart from the existing `{args.0}` deprecation warning that's already present on 2.7.4, and I didn't hit any proxy errors in LMS or CMS.

I exercised the main request paths:

* LMS, CMS/Studio, the MFE host, and Meilisearch all served normally with HTTP 200 responses.
* `/login` still routes through the authn MFE with `csrftoken` and `sessionid` set correctly, and CSRF plus `X-Forwarded-*` headers remained intact.
* `encode gzip` is working, requests larger than 4 MB are rejected with a 413 at the edge, and smaller POST requests pass through normally.
* Verified the new underscore header handling in 2.11.4: headers such as `X_Test` are dropped, while `X-Test` is preserved. This doesn't affect Open edX since it uses hyphenated headers.
* For TLS, I temporarily enabled `local_certs` since `local.openedx.io` can't obtain a public certificate. Internal CA certificates were issued successfully, TLS 1.3 negotiated correctly, HTTP/3 was advertised, and HTTP→HTTPS redirects worked as expected.

I also tested rollback by reverting the image back to 2.7.4 and recreating the container. Caddy started cleanly, all routes continued to work, and upgrading back to 2.11.4 was uneventful. Since Caddy has no persistent data, rollback is simply reverting `DOCKER_IMAGE_CADDY`. No migration or cleanup is required.

HammadYousaf01:
I don't think the Meilisearch upgrade should be merged as part of this PR.

The issue isn't with the image update itself. It's that existing Meilisearch indexes from older versions are incompatible with 1.36. An upgraded instance won't start until the old `data.ms` is removed, `init` recreates the API key and index metadata, and the indexes are rebuilt. That's a migration path that deserves its own discussion and testing.

I'd suggest splitting the Meilisearch upgrade into a separate PR so we can document the required upgrade steps and review it independently of the other dependency bumps.

HammadYousaf01:
Tested this locally for the Python and Node version bumps against both `tutor local` and `tutor dev`.

As submitted, `tutor images build openedx` fails because the Python version was bumped to **3.12.13**, but the pinned `pyenv` version was not. `pyenv` **v2.6.18** doesn't include a build definition for Python 3.12.13, so the build fails at `pyenv install`. I bumped the `pyenv` pin to **v2.7.2** as well, after which the build completed successfully.

With that change in place, both the `openedx` and `openedx-dev` images build cleanly. I verified:

* Python 3.12.13 and Node 24.16.0 are present in the running `local` and `dev` containers.
* Asset builds (`webpack`, `build-dev`) and `collectstatic` complete successfully.
* LMS, Studio, and Celery all start normally under both `tutor local` and `tutor dev`.

The Node bump doesn't require any additional changes.

HammadYousaf01:
I've dropped Meilisearch from this PR and smoke-tested the platform. Caddy, Redis, Python, and Node all come up cleanly and stay healthy. The Meilisearch upgrade needs its own index migration, so I'll handle it in a separate PR and merge these for now.

Issue #1409, opened 2026-06-24
Secured cookies for LMS and CMS


jfavellar90:
**Is your feature request related to a problem? Please describe.**

Secured cookies (session and CSRF) are enforced for the LMS when enabling HTTPS, via these configuration [lines](https://github.com/overhangio/tutor/blob/verawood/tutor/templates/apps/openedx/settings/lms/production.py#L13-L24). However, CMS does not get secured cookies. This is a security requirement asked by organizations with strong security policies.

**Describe the solution you'd like**
I would like to move this configuration [block](https://github.com/overhangio/tutor/blob/verawood/tutor/templates/apps/openedx/settings/lms/production.py#L13-L24)  to the [common_all file](https://github.com/overhangio/tutor/blob/verawood/tutor/templates/apps/openedx/settings/partials/common_all.py) so the CMS gets the benefits of the secured cookies configuration.

**Describe alternatives you've considered**
For now, we are handling this via Tutor plugins. However, I consider it to be the type of change that could be offloaded into the Tutor core.

jfavellar90:
I created this https://github.com/overhangio/tutor/pull/1410, implementing the solution

muhammadadeeltajamul:
PR is merged... Moving this ticket to done

PR #1410, opened 2026-06-24
chore: CMS cookie secure settings

jfavellar90:
This PR moves the cookie secure configuration from the LMS production files to the common configuration file. This change makes the CMS service getting the secure cookies configuration

It solves https://github.com/overhangio/tutor/issues/1409

### DIFF
--- a/changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
+++ b/changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
@@ -1,0 +1,9 @@
+<!--
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+- [Chore] Move secure cookie configuration to the common production settings so it applies to both LMS and CMS. (by @jfavellar90)
--- a/changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
+++ b/changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
@@ -1,0 +1,1 @@
+- [Security] Move secure cookie configuration to the common production settings so it applies to both LMS and CMS. (by @jfavellar90)
--- a/tutor/templates/apps/openedx/settings/lms/production.py
+++ b/tutor/templates/apps/openedx/settings/lms/production.py
@@ -10,19 +10,6 @@ ALLOWED_HOSTS = [
 ]
 CORS_ORIGIN_WHITELIST.append("{% if ENABLE_HTTPS %}https{% else %}http{% endif %}://{{ LMS_HOST }}")
 
-{% if ENABLE_HTTPS %}
-# Properly set the "secure" attribute on session/csrf cookies. This is required in
-# Chrome to support samesite=none cookies.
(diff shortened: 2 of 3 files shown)


ahmed-arb:
Thanks @jfavellar90 for the PR. Could you please target main and rebase your branch?

jfavellar90:
@ahmed-arb done, just need to generate a changelog file

jfavellar90:
@ahmed-arb changelog is ready

changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
```diff
@@ -0,0 +1,9 @@
+<!--
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
+- You may optionally append "(by @<author>)" at the end of the line, where "<author>" is either one (just one)
+  of your GitHub username, real name or affiliated organization. These affiliations will be displayed in
+  the release notes for every release.
+-->
+- [Chore] Move secure cookie configuration to the common production settings so it applies to both LMS and CMS. (by @jfavellar90)
```
ahmed-arb: nit: The changelog entry is tagged [Chore], but this is a user-facing security improvement, so [Security] or [Improvement] reads better.

ahmed-arb (commented):
I have one concern with common_all.py as the destination. It is included by the development settings as well as production.

Could we instead just duplicate the block into cms/production.py and leave lms/production.py as it is? It keeps the setting production-scoped, which is where it belongs, and duplication across the two production files is already the pattern for `ALLOWED_HOSTS` and `CORS_ORIGIN_WHITELIST.append(...)`.                                                                                                                

Once that has landed, I would like to follow up with the nicer version: add a partials/common_production.py included from both lms/production.py and cms/production.py, and move all the settings that are common to L it, the cookie block included.

ahmed-arb:
On timing: we are cutting v22.0.0 for Verawood imminently, so I would rather not pull this into the verawood branch as-is. Once Verawood is out, please retarget the reworked version at release instead of main.

changelog.d/20260724_015046_donato.bracuto_cms_secure_cookies.md
jfavellar90: Addresed

jfavellar90:
@ahmed-arb I proceeded as you suggested since Verawood had already been cut. Regarding your suggestion to duplicate the block in the `cms/production.py` file instead of moving it to `common_all.py`, I have the following comments:

- `ENABLE_HTTPS` is already being used in conditionals for variable definitions in `common_all.py`.
- The reason I think `ALLOWED_HOSTS` or `CORS_ORIGIN_WHITELIST` were defined independently in every LMS/CMS production.py file is because such variables depend on LMS/CMS service-specific variables (like `LMS_HOST` or `LMS_BASE`). This is not the case for the session cookie settings.
- The landing of these variables in the development settings only depends on the `ENABLE_HTTPS` value. If the variable is `false`, the configured settings have the exact values Django uses by default (which makes me think we could get rid of the `else` block):

SESSION_COOKIE_SECURE = False
CSRF_COOKIE_SECURE = False
SESSION_COOKIE_SAMESITE = 'Lax'

I think we should consider keeping the original approach; what do you think?

ahmed-arb:
Fair point. Approving it then.

ahmed-arb (approved):

jfavellar90:
@ahmed-arb thanks for the review. Would you need me to add a tutor version bump (it would be 22.0.1) and collect the changelog?

ahmed-arb:
Sure, but let's bundle a few more changes in 22.0.1. We'll try to close a few other PRs this week.

Issue #1347, opened 2026-02-09
OpenEdx image build fails with setuptools >= 81.0.0 (pkg_resources removed)


cmersinli:
## Description

OpenEdx Docker image build fails during `uv pip install -r base.txt` because `loremipsum==1.0.5` (a transitive dependency via `ora2`) cannot be built.

**Root cause:** `setuptools 81.0.0` (released Feb 6, 2026) [removed the `pkg_resources` module](https://discuss.python.org/t/will-setuptools-remove-pkg-resources-module-in-the-future/27182). When `uv` builds `loremipsum` in an isolated environment, it installs the latest `setuptools` (now >= 81), which no longer includes `pkg_resources`. The `loremipsum` package imports `pkg_resources` at build time in its `setup.py` but doesn't declare `setuptools` as a build dependency.

## Error

```
[python-requirements 3/9] RUN uv pip install -r base.txt -r assets.txt

× Failed to build `loremipsum==1.0.5`
├─▶ The build backend returned an error
╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed

  File ".../loremipsum/generator.py", line 8, in <module>
    from pkg_resources import resource_string
  ModuleNotFoundError: No module named 'pkg_resources'
```

## Environment

- **Tutor:** 21.0.0 (Ulmo)
- **uv:** 0.9.18 (from Dockerfile template)
- **edx-platform:** release/ulmo.1
- **setuptools in build isolation:** >= 81.0.0 (latest from PyPI)

## Impact

This affects **all** Tutor 21.x OpenEdx image builds starting Feb 6, 2026 when `setuptools 81.0.0` was released. Any build without a Docker layer cache for the base requirements step will fail.

## Workaround

Disable uv build isolation for `loremipsum` using a Tutor plugin with `openedx-dockerfile-minimal` patch:

```python
from tutor import hooks

hooks.Filters.ENV_PATCHES.add_item((
    "openedx-dockerfile-minimal",
    """
# Fix: loremipsum 1.0.5 needs setuptools (pkg_resources) at build time
# but doesn't declare it. Disable uv build isolation for this package.
RUN printf '[pip]\\nno-build-isolation-package = ["loremipsum"]\\n' > /etc/uv.toml
ENV UV_CONFIG_FILE=/etc/uv.toml
"""
))
```

## Suggested fix

Consider one of:
1. Add `no-build-isolation-package = ["loremipsum"]` to the Dockerfile template's uv config
[...]

PR #1411, opened 2026-06-29
fix: pin build-time setuptools to stabilize source builds

### LINKED ISSUE #1350: `uv` ignores the setuptools pin, so the openedx image build is unstable
The Dockerfile pins setuptools to v62 so that breaking changes to setuptools don't affect our build: https://github.com/overhangio/tutor/blob/release/tutor/templates/build/openedx/Dockerfile#L107

but `uv` seems to ignore that pin: https://github.com/astral-sh/uv/issues/2500

so it seems we actually build with setuptools v82.

The release of v82 recently broke both Ulmo and Main build: https://github.com/overhangio/tutor/issues/1347 . We have resolved that, but we are still vulnerable to future breaking changes in setuptools. If we do not address this in Ulmo, then even as we move on to Verawood, the Ulmo image build will be at risk of breakage at the whim of setuptools.

Abdul-Muqadim-Arbisoft:
## Problem

The Open edX image build is unstable because pinning `setuptools` in the
virtualenv does **not** control the `setuptools` used to *build* packages.

When `uv` (or `pip`) builds any package from source (an sdist), it creates an
isolated PEP 517 build environment and installs the **latest** `setuptools`
into it — completely ignoring the `setuptools==69.1.1` pinned into
`/openedx/venv`. So the runtime pin has no effect on builds.

This is what broke Ulmo and Main builds in #1347: `setuptools` 81 removed the
`pkg_resources` module, and `loremipsum` (a transitive dependency of
`edx-ora2`) imports `pkg_resources` in its build script without declaring
`setuptools` as a build dependency → `ModuleNotFoundError: No module named
'pkg_resources'`. The build remains exposed to any future `setuptools`
breaking change.

Fixes #1350.

## Fix

Pin the build-time dependencies used inside the isolated build environments by
setting, in the base `minimal` stage:

- `UV_BUILD_CONSTRAINT` — read by `uv pip` (build environments only)
- `PIP_CONSTRAINT` — read by plain `pip`

Both point at a one-line constraints file (`setuptools==69.1.1`). Because the
ENV vars are set in the base stage, they are inherited by every downstream
stage, so the constraint applies to **all** `$PIP_COMMAND install` calls:
`base.txt`, `assets.txt`, `development.txt`, extra requirements, scorm, uWSGI
and the editable installs (`-e .`, `-e /mnt/...`).


tutor/templates/build/openedx/Dockerfile
```diff
@@ -14,6 +14,24 @@ ENV LC_ALL=en_US.UTF-8
...
+# which some build scripts still import (e.g. loremipsum, a transitive
+# dependency of edx-ora2). The constraint is inherited by every downstream
+# stage, so it applies to all "$PIP_COMMAND install" calls below (base.txt,
+# assets.txt, development.txt, extra requirements, editable installs, ...).
+# UV_BUILD_CONSTRAINT is read by "uv pip" (build environments only); plain
+# "pip" reads PIP_CONSTRAINT. We only constrain setuptools: wheel is pinned by
+# edx-platform's own requirements, and pip is not a build dependency.
+# Keep this version in sync with the setuptools installed in the venv below.
+# https://github.com/overhangio/tutor/issues/1350
+RUN echo "setuptools==69.1.1" > /etc/openedx-build-constraints.txt
```
ahmed-arb, line 23: Can we think of a way so that we don't have to update setuptools' version in two places?

tutor/templates/build/openedx/Dockerfile
```diff
@@ -14,6 +14,24 @@ ENV LC_ALL=en_US.UTF-8
...
+# non-reproducible and broke them when setuptools 81 removed pkg_resources,
+# which some build scripts still import (e.g. loremipsum, a transitive
+# dependency of edx-ora2). The constraint is inherited by every downstream
+# stage, so it applies to all "$PIP_COMMAND install" calls below (base.txt,
+# assets.txt, development.txt, extra requirements, editable installs, ...).
+# UV_BUILD_CONSTRAINT is read by "uv pip" (build environments only); plain
+# "pip" reads PIP_CONSTRAINT. We only constrain setuptools: wheel is pinned by
+# edx-platform's own requirements, and pip is not a build dependency.
+# Keep this version in sync with the setuptools installed in the venv below.
+# https://github.com/overhangio/tutor/issues/1350
```
ahmed-arb: nit: Please make this comment a bit concise. The PR description already has this information. If someone wants to know the thought behind this change, they can always check git blame.

ahmed-arb (approved):

ahmed-arb:
Let's update the setuptool version for the next release and pin the current version in Ulmo. This way we won't have to worry about Ulmo randomly breaking after the Verawood release.

ahmed-arb (approved):

Issue #1242, opened 2025-06-18
ERROR 4006 (HY000) at line 1: Operation CREATE USER failed


0x29a:
[This command](https://github.com/overhangio/tutor/blob/67e6f0c637d40d0ed8ec2f429a5eefde40f27122/tutor/templates/jobs/init/mysql.sh#L19) from the MySQL initialization script has been failing since commit https://github.com/openedx/edx-platform/commit/6c2c4333bbe3b66db5694a45689bf8bf222a7c59:
```
ERROR 4006 (HY000) at line 1: Operation CREATE USER failed for 'pr_1758_49e51b_openedx'@'%' as it is referenced as a definer account in a view.
```
[This function](https://github.com/mysql/mysql-server/blob/19feac3674e2ae254b3e7fa7e116fc64ac284d8d/sql/auth/sql_user.cc#L2563-L2612) in MySQL raises the error.

The failure occurs because the commit upgraded `edx-enterprise` to version `5.13.1`, which introduces [this migration](https://github.com/openedx/edx-enterprise/blob/3702e3c34febf5060fb0b095f7999ab1b6541e3b/enterprise/migrations/0236_create_vw_enterprise_customer_ecus_pecus.py#L13-L49). That migration creates the `view_enterprise_customer_support_users` view.

`CREATE USER` does not fail on a local MySQL instance, because local users possess the `SET_USER_ID` permission. In DigitalOcean's managed MySQL (that we use), this dynamic privilege is intentionally disallowed by DigitalOcean and cannot be enabled, so the command fails there.

This error has not yet appeared on AWS RDS, but is very likely to surface soon. As of MySQL >= 8.4.3, the privilege [was renamed to](https://github.com/mysql/mysql-server/commit/3d0cfd5c06f13d6d4cd327b2506877ea2359a96c) `ALLOW_NONEXISTENT_DEFINER` and is explicitly disallowed, as shown in [this table](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.MySQL.CommonDBATasks.dynamic-privileges.html).

[...]

PR #1412, opened 2026-06-30
fix: mysql user create if not exists error

Kelketek:
## Description

This pull request fixes an issue where MySQL may choke on the 'create user if not exists' command in the case that a user has a view that they have created. This is a bug in MySQL that has existed for at least a few years: https://bugs.mysql.com/bug.php?id=107139

The bug is more likely to occur in some managed database environments where the provided 'admin user' does not have the "SET_USER_ID" privilege (in version 8.0) or the "ALLOW_NONEXISTENT_DEFINER" privilege (in version 8.4). This is the case, for instance, with DigitalOcean. It may also happen for reasons not yet fully understood even if this privilege is available.

## Supporting information

Internal ticket: https://tasks.opencraft.com/browse/BB-10998

Edx-Enterprise migration which creates a view: https://github.com/openedx/edx-enterprise/blob/master/enterprise/migrations/0236_create_vw_enterprise_customer_ecus_pecus.py

## Testing instructions

1. Run launch and verify nothing breaks in practice for normal migrations/initialization
2. Create an admin user with all privileges except for SET_USER_ID (if using MySQL 8.0) or ALLOW_NONEXISTENT_DEFINER (if using MySQL 8.4), and set its settings:
* MYSQL_ROOT_USERNAME
* MYSQL_ROOT_PASSWORD
* MYSQL_HOST
* MYSQL_PORT
... as needed.
* Try launching once more. Verify launch completes.

### DIFF
--- a/tutor/templates/jobs/init/mysql.sh
+++ b/tutor/templates/jobs/init/mysql.sh
@@ -16,6 +16,15 @@ echo "MySQL is up and running"
 
 # edx-platform database
 mysql -u {{ MYSQL_ROOT_USERNAME }} --password="{{ MYSQL_ROOT_PASSWORD }}" --host "{{ MYSQL_HOST }}" --port {{ MYSQL_PORT }} -e "CREATE DATABASE IF NOT EXISTS {{ OPENEDX_MYSQL_DATABASE }};"
-mysql -u {{ MYSQL_ROOT_USERNAME }} --password="{{ MYSQL_ROOT_PASSWORD }}" --host "{{ MYSQL_HOST }}" --port {{ MYSQL_PORT }} -e "CREATE USER IF NOT EXISTS '{{ OPENEDX_MYSQL_USERNAME }}';"
+# This command may fail if the specified user has created a custom view, such
+# as is done in the enterprise app.
+#
+# Yes, even 'if not exists' does not prevent this. :/
+# https://bugs.mysql.com/bug.php?id=107139
+#
+# When our minimum version of MySQL is 8.4, we should be able to drop this
+# || true workaround, as the SET_USER_ID privilege has been removed.
+mysql -u {{ MYSQL_ROOT_USERNAME }} --password="{{ MYSQL_ROOT_PASSWORD }}" --host "{{ MYSQL_HOST }}" --port {{ MYSQL_PORT }} -e "CREATE USER IF NOT EXISTS '{{ OPENEDX_MYSQL_USERNAME }}';" || true
+# If the user truly doesn't exist, we'll get an error here.
 mysql -u {{ MYSQL_ROOT_USERNAME }} --password="{{ MYSQL_ROOT_PASSWORD }}" --host "{{ MYSQL_HOST }}" --port {{ MYSQL_PORT }} -e "ALTER USER '{{ OPENEDX_MYSQL_USERNAME }}'@'%' IDENTIFIED BY '{{ OPENEDX_MYSQL_PASSWORD }}';"
 mysql -u {{ MYSQL_ROOT_USERNAME }} --password="{{ MYSQL_ROOT_PASSWORD }}" --host "{{ MYSQL_HOST }}" --port {{ MYSQL_PORT }} -e "GRANT ALL ON {{ OPENEDX_MYSQL_DATABASE }}.* TO '{{ OPENEDX_MYSQL_USERNAME }}'@'%';"


tecoholic:
@Kelketek This was kind of insane that I didn't believe no one has run into this before, so I took a look.

1. The bug related to SET_USER_ID only affects MySQL Version 8.0 and not 8.4.
  > The SET_USER_ID privilege, deprecated in MySQL 8.2.0, has been removed in this release, and its use in [GRANT](https://dev.mysql.com/doc/refman/8.4/en/grant.html) statements now causes a syntax error. Use the [SET_ANY_DEFINER](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#priv_set-any-definer) and [ALLOW_NONEXISTENT_DEFINER](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#priv_allow-nonexistent-definer) privileges instead. (WL #15875)
2. That means, theoretically we need to set `DOCKER_IMAGE_MYSQL` to `docker.io/mysql:8.0` in our Tutor environment and run `tutor dev launch` twice to observe the failure.

If anyone like me doesn't want to spend that time confirming this using Tutor, this file contains the steps to reproduce the same issue in a standalone docker container - 
[mysql-shop-walkthrough.md](https://github.com/user-attachments/files/29582997/mysql-shop-walkthrough.md)

Running the same experiment using a MySQL 8.4 container returns cleanly.

tecoholic (approved):
@Kelketek Can you kindly update the comments to mention that this is needed for MySQL 8.0 only.

Kelketek:
@tecoholic Done. Thank you for digging further into that-- it makes sense why others may not have hit it yet.

muhammadadeeltajamul:
I think this change is not needed as current MySQL version on release branch when setting up Tutor is 8.4.11 ([See here](https://github.com/overhangio/tutor/blob/release/tutor/templates/config/defaults.yml#L24))

tecoholic:
@muhammadadeeltajamul This bugfix was for Tutor instances that used MySQL 8.0 from other sources like managed cloud DB service, and not for the one bundled by default with Tutor. As I mentioned [in my earlier comment](https://github.com/overhangio/tutor/pull/1412#issuecomment-4862432827) it's not commonly occurring and requires specific setup to reproduce locally.

DigitalOcean for example made MySQL 8.4 available in their managed DB offering only on May 2026. So the instances deployed earlier all had to use Mysql 8.0 and would run into issues.

Kelketek:
@muhammadadeeltajamul Just a ping here. Can you take another look in light of @tecoholic 's comment?

Abdul-Muqadim-Arbisoft:
> @Kelketek This was kind of insane that I didn't believe no one has run into this before, so I took a look.
> 
> 1. The bug related to SET_USER_ID only affects MySQL Version 8.0 and not 8.4.
> 
> > The SET_USER_ID privilege, deprecated in MySQL 8.2.0, has been removed in this release, and its use in [GRANT](https://dev.mysql.com/doc/refman/8.4/en/grant.html) statements now causes a syntax error. Use the [SET_ANY_DEFINER](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#priv_set-any-definer) and [ALLOW_NONEXISTENT_DEFINER](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#priv_allow-nonexistent-definer) privileges instead. (WL #15875)
> 
> 2. That means, theoretically we need to set `DOCKER_IMAGE_MYSQL` to `docker.io/mysql:8.0` in our Tutor environment and run `tutor dev launch` twice to observe the failure.
> 
> If anyone like me doesn't want to spend that time confirming this using Tutor, this file contains the steps to reproduce the same issue in a standalone docker container - [mysql-shop-walkthrough.md](https://github.com/user-attachments/files/29582997/mysql-shop-walkthrough.md)
> 
> Running the same experiment using a MySQL 8.4 container returns cleanly.

Hi @tecoholic , thanks for the walkthrough, it was really helpful. I tried it on MySQL 8.4.11 (the version Tutor pins right now) and I think the "returns cleanly on 8.4" result is misleading.

The reason is step 6. REVOKE SET_USER_ID on 8.4 does nothing, because that privilege no longer exists there. MySQL accepts it silently, so root keeps SET_ANY_DEFINER and ALLOW_NONEXISTENT_DEFINER, and step 7 passes only because of that.

If I revoke the 8.4 equivalent instead, I get the same failure:


```
REVOKE ALLOW_NONEXISTENT_DEFINER ON *.* FROM 'root'@'%';
CREATE USER IF NOT EXISTS 'shopkeeper'@'%' IDENTIFIED BY 'shopkeeper';
-- ERROR 4006 (HY000): Operation CREATE USER failed for 'shopkeeper'@'%' as it is referenced as a definer account in a view.
```
I also ran the unchanged mysql.sh from the release branch twice against 8.4.11 with an admin lacking ALLOW_NONEXISTENT_DEFINER, and the second run fails with the same error.

So the bug is still present on 8.4, only the privilege name changed (SET_USER_ID on 8.0, ALLOW_NONEXISTENT_DEFINER on 8.4). The upstream bug 107139 is also still open and marked "8.0 and higher".

tecoholic:
@Abdul-Muqadim-Arbisoft Ah! I see. Thank you for looking deeply into this. I see another issue - https://github.com/overhangio/tutor/issues/1242 - reporting that this will cause a potential issue down the line. So, that confirms that we need to have the `|| true` in the end to keep the script working. We should probably update the inline comment as it won't be resolved just by using MySQL 8.4.

Kelketek:
@tecoholic @Abdul-Muqadim-Arbisoft I've removed the comment about removing it when moving to 8.4. Anything else needed or is this good to go now?

Abdul-Muqadim-Arbisoft:
@Kelketek  thank you for the contribution,  The change looks good to me and I confirmed it fixes the failure on both MySQL 8.0 and 8.4. A few things are needed before it can merge, per the Tutor contributing guidelines:

1. Add a changelog entry with `make changelog-entry`. Every user-facing fix needs one so it lands in the release notes. Suggested text:

[Bugfix] Make the MySQL init job tolerate CREATE USER IF NOT EXISTS failing with "referenced as a definer account in a view" (MySQL bug 107139). This happens on managed databases where the admin user lacks SET_USER_ID (MySQL 8.0) or ALLOW_NONEXISTENT_DEFINER (MySQL 8.4) once edx-enterprise has created its view. (by @Kelketek)

2. Add `Close #1242` to the commit message. It is the same bug, and the guidelines ask for the issue reference so it gets closed on merge.

3. Update the PR description and testing steps. They still say to revoke `SET_USER_ID`, which does not exist on 8.4, the version Tutor pins, so a reviewer following them will not reproduce the bug. Mention `ALLOW_NONEXISTENT_DEFINER` for 8.4.

Once those are in, this should be good to go.

Kelketek:
@Abdul-Muqadim-Arbisoft Done. Thanks!

Kelketek:
Realized I rebased on the wrong branch. Fixed.

Abdul-Muqadim-Arbisoft (approved):
LGTM! thank you for the contribution

Kelketek:
Thanks, @Abdul-Muqadim-Arbisoft !

Do I need to create any other follow-up PRs for reconciliation, or will this be both in the next point release and all moving forward now?

Issue #1182, opened 2024-12-27
SITE_ID misconfiguration


dyudyunov:
## Bug description
**Bottom line up:** 
Tutor installations have the SITE_ID setting pointing to the Django default `example.com` site while [trying to avoid it](https://github.com/overhangio/tutor/blob/8a09339164bc8d7a27de05098f5a36f00de525c8/tutor/templates/apps/openedx/settings/partials/common_all.py#L90).

While Tutor [explicitly says](https://github.com/overhangio/tutor/blob/8a09339164bc8d7a27de05098f5a36f00de525c8/tutor/templates/apps/openedx/settings/partials/common_all.py#L90) that it uses SITE_ID = 2 because example.com already uses id 1, it doesn't work that way.

Actual state on any Tutor-based openedx instance I checked:
1. We don't have a record for id == 1 at all
2. id 2 is used by example.com
3. The target site (the site I'm expecting on id 2) uses id 3

After a shallow investigation I found:
- The problem is not reproducible for ecommerce (and I suspect for credentials/discovery as well).
- Most probably that's because when Django creates the example.com record - [it uses SITE_ID from settings](https://github.com/django/django/blob/fcd9d08379a2aee3b2c49eab0d0b8db6fd66d091/django/contrib/sites/management.py#L36) and sets it explicitly.
- Edx-platform has the [SITE_ID setting set to 2](https://github.com/overhangio/tutor/blob/8a09339164bc8d7a27de05098f5a36f00de525c8/tutor/templates/apps/openedx/settings/partials/common_all.py#L90) while [e-commerce](https://github.com/overhangio/ecommerce/blob/c40aa6e5fb6bb2a83a53fa4de5f1d390007c78d4/ecommerce/settings/base.py#L82) and [discovery](https://github.com/overhangio/course-discovery/blob/3c477a113a78e0f0b24d7a9485d112a6e8207ed1/course_discovery/settings/base.py#L565) use the default value 1

## Consequences

For cases when the platform can't define the site (e.g., from the request), it falls back to the site with id = SITE_ID with the appropriate consequences, like bulk emails being sent without the site branding.

## How to check/reproduce

Go to the LMS admin site -> Sites and check IDs for `example.com` and your target domain (check the record URL for ID).

You can also try to navigate to the site object with ID = 1 (`<LMS>/admin/sites/site/1/`):
![image](https://github.com/user-attachments/assets/97212973-b496-4a05-85ac-49f2f19f616e)

## Environment
[...]

PR #1431, opened 2026-07-17
fix: point SITE_ID at the LMS domain during init (supersedes #1389)

ahmed-arb:
Fixes https://github.com/overhangio/tutor/issues/1182
Supersedes #1389 (and replaces the reverted #1323).

### Problem

Django auto-creates an `example.com` `Site` at `id=SITE_ID` during the first migration. Code paths that run **outside of an HTTP request** (e.g. bulk emails sent from Celery workers) have no request host to resolve the site from, so the lookup falls back to the `SITE_ID` setting and renders links and branding for `example.com` instead of the real domain.

Worth spelling out, because it shapes the fix: Open edX patches `Site.objects.get_current()` (the `django-sites-extensions` package) so that a request is matched against `domain__iexact=request.get_host()` **first**, and `SITE_ID` is consulted only when there is no request or no site matches the host. `SITE_ID` therefore only governs request-less code paths, which is exactly where this bug lives.

### Approach

This takes the init-layer, data-fix approach discussed on #1389 (rather than the reverted `SITE_ID=1` change, or a runtime `django.setup` monkeypatch):

- **`jobs/init/lms.sh`** (after `migrate`): renames the auto-created `example.com` site **in place** to the LMS domain (`settings.LMS_BASE`). This is [Django's recommended pattern](https://docs.djangoproject.com/en/5.2/ref/contrib/sites/#enabling-the-sites-framework): there is then a single `Site`, and `SITE_ID` (still a plain `= 2`) naturally points at it. Renaming in place keeps everything attached to that site (`SiteConfiguration`, themes, ...) valid. Idempotent.
- **Ambiguous installations** where both an `example.com` site *and* a separate site for the domain already exist are **left untouched**, with a warning printing both site ids and how to consolidate. This addresses @kdmccormick's concern (#1389 (comment)) that data may be tied to either site.
- **No CMS-side change.** The LMS and the CMS share one `django_site` table and one `SITE_ID`, and Studio resolves its site from the request host, so there is nothing for Tutor core to create here. On a default installation Indigo's init already creates a Studio-domain `Site` through `do settheme`, and Studio resolves to that one. Where no such row exists, Studio falls back to the `SITE_ID` site, which is better than a bare row with no `SiteConfiguration` attached. See the review discussion below.

`SITE_ID` remains declarative (`= 2`); no runtime mutation, no `django.setup` patching, no per-process DB writes.

### Why not #1389's approach

#1389 wraps `django.setup()` from the settings template to rewrite `settings.SITE_ID` on every process boot. As raised in review, that is fragile (patches a core Django bootstrap hook, DB write on every start), racy with import-time reads of `SITE_ID`, and treats a data-state problem as a runtime workaround. Fixing the data once, in the init step, avoids all of that.

### Testing

Verified with these commits cherry-picked onto `main` (they apply cleanly), with `overhangio/openedx:22.0.1`, under `tutor local` in an isolated `TUTOR_ROOT`:

- **Fresh install**: `Renamed default site id=2 from example.com to local.openedx.io`, leaving a single `Site`. A second `do init` prints `SITE_ID 2 already points at local.openedx.io` and writes nothing.
- **The bug itself, before and after**, on the same installation, through the path `bulk_email` uses (`emulate_http_request` + `theming.helpers.get_current_site`):
  ```
  CELERY-TASK sees site: 2 example.com        <- before
  CELERY-TASK sees site: 2 local.openedx.io   <- after
  ```
- **Upgrade** from a pre-fix installation carrying a `SiteConfiguration` and a `SiteTheme` on site 2: renamed in place, both rows still attached to site 2 with its new domain.
- **Upgrade after `do settheme`** (`example.com@2` plus sites for the LMS and CMS domains): warning printed, database left unchanged.
- **`SITE_ID` row deleted**: recreated for the LMS domain. **`SITE_ID` site already on a custom domain**: left unchanged.
- **`do settheme` after a fresh install**: the theme lands on site 2 (the renamed site), no duplicate LMS site.
- **`do init --limit=cms`** passes, and Studio resolves to site 2 on an installation with no Studio-domain `Site` (with one, for instance after `do settheme`, Studio resolves to that row, as it does today).
- **Platform names containing an apostrophe**: the init step reads `settings.PLATFORM_NAME` rather than interpolating `{{ PLATFORM_NAME }}` into the Python snippet, which would otherwise be an unterminated string literal and abort init under `sh -e`.
- Unit tests asserting the rendered init script, plus `unittest`, `mypy --strict`, `ruff check` and `ruff format`.

Known limitation: installations that have run `do settheme` (which includes every Indigo user) land in the ambiguous branch and need the manual consolidation now documented in `docs/troubleshooting.rst`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


muhammadadeeltajamul:
@ahmed-arb I have one concern about the CMS side that I think needs to be resolved before merging.

`lms.sh` renames the auto created `example.com` site in-place to `LMS_BASE`. `cms.sh` then do get_or_creates a Site row for the CMS domain. If the site object doesn't exist, it defaults back to SITE_ID = 2 (from `common_all.py`), so it will actually points to the lms instance. 

If any instance has a site with cms domain, any CMS code path that runs outside an HTTP request (Celery tasks, management commands, etc.) will falls back to `SITE_ID=2` and will still use LMS domain configurations instead of the CMS domain i.e. this change will reintroduce a variant of the exact bug this PR fixes, just scoped to Studio instead of LMS.

To actually make use of the site cms.sh creates, I think CMS needs its own SITE_ID (e.g. SITE_ID = 3) rather than inheriting 2 from common_all.py.

Any thoughts?

ahmed-arb:
Thanks @muhammadadeeltajamul . You're right that request-less CMS code falls back to `SITE_ID=2`, but I ended up going the other way: I dropped the Site creation from `cms.sh` rather than giving the CMS its own `SITE_ID`.

That block was doing harm rather than nothing. Open edX patches `Site.objects.get_current()` (in `django_sites_extensions`) to match the request host first, and to fall back to `SITE_ID` only when there is no request, so Studio requests already resolve by host. Creating a bare Site for the CMS domain takes that over, and Studio loses the SiteConfiguration it used to inherit from the `SITE_ID` site. Same database, the only difference being whether that row exists:

```
no CMS site row:  studio request -> site 2  (local.openedx.io),        SiteConfiguration={'PLATFORM_NAME': 'Configured on the LMS site'}
CMS site row:     studio request -> site 12 (studio.local.openedx.io), SiteConfiguration=None
```

`SITE_ID = 3` also isn't safe: `get_or_create` takes the next auto-increment id, and any instance that has run `do settheme` already has Sites at 3 to 6. `create_default_site` creates example.com at `pk=SITE_ID` for whichever service migrates first, so with different ids a `do init --limit=cms` on an empty database would leave the LMS `SITE_ID` pointing at a Site that doesn't exist. And I couldn't find a Studio code path that needs it: no `emulate_http_request` or `Site.objects.get_current()` under `cms/`, and `cms_user_tasks` reads site config through crum, which returns None without a request.

Tested with `tutor local` on a scratch TUTOR_ROOT, openedx 22.0.1:

* fresh install renames site 2 to the LMS domain; a second init is a no-op
* upgrade with a SiteConfiguration and a SiteTheme on site 2: renamed in place, both still attached
* upgrade after `do settheme`: warning printed, database untouched
* `SITE_ID` row deleted: recreated. Already on some other domain: left alone
* `do init --limit=cms` passes, and Studio resolves to site 2
* the bug itself, before and after, on the path `bulk_email` takes:

```
CELERY-TASK sees site: 2 example.com
CELERY-TASK sees site: 2 local.openedx.io
```

One other thing I hit while testing: the step now reads `settings.PLATFORM_NAME` instead of interpolating `{{ PLATFORM_NAME }}`, which was a syntax error for a platform name like `Bob's Academy` and aborted init.

Let me know if you disagree with any of this.

Abdul-Muqadim-Arbisoft (approved):
Thanks @ahmed-arb, totlly agree with the studio suggestion as well .

I tried this out on a real `tutor local` setup (openedx 22.0.2) with a bunch of different cases, upgrading an existing pre-fix DB, a fresh install with the default plugins (indigo on), the two-site ambiguous case, deleting the SITE_ID row, a custom domain, and a platform name with an apostrophe. All good, works exactly as you described. I'm also on board with dropping the Studio site creation and keeping a single SITE_ID.

One thing I noticed while testing though. The "Studio inherits the LMS site's SiteConfiguration" bit is only true as long as there's no Site row for the studio domain. But `do settheme` and indigo's init task both get_or_create `studio.<domain>`, so on a stock install that row exists anyway and Studio ends up resolving to it with no SiteConfiguration (I got site 4 on a fresh indigo install). Since indigo is on by default, that's probably where most people are.

Doesn't change anything in this PR, but could you:

1. tweak the description (and maybe the comment in common_all.py) so it doesn't sound like Studio inherits the LMS config out of the box?
2. open a follow-up for `settheme` + indigo so they either stop creating the bare studio site or attach the theme without breaking the inheritance? Separate change, no need to block this one on it.

Approving this one though with a request for follow up pr for above issues

ahmed-arb:
Thanks for digging into it @Abdul-Muqadim-Arbisoft , you're right.

I've reworded the Approach bullet and the `common_all.py` comment so neither implies Studio inherits the LMS config out of the box.

Two things worth carrying into the follow-up (opening it now): settheme does attach the SiteTheme to the Studio site, so theming works, it's only the SiteConfiguration that's missing. And tutor never creates a SiteConfiguration itself, so on a stock install there's nothing to lose. It only bites installs that add one through a plugin.

muhammadadeeltajamul (approved):

PR #1436, opened 2026-07-23
fix: bump build-time setuptools to 80.9.0 for the main branch

Abdul-Muqadim-Arbisoft:
The setuptools build constraint added in #1411 pins 69.1.1, which is forward-ported from the release branch to main. On main we build against edx-platform's master branch, where mysqlclient==2.2.8 declares its license using the PEP 639 `project.license` format. setuptools only understands that format from 77.0.0 onwards, so building the openedx image with setuptools 69.1.1 fails:

    configuration error: `project.license` must be valid exactly by one

Bump the pinned version to 80.9.0 on main. It is new enough to build the master requirements (mysqlclient 2.2.8) and still ships pkg_resources, which setuptools 82.0.0 removes but which is still required at runtime by dependencies such as fs (pyfilesystem2) matching edx-platform's own `setuptools<82` constraint.

The release branch intentionally stays on 69.1.1: Ulmo's requirements build fine with it, and this avoids changing the setuptools version shipped by the current stable release.


ahmed-arb (commented):
We should add an entry in changelog for this.

HammadYousaf01 (approved):

PR #1470, opened 2026-08-21
fix: handle comma-separated MONGODB_HOST in LMS init dockerize wait

carlos-marquez-wgu:
# Summary

When running the LMS init job with MONGODB_HOST set to a comma-separated string (ReplicaSet configuration), it fails at [lms.sh#6](https://github.com/overhangio/tutor/blob/v22.0.1/tutor/templates/jobs/init/lms.sh#L6).

Issue: https://github.com/overhangio/tutor/issues/1473

## Logs

```
2026/08/20 21:49:07 Waiting for tcp://mongodb,mongodb,mongodb:27017: dial tcp: lookup mongodb,mongodb,mongodb: no such host.
2026/08/20 21:49:08 Waiting for tcp://mongodb,mongodb,mongodb:27017: dial tcp: lookup mongodb,mongodb,mongodb: no such host.
2026/08/20 21:49:09 Waiting for tcp://mongodb,mongodb,mongodb:27017: dial tcp: lookup mongodb,mongodb,mongodb: no such host.
...
2026/08/20 21:49:09 Waiting for tcp://mongodb,mongodb,mongodb:27017: dial tcp: lookup mongodb,mongodb,mongodb: no such host.
2026/08/20 21:49:10 Failed to wait: timed out: tcp://mongodb,mongodb,mongodb:27017.
```

## Fix

When `MONGODB_HOST` contains a comma, `dockerize -wait` now targets only the first member.

## How-to test

While you can configure your own ReplicaSet, a quick local test is to repeat the same host, then run the LMS init job again like:

```
tutor config save --set MONGODB_HOST=mongodb,mongodb,mongodb
tutor dev do init -l lms
```

### DIFF
--- a/tutor/templates/jobs/init/lms.sh
+++ b/tutor/templates/jobs/init/lms.sh
@@ -2,6 +2,9 @@ dockerize -wait tcp://{{ MYSQL_HOST }}:{{ MYSQL_PORT }} -timeout 20s
 
 {%- if MONGODB_HOST.startswith("mongodb+srv://") %}
 echo "MongoDB is using SRV records, so we cannot wait for it to be ready"
+{%- elif ',' in MONGODB_HOST %}
+echo "Waiting for MongoDB (replica set, checking first member)"
+dockerize -wait tcp://{{ MONGODB_HOST.split(',')[0] }}:{{ MONGODB_PORT }} -timeout 20s
 {%- else %}
 dockerize -wait tcp://{{ MONGODB_HOST }}:{{ MONGODB_PORT }} -timeout 20s
 {%- endif %}
--- a/tutor/templates/jobs/init/lms.sh
+++ b/tutor/templates/jobs/init/lms.sh
@@ -3,8 +3,10 @@ dockerize -wait tcp://{{ MYSQL_HOST }}:{{ MYSQL_PORT }} -timeout 20s
 {%- if MONGODB_HOST.startswith("mongodb+srv://") %}
 echo "MongoDB is using SRV records, so we cannot wait for it to be ready"
 {%- elif ',' in MONGODB_HOST %}
-echo "Waiting for MongoDB (replica set, checking first member)"
-dockerize -wait tcp://{{ MONGODB_HOST.split(',')[0] }}:{{ MONGODB_PORT }} -timeout 20s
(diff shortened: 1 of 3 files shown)


tutor/templates/jobs/init/lms.sh
```diff
@@ -2,6 +2,9 @@ dockerize -wait tcp://{{ MYSQL_HOST }}:{{ MYSQL_PORT }} -timeout 20s
 
 {%- if MONGODB_HOST.startswith("mongodb+srv://") %}
 echo "MongoDB is using SRV records, so we cannot wait for it to be ready"
+{%- elif ',' in MONGODB_HOST %}
+echo "Waiting for MongoDB (replica set, checking first member)"
+dockerize -wait tcp://{{ MONGODB_HOST.split(',')[0] }}:{{ MONGODB_PORT }} -timeout 20s
```
Faraz32123: Thanks for the fix — this resolves the immediate crash, but it only waits for the first member of the replica set. If that host is up but the others aren't, the init job could still move on before MongoDB is actually ready. Could you loop over all comma-separated hosts instead, so each one is waited on? Something like:
```
{%- elif ',' in MONGODB_HOST %}
echo "Waiting for MongoDB (replica set, checking all members)"
{%- for host in MONGODB_HOST.split(',') %}
dockerize -wait tcp://{{ host.strip() }}:{{ MONGODB_PORT }} -timeout 20s
{%- endfor %}
{%- else %}
```

Added .strip() on each host too, so a space after the comma (e.g. mongodb1, mongodb2) doesn't cause the same failure as the original bug.

Faraz32123 (commented):
Nit: It might also be worth a one-line doc update on `MONGODB_HOST` in `docs/configuration.rst:250` noting it accepts a comma-separated list of hosts for replica set configs.

tutor/templates/jobs/init/lms.sh
carlos-marquez-wgu: Done, looping over all hosts now, with `.strip()` on each.

carlos-marquez-wgu:
Added one-line doc, thanks!

Faraz32123 (approved):
LGTM!

ahmed-arb (approved):

PR #1475, opened 2026-08-31
chore: restore main version suffix and plugin main branches

arbrandes:
### Description

The v22.0.1 release left `__version_suffix__` empty on main, so main builds identify as plain 22.0.1, use the `tutor` app name and root instead of `tutor-main`, and resolve the plugin index and Open edX revisions to the release instead of master. This restores the `main` suffix and repoints requirements/plugins.txt at the plugins' main branches, as described in the Tutor Main docs.

It also stops the suffix from being appended to an app name that was explicitly set via `TUTOR_APP`, which would otherwise turn a deliberate `TUTOR_APP=myapp` into `myapp-main`.

### LLM usage notice

### DIFF
--- a/requirements/plugins.txt
+++ b/requirements/plugins.txt
@@ -1,14 +1,14 @@
-# change version ranges when upgrading from verawood
-tutor-android>=22.0.0,<23.0.0
-tutor-cairn>=22.0.0,<23.0.0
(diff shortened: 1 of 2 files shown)


ahmed-arb (approved):
Thanks for the PR!

PR #1476, opened 2026-08-31
fix: don't append version suffix to an explicit TUTOR_APP

arbrandes:
### Description

On branches with a version suffix, such as main, the suffix was appended to the app name even when that name came from the `TUTOR_APP` environment variable, turning a deliberate `TUTOR_APP=myapp` into `myapp-main`. The suffix is now applied only to the default app name, and `TUTOR_APP` is used as-is.

### LLM usage notice


ahmed-arb (approved):
LGTM.

arbrandes:
Mind merging it?  Thanks!

ahmed-arb:
Fixing a little space crisis on the CI host 🫠. Waiting for the previous pipelines to finish before I merge this.

PR #1477, opened 2026-09-02
fix: set Open edX feature toggles as flat Django settings

arbrandes:
### Description

edx-platform recently flattened `FEATURES` into the settings namespace ([PR #37067](https://github.com/openedx/openedx-platform/pull/37067)), so entries like `ENABLE_COURSEARE_SEARCH` are now plain Django settings. Assigning to `FEATURES[...]` from a downstream settings module no longer has any effect, and the value silently stays at its default.

This sets all of Tutor's feature toggles as flat settings instead, and drops `ENABLE_COURSEWARE_MICROFRONTEND`, which no longer exists upstream. The plugin tutorial is updated too, since it taught the now-broken pattern. The `FEATURES:` blocks in the env YAML files are unaffected and left as-is.

### LLM usage notice


Abdul-Muqadim-Arbisoft:
Thanks @arbrandes. I tested this against a running `tutor dev` (edx-platform master mounted; LMS and CMS, dev and prod settings rendered). Two points.

### 1. Should this target `release` as well?

The upstream flattening (openedx/edx-platform#37067) landed in September 2025, so it is already in `release/verawood.1`, which Tutor v22 pins. Do you think this should go to `release` too, or be backported once it lands on `main`?

For context, on verawood.1 the bug is small today: of the 23 places edx-platform reads the toggles Tutor sets, 22 still go through `settings.FEATURES.get(...)`, and those do see our current `FEATURES[...]` writes. The one miss is `is_discussion_enabled()` in `openedx/core/djangoapps/discussions/services.py`, which reads `settings.ENABLE_DISCUSSION_SERVICE` and gets upstream's True instead of our False.

### 2. Either way, the flat assignments need the proxy rebind

The `FEATURES` our settings inherit via `from lms.envs.production import *` is a `FeaturesProxy` bound to edx-platform's module namespace, not ours. The flat assignments fix `settings.X`, but anything upstream that still reads `settings.FEATURES.get("X")` now sees the upstream default instead of what Tutor set:

- On master: the Mako templates (`lms/templates/index.html`, `courseware/courses.html`, both `navbar-not-authenticated.html`, `dashboard.html`). Measured: `settings.ENABLE_COURSE_DISCOVERY` is True but `settings.FEATURES["ENABLE_COURSE_DISCOVERY"]` is False, so the "Explore courses" link, the `/courses` search box and dashboard search disappear. Today those templates see True.
- On verawood.1: the 22 reads above flip to defaults. CORS middleware, courseware search, CMS reindex and course discovery would all silently turn off. So this PR must not be backported without the fix below.
- Plugins still writing `FEATURES["X"] = True` stay broken on both branches either way.

The fix is what openedx/edx-platform#37345 asks of every custom settings module: rebind the proxy in `partials/common_all.py` right after the upstream import, keeping your flat assignments:

```python
try:
    from openedx.core.lib.features_setting_proxy import FeaturesProxy
    FEATURES = FeaturesProxy(globals())
except ImportError:
    pass
```

Verified in all four settings modules: with this, `settings.X` and `settings.FEATURES["X"]` agree for every toggle, and plugin `FEATURES["X"]` writes work again (with upstream's deprecation warning). It also makes the forum/notes/discovery PRs cleanups rather than blockers (#1479).

Abdul-Muqadim-Arbisoft:
> Thanks @arbrandes. I tested this against a running `tutor dev` (edx-platform master mounted; LMS and CMS, dev and prod settings rendered). Two points.
> 
> ### 1. Should this target `release` as well?
> The upstream flattening ([openedx/openedx-platform#37067](https://github.com/openedx/openedx-platform/pull/37067)) landed in September 2025, so it is already in `release/verawood.1`, which Tutor v22 pins. Do you think this should go to `release` too, or be backported once it lands on `main`?
> 
> For context, on verawood.1 the bug is small today: of the 23 places edx-platform reads the toggles Tutor sets, 22 still go through `settings.FEATURES.get(...)`, and those do see our current `FEATURES[...]` writes. The one miss is `is_discussion_enabled()` in `openedx/core/djangoapps/discussions/services.py`, which reads `settings.ENABLE_DISCUSSION_SERVICE` and gets upstream's True instead of our False.
> 
> ### 2. Either way, the flat assignments need the proxy rebind
> The `FEATURES` our settings inherit via `from lms.envs.production import *` is a `FeaturesProxy` bound to edx-platform's module namespace, not ours. The flat assignments fix `settings.X`, but anything upstream that still reads `settings.FEATURES.get("X")` now sees the upstream default instead of what Tutor set:
> 
> * On master: the Mako templates (`lms/templates/index.html`, `courseware/courses.html`, both `navbar-not-authenticated.html`, `dashboard.html`). Measured: `settings.ENABLE_COURSE_DISCOVERY` is True but `settings.FEATURES["ENABLE_COURSE_DISCOVERY"]` is False, so the "Explore courses" link, the `/courses` search box and dashboard search disappear. Today those templates see True.
> * On verawood.1: the 22 reads above flip to defaults. CORS middleware, courseware search, CMS reindex and course discovery would all silently turn off. So this PR must not be backported without the fix below.
> * Plugins still writing `FEATURES["X"] = True` stay broken on both branches either way.
> 
> The fix is what [openedx/openedx-platform#37345](https://github.com/openedx/openedx-platform/issues/37345) asks of every custom settings module: rebind the proxy in `partials/common_all.py` right after the upstream import, keeping your flat assignments:
> 
> ```python
> try:
>     from openedx.core.lib.features_setting_proxy import FeaturesProxy
>     FEATURES = FeaturesProxy(globals())
> except ImportError:
>     pass
> ```
> 
> Verified in all four settings modules: with this, `settings.X` and `settings.FEATURES["X"]` agree for every toggle, and plugin `FEATURES["X"]` writes work again (with upstream's deprecation warning). It also makes the forum/notes/discovery PRs cleanups rather than blockers (#1479).

@arbrandes did you got time to go through this?

arbrandes:
@Abdul-Muqadim-Arbisoft, nice catch: I addressed the rebinding in the last commit.

We can backport it to Verawood after this lands.

Abdul-Muqadim-Arbisoft:
Thank you @arbrandes, can you change the base branch to release instead it will be ported to verawood

arbrandes:
@Abdul-Muqadim-Arbisoft, done!

Abdul-Muqadim-Arbisoft:
@arbrandes just a short change, Changelog doesn't mention the ENABLE_COURSEWARE_MICROFRONTEND removal.Can you just add that then we are good to merge this pr.

arbrandes:
@Abdul-Muqadim-Arbisoft, done!

Abdul-Muqadim-Arbisoft:
@arbrandes, I think the last update went in the wrong direction, the branch got rebased onto `main` rather than `release`.
Yesterday the PR was 2 commits / 6 files. It's now 9 commits / 22 files .

arbrandes:
Apologies.  That wasn't intentional.  Fixed.

Issue #1294, opened 2025-10-28
Suggestion: Early dev releases for named release branches


bmtcril:
As a Tutor plugin developer, it's difficult to manage versioning before Tutor makes a PyPI release for a named release cut. We manage several plugins that we are developing against unreleased Tutor / Tutor MFE versions, and cannot release those plugins for testing until Tutor does a release. The plugins also cannot have the correct Tutor required versions until that release is done leading to issues where, for example, the plugin needs to require Tutor 20 in order to test against the Ulmo branch.

As a way to work around this, I'd like to suggest the following:

- When a new release is cut, create the branch for the next release and immediately bump the major version in that branch
- On a cadence (on merge, daily, weekly, etc) tag and create a new dev release (ex: `22.0.0dev1`) without changing the actual version in `__about__.py`.

This will allow plugin developers to target `>=22` immediately and enable finer grained dependency control during development so we can target something like `>22.0.0dev55` when breaking changes land in the branch. It should also simplify setting up testing sandboxes against named releases going forward since we should be able to release plugin versions that target the right things instead of having to install from branches until the main Tutor packages are bumped.

ahmed-arb:
Development releases are live: [`tutor 23.0.0.dev3180`](https://pypi.org/project/tutor/23.0.0.dev3180/) is on PyPI, published from `main` by the release pipeline.

What landed, in three parts:

1. `main` now carries the version of the upcoming release (#1483). It derives that version from the RELEASE number rather than storing it, so the `__version__` line stays identical on both branches and the automated release-to-main merge never conflicts on it. No post-release bump to remember, and it rolls over on its own at each cut.
2. Every merge to `main` publishes a development release (#1484), numbered from the pipeline id. No tag is created, and the dev number applies to the package version only, so the Tutor Main image tags stay at `23.0.0-main` for the whole cycle instead of churning on every commit.
3. The 13 officially supported plugins now declare `tutor>=23.0.0,<24.0.0` on their `main` branches, so installing them no longer pulls a released Tutor over the one being developed against.

The pins you were after:

- `tutor>=23.0.0.dev0,<24.0.0.dev0` to follow the development cycle
- `tutor>=23.0.0.dev3180` to require a change that landed in a specific build

[...]

PR #1482, opened 2026-09-09
fix: derive Tutor Main plugin sources from the version suffix

### LINKED ISSUE #1310: Recurring Merge Conflicts in plugins.txt Between Release and Main Branches
## Problem

Every time changes are merged from the `release` branch to `main` in the GitLab pipeline, merge conflicts occur in `plugins.txt` because:

- **Release branch**: Installs plugins from PyPI with version constraints (e.g., `tutor-android>=20.0.0,<21.0.0`)
- **Main branch**: Installs plugins directly from their GitHub main branches (e.g., `https://github.com/overhangio/tutor-android@main`)

This structural difference causes predictable conflicts on every release merge, requiring manual intervention to resolve.

## Proposed Solution

Handle plugin installation similarly to how we handle `OPENEDX_COMMON_VERSION` between main and release branches. Use the `__version_suffix__` to determine which plugins configuration to use.

### Implementation Strategy

Modify `.hatch_build.py` to conditionally load plugins based on `__version_suffix__`:

- If `__version_suffix__` is `"main"` → load plugins from a separate config file (e.g., `plugins-main.txt`) that uses GitHub URLs
- Otherwise → load plugins from `plugins.txt` with PyPI version constraints

This way:
- Both branches can have the plugin lists in their respective config files
- No merge conflicts occur since each branch uses its own config file
- The logic to determine which file to use lives in `.hatch_build.py`

## Impact

Eliminates recurring manual merge conflict resolution in the GitLab CI/CD pipeline, reducing maintenance overhead and preventing potential deployment delays.

## References

- Documentation: [GitLab and Release Pipeline](https://docs.edly.space/doc/gitlab-and-release-pipeline-TaepmuSZ25)

Abdul-Muqadim-Arbisoft:
plugins.txt used to hold different content on the release and main branches: PyPI version ranges on release, GitHub main branch URLs on main. Every merge from release into main overwrote the main version, and it then had to be restored by hand (see #1343, #1450, #1475).

The file now always contains PyPI version ranges. When the version suffix is "main", the hatch metadata hook rewrites each requirement to install from the main branch of the plugin's GitHub repository instead. This is the same mechanism that sets OPENEDX_COMMON_VERSION to "master" in Tutor Main, and it keeps plugins.txt identical between branches so it no longer conflicts.

The smoke workflows install plugins via the "full" extra rather than the requirements file, so that the branch logic applies there too.

Closes #1310

### DIFF
--- a/.github/workflows/smoke-k8s.yml
+++ b/.github/workflows/smoke-k8s.yml
@@ -62,11 +62,11 @@ jobs:
       - name: Install tutor
         run: pip install -e .[dev]
 
-      # We install using the plugins file instead of tutor plugins install
+      # We install using the "full" extra instead of tutor plugins install
       # as we want the workflow in the main branch to install plugins on
-      # the main branch as well. plugins.txt automatically manages this.
+      # the main branch as well. .hatch_build.py automatically manages this.
       - name: Install tutor plugins
-        run: pip install -r requirements/plugins.txt
+        run: pip install -e .[full]
 
       - name: Enable tutor plugins
         run: tutor plugins enable ${{ github.event.inputs.plugins || 'indigo mfe' }}
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -43,7 +43,7 @@ jobs:
         run: pip install -e .[dev]
 
       - name: Install tutor plugins
-        run: pip install -r requirements/plugins.txt
+        run: pip install -e .[full]
 
       - name: Enable tutor plugins
         run: tutor plugins enable ${{ github.event.inputs.plugins || 'indigo mfe' }}
--- a/.hatch_build.py
+++ b/.hatch_build.py
@@ -1,10 +1,15 @@
 # https://hatch.pypa.io/latest/how-to/config/dynamic-metadata/
 import os
+import re
 import typing as t
 
 from hatchling.metadata.plugin.interface import MetadataHookInterface
 
 HERE = os.path.dirname(__file__)
+
+# Official plugins listed in requirements/plugins.txt are all hosted under this
+# GitHub organization, in a repository that has the same name as the package.
+PLUGINS_GITHUB_ORG = "https://github.com/overhangio"
 
 
 class MetaDataHook(MetadataHookInterface):
@@ -12,7 +17,7 @@ class MetaDataHook(MetadataHookInterface):
         metadata["dependencies"] = load_requirements("base.in")
         metadata["optional-dependencies"] = {
             "dev": load_requirements("dev.txt"),
-            "full": load_requirements("plugins.txt"),
+            "full": load_plugin_requirements(),
         }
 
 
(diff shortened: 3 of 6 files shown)


ahmed-arb (commented):
This is the right mechanism. Deriving the main-branch URLs from the requirement names keeps a single list, so it cannot drift the way a separate `plugins-main.txt` would.

Three things before it goes in:

1. Line 1 of `plugins.txt` now differs from the one on `release`, so it will conflict at the next release cut, when the "change version ranges when upgrading from verawood" comment gets updated. Since the point of the PR is to stop that file from conflicting, better to leave line 1 alone, or make the same edit on both branches.

2. The merge that lands this is itself the last conflict: `deploy:main` will see `plugins.txt` changed on both sides and stop. It has to be resolved by taking the release side whole. Resolving it the other way is what happened in #1475, and it would silently restore the git refs.

3. `load_plugin_requirements` assumes the repository name matches the package name, and the regex drops everything after the name, so extras or environment markers would be discarded silently. That holds for all 13 plugins today; worth a comment so the next person adding one knows.

Unrelated to this PR, but for context: on main this still puts direct references in the sdist metadata, which is what PyPI rejects (see #1450). That is fine here, and it is the piece we will have to revisit if we start publishing dev releases from main (#1294).

Abdul-Muqadim-Arbisoft:
> This is the right mechanism. Deriving the main-branch URLs from the requirement names keeps a single list, so it cannot drift the way a separate `plugins-main.txt` would.
> 
> Three things before it goes in:
> 
> 1. Line 1 of `plugins.txt` now differs from the one on `release`, so it will conflict at the next release cut, when the "change version ranges when upgrading from verawood" comment gets updated. Since the point of the PR is to stop that file from conflicting, better to leave line 1 alone, or make the same edit on both branches.
> 2. The merge that lands this is itself the last conflict: `deploy:main` will see `plugins.txt` changed on both sides and stop. It has to be resolved by taking the release side whole. Resolving it the other way is what happened in #1475, and it would silently restore the git refs.
> 3. `load_plugin_requirements` assumes the repository name matches the package name, and the regex drops everything after the name, so extras or environment markers would be discarded silently. That holds for all 13 plugins today; worth a comment so the next person adding one knows.
> 
> Unrelated to this PR, but for context: on main this still puts direct references in the sdist metadata, which is what PyPI rejects (see #1450). That is fine here, and it is the piece we will have to revisit if we start publishing dev releases from main (#1294).
@ahmed-arb totally agree,

I'll put line 1 back exactly as it was and add the explanation below it, so the verawood marker stays searchable like the other "upgrade from verawood" comments. Small nuance: once the forward-port in (2) is resolved, the file is identical on both branches, so a later range bump on release merges cleanly. I simulated a Willow bump after the reset and got no conflict.

Yes. I also tried the other way round: if release doesn't touch plugins.txt at all, the merge into main goes through cleanly and main silently keeps its git refs. So the conflict is actually what forces the reset. I'll add a note to the description: resolve it once by taking the release side whole.

I'll go one step further than a comment and make the parser strict: only <name><version specifier> is accepted, anything with extras, markers or a URL raises. That way a badly resolved merge that leaves @git+ refs in the file fails the build on main instead of quietly working like in #1475.

On the aside, now that #1484 is in: I pulled the published 23.0.0.dev3180 sdist. It ships an empty plugins.txt and no full entries, so the pipeline seems to blank the file before building. This hook returns an empty list in that case, so dev releases are unaffected. If you want, I can follow up so the hook does that itself when TUTOR_DEV_RELEASE is set and the pipeline step can go.

ahmed-arb (approved):

Abdul-Muqadim-Arbisoft:
@ahmed-arb jsut pushed the three items from my comment above: marker line restored, strict parser, and the Merging note in the description. Ready to merge from my side.

